PR #25, opened 2020-07-21
Fix overrun

seabbs:
* Reduced number of checks on variable limits 
* Corrected initial R value as per #22 
* Constrained Poisson and NB random number generators to be less than or equal to 1e9. This prevents the exception and should not impact real results. Another approach that I explored was restricting the prior. This worked but it was difficult to find a valid reason beyond fixing this issue for constraining the prior.


seabbs:
Just updating the docs and then will probably merge this as no major changes.

PR #27, opened 2020-07-24
reparameterise lognorm

jdmunday:
reparameterising lognorm_dist_def with 

    mean_location <- log(mean^2 / sqrt(mean_sd^2 + mean^2))
    mean_shape <- sqrt(log(1 + (mean_sd^2 / mean^2)))
    mean <- mean_location
    mean_sd <- mean_shape

rather than previous log(mean) and log(sd)


NAMESPACE
```diff
@@ -1,36 +1,5 @@
 # Generated by roxygen2: do not edit by hand
 
-export(R_to_growth)
-export(adjust_infection_to_report)
-export(bootstrapped_dist_fit)
```
seabbs, line 5: Ah - removing this is what is causing the build to fail. This has happened during the failed compilation I think. I will clean up and merge

R/dist.R
```diff
@@ -317,10 +317,15 @@ lognorm_dist_def <- function(mean, mean_sd,
                              to_log = FALSE) {
   
   if (to_log) {
-    mean <- log(mean)
```
seabbs, line 320: Thinking about this should this be happening after sampling? I think yes. Will check out and fix

PR #41, opened 2020-08-04
Add import rate

seabbs:
* Adds a constant but uncertain level of imports 
* Set this to be gamma distributed with a default prior of mean = 1, sd = 1
* Currently introduce instability in fitting with 100s of divergent transitions and chains not converging. 

As this is just a scaling of observed data potentially it can't be fit without more data and/or a highly penalised prior.

### DIFF
--- a/R/estimate_infections.R
+++ b/R/estimate_infections.R
@@ -93,14 +93,16 @@
 #' # Run model with default settings
 #' def <- estimate_infections(reported_cases, family = "negbin",
 #'                            generation_time = generation_time,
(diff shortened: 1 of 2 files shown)


seabbs:
Aruba really highlights this issue with very few cases until recently. This may be a data problem, a result of lots of imports or maybe due to trying to estimate Rt over a long period of time with no data (but the Rt plot really suggests this not the case). 

Screenshot 2020-08-13 at 18 09 51

hsbadr:
@seabbs Do you want to rebase this?

PR #66, opened 2020-09-15
Data process update

seabbs:
* Added 20% HDI in plotting and summary output
* Added growth rate summary csv
* Updated handling of samples internally when not returned to reduce RAM usage. 
* Updated documentation.


R/epinow.R
```diff
@@ -445,17 +454,19 @@ regional_epinow <- function(reported_cases,
                                               future.scheduling = Inf)
 
   # names on regional_out
-  names(regional_out) <- regions # ["foo" => a, "bar" => b$error, "baz" => c, "parrot" => d$error]
+  names(regional_out) <- regions
```
joeHickson, line 457: sorry, I typed this talking a colleague through a problem I was having and managed to leave it in!

NEWS.md
```diff
@@ -1,6 +1,10 @@
 # EpiNow2 1.2.0
+
 * Added timeout and timing option to `regional_epinow`
 * Improved logging of warnings in `regional_epinow`
+* Added inner 20% high density region.
```
joeHickson, line 5: I can see we don't have a 1.2.0 tag officially published yet but we are using 1.2 in covid-rt-estimates. We should probably have a chat on a weekly catchup about how we are going to manage releases as it's an easy way to get in a sticky mess with multiple projects and dependencies.

R/epinow.R
seabbs: no problem!

NEWS.md
seabbs: i've been publishing tags when a version is wrapped up but happy to discuss. Sounds like a good idea!

joeHickson (changes requested):
A couple of bits to double check but otherwise LGTM

R/estimate_infections.R
```diff
@@ -518,10 +518,12 @@ estimate_infections <- function(reported_cases, family = "negbin",
     model <- stanmodels$estimate_infections
   }
   
-  futile.logger::flog.debug(paste0("Running for ", samples," samples (across ", chains,
-                   " chains each with a warm up of ", warmup, " iterations each) and ",
-                   data$t," time steps of which ", horizon, " are a forecast"))
-  
+  if (verbose) {
```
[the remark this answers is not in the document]
seabbs: So I added this back in as I was seeing the debug message showing by default (i.e without calling a logger). Most users are probably not going to call the logger so this behaviour needs to have the correct defaults. I am happy to use another approach for this?

R/summarise.R
```diff
@@ -166,8 +168,8 @@ summarise_results <- function(regions,
 #' regional_out <- regional_epinow(reported_cases = cases,
 #'                                 generation_time = generation_time,
 #'                                 delays = list(incubation_period, reporting_delay),
-#'                                 samples = 2000, warmup = 200, cores = 4,
-#'                                 adapt_delta = 0.95, chains = 4, verbose = TRUE,
```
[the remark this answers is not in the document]
seabbs: Yes, I dropped it as all output is printed in the rendered site docs so having the stan output, etc can make for a very long web page.

seabbs:
Merging for now as I think covered the main points but happy to revisit/improve in another PR.

PR #71, opened 2020-09-25
fix discretised delay distributions

sbfnk:
Currently, the discretised delay distributions aren't correctly normalised. Because of this, they can sum up to less than 1, and this currently happens in the second lognormal delay. I suspect this is behind the current issue where infections can be consistently above reported cases. I think it is fixed in this commit.

### DIFF
--- a/inst/stan/estimate_infections.stan
+++ b/inst/stan/estimate_infections.stan
@@ -195,7 +195,7 @@ transformed parameters {
     vector[max_delay[s]] rev_delay = rep_vector(1e-5, max_delay[s]);
     for (j in 1:(max_delay[s])) {
       rev_delay[j] +=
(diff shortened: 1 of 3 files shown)


seabbs:
Yes makes sense - good find. This was reimplemented since we validated on synthetic data last so makes sense as the source of the issue. 

Should the lognormal be truncated at 1? The gamma is truncated at 1 due to being used for the Cori method (as same day infections need to be excluded) but this isn't the case for delays.

seabbs (approved):

sbfnk:
Good point - second commit changes the delays to start from 0 if I've wrapped my head around the convolution correctly.

seabbs (approved):

seabbs:
Yes, I think that is now correct I think - as is perhaps clear I find the indexing a little mind-bending.

PR #78, opened 2020-09-29
epinow: Reorder the arguments passed to estimate_infections

hsbadr:
`epinow` is disconnected from `estimate_infections` after implementing Automatic Differentiation Variational Inference (ADVI; Kucukelbir et al. 2015) via `method = "approximate"` and restructuring the arguments. This builds on #74 to update the arguments passed to `estimate_infections`.

### DIFF
--- a/R/epinow.R
+++ b/R/epinow.R
@@ -121,15 +121,14 @@ epinow <- function(reported_cases, model, samples = 1000, stan_args,
     model <- NULL
   }
 
-  estimates <- estimate_infections(reported_cases = reported_cases, model = model, samples = samples,
-                                   family = family, generation_time = generation_time,
-                                   delays = delays, gp = gp, rt_prior = rt_prior,
-                                   stan_args = stan_args, future = future,
-                                   max_execution_time = max_execution_time, estimate_rt = estimate_rt, 
-                                   estimate_week_eff = estimate_week_eff, estimate_breakpoints = estimate_breakpoints,
-                                   burn_in = burn_in, stationary = stationary, fixed = fixed,
-                                   fixed_future_rt = fixed_future_rt, horizon = horizon,
-                                   verbose = verbose, return_fit = return_fit)
+  estimates <- estimate_infections(reported_cases = reported_cases, model = model, samples = samples, stan_args = stan_args,
+                                   method = method, family = family,
+                                   generation_time = generation_time, delays = delays, horizon = horizon,
+                                   gp = gp, rt_prior = rt_prior,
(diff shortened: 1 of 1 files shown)


joeHickson:
LGTM - correctly ordering arguments to match the function signature is good.

seabbs (approved):
Looks good, thanks for this @hsbadr. Will merge once #74 is merged.

hsbadr:
@joeHickson It isn't about the order. The new argument `method`, @seabbs introduced, was ignored. So, without this, `epinow` will call `estimate_infections` using the default NUTS method (`method = "exact"`) even if the user selected variational inference (`method = "approximate"`), which would also fail if the user supplied incompatible STAN arguments via `stan_args`.

seabbs:
yes - great spot!

seabbs:
actually can merge into PR branch now it seems - thanks again @hsbadr

hsbadr:
@seabbs Sounds good! I have run a few tests with different combinations of `stan_args`. It seems that variational inference could be improved with tuning the parameters (e.g., better uncertainty estimates). Moreover, it fails more often if `R` is compiled and optimized by Intel compilers with Intel Math Kernel Library (MKL). I think _the model and its initialization may need to be updated for variational inference_ to work consistently from the first trail, without the new `trials` argument.

seabbs:
Sounds interesting. I am still reading around the approach so any suggestions appreciated (its definitely still an experimental feature) - perhaps opening an issue would be a good way to kickstart discussions. So for all my tweaks away from default settings have actually made performance worse. 

Agree on changes to the model + initialization. In general, we are hoping for some updates in the near term to increase performance and decrease compute time. Obviously also interested in suggestions there as well.

hsbadr:
@seabbs JFYI. I've been testing the ADVI arguments:
```
  stan_args = list(
    algorithm = "fullrank",           # defaulting to "meanfield", or "fullrank"
    importance_resampling = FALSE,    # defaulting to FALSE
    iter = 10000,                     # defaulting to 10000
    grad_samples = 1,                 # defaulting to 1
    elbo_samples = 100,               # defaulting to 100
    tol_rel_obj = 0.01,               # defaulting to 0.01
    eval_elbo = 100,                  # defaulting to 100
    output_samples = 2000,            # defaulting to 1000
    adapt_engaged = TRUE,             # defaulting to TRUE
    adapt_iter = 100,                 # defaulting to 50, ignored if adapt_engaged = FALSE
    eta = 0.1,                        # ignored if adapt_engaged = TRUE
    trials = 10                       # defaulting to 10
  )
```
Unfortunately, as you've noticed, the convergence isn't guaranteed and I've to tweak the `create_initial_conditions` function with unreliable results, especially for real noisy data. I've reverted back to NUTS algorithm.

PR #84, opened 2020-09-30
Suppress region printing

seabbs:
This PR adds a switch when there are more than 30 regions so that instead of printing the names of all the regions to produce estimates for it logs the number of regions (and the total number of regions). It also adds a region excluded logging message in order to highlight when this is happening.


seabbs:
Just fixing the exclude message - do not merge.

PR #87, opened 2020-10-01
More ram efficiency

seabbs:
* Adds new arguments that allow skipping plot construction and drop samples early. Also turns off estimate returns in regional epinow if no summary is requested and estimates are not otherwise wanted.

### DIFF
--- a/R/epinow.R
+++ b/R/epinow.R
@@ -5,6 +5,9 @@
 #' @param target_date Date, defaults to maximum found in the data if not specified.
 #' @param target_folder Character string specifying where to save results (will create if not present).
 #' @param return_estimates Logical, defaults to TRUE. Should estimates be returned.
+#' @param keep_samples Logical, defaults to TRUE. Should samples be kept (either returned or saved or both 
+#' depending on other settings).
+#' @param make_plots Logical, defaults to TRUE. Should plots be made and returned (or saved).
 #' @return A list of output from estimate_infections, forecast_infections,  report_cases, and report_summary.
 #' @export
 #' @inheritParams estimate_infections
@@ -254,9 +257,15 @@ epinow <- function(reported_cases, model, samples = 1000, stan_args,
   if (return_estimates) {
     out <- list()
     out$estimates <- estimates
-
+    if (!keep_samples) {
+      out$estimates$samples <- NULL
+    } 
+    
     if (!missing(forecast_model)) {
       out$forecast <- forecast
+      if (!keep_samples) {
+        out$forecast$samples <- NULL
+      } 
     }
 
     out$estimated_reported_cases <- estimated_reported_cases
--- a/R/epinow.R
+++ b/R/epinow.R
@@ -346,7 +346,7 @@ regional_epinow <- function(reported_cases, target_folder, target_date,
                                               target_folder = target_folder, 
                                               target_date = target_date,
                                               return_estimates = return_estimates,
-                                              return_summarised_estimates = summary,
+                                              return_partial_estimates = summary,
                                               ...,
                                               future.scheduling = Inf)
 
--- a/R/epinow.R
+++ b/R/epinow.R
@@ -346,7 +346,7 @@ regional_epinow <- function(reported_cases, target_folder, target_date,
                                               target_folder = target_folder, 
                                               target_date = target_date,
                                               return_estimates = return_estimates,
-                                              return_partial_estimates = summary,
+                                              return_partial_estimates = summary | return_estimates,
                                               ...,
                                               future.scheduling = Inf)
 
--- a/R/epinow.R
+++ b/R/epinow.R
@@ -60,7 +60,8 @@ epinow <- function(reported_cases, model, samples = 1000, stan_args,
                    burn_in = 0, stationary = FALSE, fixed = FALSE, fixed_future_rt = FALSE,
                    future = FALSE, max_execution_time = Inf, prior_smoothing_window = 7,
                    return_fit = FALSE, forecast_model, ensemble_type = "mean",
-                   return_estimates = TRUE, target_folder, target_date, verbose = FALSE) {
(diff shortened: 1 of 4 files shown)


hsbadr (commented):
It may be cleaner to combine all outputs that user can select to return/save into a vector of character strings; something like: `outputs = c("fit", "estimates", "samples", "plots")`. This would replace four arguments: `return_fit = TRUE`, `return_estimates = TRUE`, `keep_samples = TRUE`, and `make_plots = TRUE`. The user can select any subset of outputs (e.g.,  `outputs = c("estimates", "plots")` to get the estimates and plots only).

Then, we check if the `"samples"`, for example, is in the selected `outputs` by `if ("samples" %in% outputs) { ... }`.

hsbadr (commented):
`return_partial_estimates = TRUE` can be replaced by adding `"partial_estimates"` in the suggested `outputs` argument. To return everything: `outputs = c("fit", "estimates", "partial_estimates", "samples", "plots")` where a subset of the `outputs` vector can be excluded by the user.

hsbadr (commented):
The new arguments (`keep_samples` and `make_plots` | or `outputs`) need to be documented.

hsbadr:
@seabbs Besides the suggested organization of arguments (replacing return/save logical arguments by a a vector of character strings `outputs`), you may also make use of the `pars` and/or `include` Stan arguments to reduce the size of the `stanfit` object too.

|      Argument      |    Description    |
|:------------------:|:------------------|
**pars** | A vector of character strings specifying parameters of interest. The default is `NA` indicating all parameters in the model. If `include = TRUE`, only samples for parameters named in pars are stored in the fitted results. Conversely, if `include = FALSE`, samples for all parameters except those named in pars are stored in the fitted results.
**include** | Logical scalar defaulting to `TRUE` indicating whether to include or exclude the parameters given by the pars argument. If `FALSE`, only entire multidimensional parameters can be excluded, rather than particular elements of them.

joeHickson:
outputs = c("fit", "estimates", "partial_estimates", "samples", "plots")

May I mention previous review comment about "Magic Strings"?

definition:
OUTPUT_TYPES = list(FIT="fit", ESTIMATES="estimates", PARTIAL_ESTIMATES="partial_estimates", SAMPLES="samples", PLOTS="plots")

usage:
( .... outputs = unlist(OUTPUT_TYPES , use.names=FALSE))  <- all by converting the list values into a vector
(... outputs = c(OUTPUT_TYPES$ESTIMATES, OUTPUT_TYPES$SAMPLES) ) <- specify a sublist


I wonder if this actually hints that the function has got too broad in scope (software engineering concepts: reduce branching code (if/else blocks), keep function scope as small as possible). Perhaps a neater solution might be achieved by splitting down the functions and making them more accessible:
EpiNow does the calculation but no more. It needs to be told what modelling it's doing but that's it. You then call separate functions with that do additional processing if you want it (e.g. saving results to disk, plotting graphs, etc)
e.g.
```R
result <- epinow(...)       # result contains all the data and stats processing
save_raw_to_disk(results, output_dir) # optionally call to dump the raw data out 
save_estimates(results, output_dir) # saving data to disk - you could keep it in memory for later processing if you prefer
plot_result(result, output_dir)         # plotting is a distinct post-processing step
my_fn(result$get_partial_estimates())  # you can interact with the results and do additional post processing
rm(result) # tidy up to free up memory just in case
```

This should reduce the number of params needed and give each bit a smaller responsibility. It passes the orchestration / sequencing logic to a layer above and gives them a larger API to use. This does pass responsibility to those using the EpiNow2 library. For those who need a really simple interface you could include an example in code rather than just in documentation:
```R
epinow_save_everything <- function(output_dir, ...) {
    result <- epinow(...)
    save_raw_to_disk(result, output_dir)
    .... # call all the output functions and graph generation etc here
    rm(result)
}
```

Those that care about performance / control can then use the "official" interface and those that don't can be lazy and use the wrapper method (although if this was documentation only they could probably cope with copy/paste).

To do this properly would also require unpacking the summary bits too. Responsibilities would then be:

covidredgionaldata <- gets the data into the right format for processing
covid-rt-estimates (or similar) <- responsible for looping over data and controlling runtime options (currently it only loops over datasets, not regions). This would have responsibility for parallelisation of data processing.
EpiNow <- responsible for modelling single regions. Has helper functions for presenting data for individual regions and groups of regions (e.g. plot_result(result, output_dir), summarise_regions(results_vector), plot_summary(summary_result) )

As with the save_everything you could provide a really simple example method of how to loop over a dataset that could be called by those needing a simple run but really should be implemented properly by the calling code.

This would be a bigger refactor though so feel free to ignore / shelve for another time!

seabbs:
@hsbadr good suggestions - will look at this for a later release (could you please open an issue). @joeHickson comment on magic strings is a sensible one but agree the reduction in args could be really beneficial. The `stanfit` changes should be possible using `stan_args`? This is likely a fairly minor issue though as most users won't be getting the stan model back for extremely large runs (i.e of more than say 100 regions)? 

@joeHickson so the design structure here is that technical user (i.e like @pearsonca) will use the internal functions directly and less technical users (those with needs that are aligned with our main use case scenarios) will make use of the wrappers which handle saving etc for them. The overarching aim is that default usage will be 1. simple and 2. give the user everything they need to track an outbreak based on "expert" knowledge. This is driven by our experience early in the pandemic of there being no out the box tools that just "worked" which felt unacceptable. 

We make use of the wrappers as we are trying to make an expert informed wrapper so we need to use them in practice in order to make sure their development is sensible/stable. Looking at the internals of `epinow` the design pattern should be fairly clear with `estimate_infections` doing all the heavy lifting and other functions doing a single task with the output (though obviously there is more work to do functionalizing and generalizing this code. 

I am happy to consider design changes that fit this basic philosophy though (as I think modular functions + an example would) so happy to see an issue discussing more optimal approaches for a refactor in a later release.

PR #92, opened 2020-10-05
Improve logging

seabbs:
This PR improves the use of `futile.logger` by adding a hierarchy of named logs to `EpiNow2`. The default unnamed logger is "EpiNow2" (as per the `futile.logger` docs) with all logging called in `epinow` being linked to the "EpiNow2.epinow" logger. This means that logs can be split or default to the overall "Epinow2" logger if not defined. Loggers have also been defined within `estimate_infections` and fitting functions. 

This PR also adds support to `setup_logging` to enable this new functionality to be controlled and adds an example to the README outlining the intended use case. 

Ideally, this setup (though writing the log to a temp file for "EpiNow2.epinow" would be the default when calling `regional_epinow` but I currently see no way to achieve this). 

This PR also improves summary logs produced by `regional_epinow` in response to region level logs potentially no longer being seen by the user.


R/epinow.R
```diff
@@ -422,7 +423,7 @@ clean_regions <- function(reported_cases, non_zero_points) {
     futile.logger::flog.info("Producing estimates for: %s regions",
                              length(eval_regions))
     message <- ifelse(length(orig_regions) == 0, 0, 
-                      (length(orig_regions) - length(eval_regions)))
+                      length(orig_regions))
     futile.logger::flog.info("Regions excluded: %s regions",
```
joeHickson, line 427: Should all the logs be named?

R/epinow.R
```diff
@@ -458,18 +459,21 @@ run_region <- function(target_region,
                        return_estimates,
                        return_partial_estimates,
                        ...) {
-  futile.logger::flog.info("Initialising estimates for: %s", target_region)
+  futile.logger::flog.info("Initialising estimates for: %s", target_region, 
+                           name = "EpiNow2.epinow")
```
joeHickson, line 463: This string is being used a lot - perhaps tuck it onto a variable?

R/dist.R
```diff
@@ -430,7 +430,7 @@ bootstrapped_dist_fit <- function(values,  dist = "lognormal",
                                              future.scheduling = Inf,
                                              future.globals = c("values", "bootstraps", "samples",
                                                                 "bootstrap_samples", "get_single_dist"),
-                                             future.packages = "data.table")
+                                             future.packages = "data.table", future.seed = TRUE)
```
joeHickson, line 433: This seems unrelated to logging - is it intentional?

src/stanExports_estimate_infections.h
```diff
@@ -186,9 +186,9 @@ discretised_gamma_pmf(const int& y,
         stan::math::fill(beta, DUMMY_VAR__);
         stan::math::assign(beta,(mu / pow(c_sigma, 2)));
         current_statement_begin__ = 32;
-        stan::math::assign(alpha, (logical_lt(alpha, 0) ? stan::math::promote_scalar<local_scalar_t__>(1e-5) : stan::math::promote_scalar<local_scalar_t__>(alpha) ));
+        stan::math::assign(alpha, (logical_lte(alpha, 0) ? stan::math::promote_scalar<local_scalar_t__>(1e-5) : stan::math::promote_scalar<local_scalar_t__>(alpha) ));
```
joeHickson, line 189: I can't see what's triggered this change, can I just get a sanity check to confirm it's intentional (even if generated)?
< to <=

joeHickson (changes requested):

R/epinow.R
seabbs: the default logger name is EpiNow2. At some point, all nested functions should have a sensible hierarchy but that can be in a latter PR.

R/epinow.R
seabbs: the logger name?

R/dist.R
seabbs: It is - caught as I tested due to an update to `{future}`.

src/stanExports_estimate_infections.h
seabbs: yes - recompiled the stan code from a previous PR built by another author (and not compiled). Fixes a marginal issue at the boundary.

joeHickson: I dislike changes coming from other tasks but if it's good that's fine. I wonder if we could put a compile step on a github action to prevent this happening? (#lowpriority)

R/dist.R
joeHickson: Do we need to update the future dependency definition in DESCRIPTION with a minimum version to support future.seed requirement?

R/epinow.R
joeHickson: yes, "EpiNow2.epinow"

R/dist.R
seabbs: no its still works with older version but now throws a warning.

R/epinow.R
seabbs: can do - think save that for an extension to the logging system in a future PR.

R/dist.R
joeHickson: brill

joeHickson (approved):
LGTM

PR #99, opened 2020-10-06
Add progress bar

seabbs:
This PR adds support for `progressr` progress bars and implements a switch on if complete estimates should be written to the "EpiNow2" logger or the "EpiNow2.epinow" logging based on the number of regions (switch set at 10). (Note: Recompilation of the stan model here - no review needed on these changes).


R/epinow.R
```diff
@@ -351,14 +353,26 @@ regional_epinow <- function(reported_cases, target_folder, target_date,
...
+                                                target_date = target_date,
+                                                return_estimates = return_estimates,
+                                                return_partial_estimates = summary | return_estimates,
+                                                complete_logger = ifelse(length(regions) > 10, 
+                                                                         "EpiNow2.epinow",
+                                                                         "EpiNow2"),
+                                                progress_fn = progress_fn,
+                                                ...,
+                                                future.scheduling = Inf,
+                                                future.seed = TRUE)
```
joeHickson, line 372: that bug fix is getting everywhere!

joeHickson (approved):
Another big batch of stan noise but otherwise good

R/epinow.R
seabbs: I know!

PR #112, opened 2020-10-12
Living in log space

### LINKED ISSUE #90: Explore inverse gamma prior for gaussian process
May offer the ability to effectively truncate away from numerically unstable areas of the parameter space (lengthscale < observed  data and lengthscale > observed data). 

See: https://betanalpha.github.io/assets/case_studies/gaussian_processes.html

seabbs:
**Note: This change does increase the rate of transitions flagged as divergent. However, given all the evidence points towards this being a better prior choice I think this is still a sensible change.**

This PR closes #90 and shifts all Rt estimation into Log space. Results on test data set are comparable (as they should be) but it is hoped that shifting everything into log space will improve the numerical stability. 

The change to the lengthscale prior is to switch it from a half normal to an inverse gamma prior as this is broadly recommended [everywhere](https://betanalpha.github.io/assets/case_studies/gaussian_processes.html#43_Approximating_Gaussian_Processes). The current prior has been set based on upper and lower truncation (2 days and 21 days respectively) with 98% of the prior density between these values. This curtails overfitting at low lengthscales and reduces the flattenness of the likelihood at high lengthscales which makes fitting easier but perhaps a longer default might be sensible. 

In order to arrive at the prior values, I have implemented a new function `tune_inv_gamma` (with stan code of the same name) that is based on code in the link above.  This can also be used by the users/us to find other priors. The default is 2 and 21 days as this fits in a stable way - higher upper bounds fail. 

Future PRs will focus on the code styling and robustness with a view to extending the complexity of the model.

Updated version: 

Screenshot 2020-10-12 at 17 19 49

Compared to original (I know it is too squat):


R/estimate_infections.R
```diff
@@ -42,8 +42,10 @@
 #'  additional compute. Must also contain the `boundary_scale` (multiplied by half the range of the input time series). Increasing this 
 #' increases the accuracy of the approximation at the cost of additional compute. 
 #' See here: https://arxiv.org/abs/2004.11408 for more information on setting these parameters.
-#' Can optionally also contain the  `lengthscale_mean` and `lengthscale_sd`. If these are specified this will override 
-#' the defaults of 0 and 2 (normal distributed truncated at zero).
+#' Must also contain the  `lengthscale_alpha` and `lengthscale_beta`. These tune the prior of the lengthscale. Principled 
+#' values can be obtained using `tune_inv_gamma` which optimises based on the desired truncation (which should be based on the scale
+#' of the observed data). The default is tuned to have 98% of the density of the distribution between 2 and 21 days. Finally the list must
```
jhellewell14, line 47: By scale of the observed data, do you mean something along the lines of the length and frequency of the time series?

R/dist.R
```diff
@@ -560,3 +560,37 @@ sample_approx_dist <- function(cases = NULL,
...
+#' 
+#' tune_inv_gamma(lower = 2, upper = 21)
+tune_inv_gamma <- function(lower = 2, upper = 21) {
+  
+  model <- stanmodels$tune_inv_gamma
+  # optimise for correct upper and lower probabilites
+  fit <- rstan::sampling(model, 
+                         data = list(u = upper,
+                                     l = lower),
+                         iter = 1,
```
jhellewell14, line 583: (∩ ` -´)⊃━━☆ﾟ.*･｡ﾟ

inst/stan/tune_inv_gamma.stan
```diff
@@ -0,0 +1,41 @@
...
+    
+  real delta = 1;
+  real a = square(delta * (u + l) / (u - l)) + 2;
+  real b =  ((u + l) / 2) * (square(delta * (u + l) / (u - l)) + 1);
+  vector[2] y_guess = [log(a), log(b)]';
+  
+  real x_r[0];
+  int x_i[0];
+  
+  vector[2] y = algebra_solver(tail_delta, y_guess, theta, x_r, x_i);
```
jhellewell14, line 35: (⁎⚈᷀᷁ᴗ⚈᷀᷁⁎)

jhellewell14 (approved):
– ̗̀ 𝓗𝔞𝔱𝔢𝔯𝔰 𝔤𝔬𝔫𝔫𝔞 𝔥𝔞𝔱𝔢 ̖́- ᕕ( ᐛ )ᕗ✧

R/estimate_infections.R
seabbs: Yes exactly. If the lengthscale is > the length of the timeseries then moving around it doesn't change the fit so the MCMC chain will get stuck. For lengthscales smaller than the smallest observed increment your essentially interpolating (though how this interacts with a latent process I don't know).

Issue #111, opened 2020-10-12
Heuristic Data Cleansing


hsbadr:
It's important to add user options to address data quality issues. Initially, this may include handling [missing values (`NA`s)](https://github.com/epiforecasts/EpiNow2/issues/75) and [negative new daily cases](https://github.com/epiforecasts/covid-rt-brazil-subregional/issues/1), by adding a new argument `clean` with a list of options as follows:
```
clean = list(
  na = 0,
  negative = 0
)
```
where `na` and `negative` are integers:
- `na = 0` replaces `NA`s with zeros
- `na > 0` for forward filling
- `na < 0` for backward filling
and
- `negative = 0` replaces negative values with zeros
- `negative > 0` for forward spread of negative values
- `negative < 0` for backward spread of negative values

This is a general version of #75.

PR #113, opened 2020-10-12
Deal with random zeros

seabbs:
This PR adds a heuristic to deal with reporting issues. Currently, a day with NA cases is treated as zero and this can cause model fitting issues. This PR adds a simple heuristic to detect and deal with these occurrences (tests the 7 day rolling average to see if cases are above 50 per day. If this is the case then it sets any detected zeros to be the same as the rolling 7 day average). Obviously, this approach isn't perfect and there are multiple edge cases (like 4 days without cases in a row) which will cause issues.

In a later PR it would be sensible to alter the internal model structure so that these issues are excluded from the likelihood rather than being fit at all. There is probably a lot more thinking needed here (i.e see #111).


jhellewell14 (approved):
(づ｡◕‿‿◕｡)づ

PR #116, opened 2020-10-14
R CMD check NOTEs

hsbadr:
This fixes the following NOTE:
```
* checking dependencies in R code ... NOTE
Namespace in Imports field not imported from: 'rstantools'
  All declared Imports should be used.
```

Other NOTEs to fix:
- It's recommended to edit `src/Makevars` and `src/Makevars.win` and remove `SystemRequirements: GNU make` from the package `DESCRIPTION`:
```
* checking for GNU extensions in Makefiles ... NOTE
```

- The package size grows, depending on the compiler and optimization flags:
```
* checking installed package size ... NOTE
  installed size is 14.6Mb
  sub-directories of 1Mb or more:
    help   2.3Mb
    libs  11.7Mb
```


seabbs (approved):
Looks good to me - thanks! and good points on the other notes.

hsbadr:
@seabbs Do you plan to release `EpiSoon` to CRAN soon? If not, I'd remove it from the `EpiNow2` dependencies.
```
checking package dependencies ... NOTE
Package suggested but not available for checking: 'EpiSoon'
```

seabbs:
Not in the near term. I think the CRAN check will throw an error if it is not included in the suggests as it is used in the code. There is a long term plan to move much of the infrastructure used into EpiNow2 by reusing the stan code (which will be in the next version). 

Having said that I am very happy to see it dropped if it can be done in a way that passes a CRAN check.

PR #118, opened 2020-10-15
R-CMD-check: NOTE: Remove 'SystemRequirements: GNU make'

hsbadr:
This fixes the following NOTE:
```
* checking for GNU extensions in Makefiles ... NOTE
```
and makes `R` makefiles (`Makevars` and `Makevars.win`) portable.

It also defines `_REENTRANT` to make reentrant version `lgamma_r` available from `cmath`. This may not be required for the current stable version of `rstan`/`stan`/`math`, but it's for the development versions. It has no side effects.


hsbadr:
@seabbs Can you review and merge this? I extensively tested it with different `rstan`/`stan`/`math` versions on all OSs.

seabbs:
Hi @hsbadr, 

Sorry for the delay on this. Technically on AL. Could you provide some brief details on why the previous set up was not good and why this one is better. I am afraid this is a little over my head without doing some reading.

hsbadr:
Hi @seabbs,

No problem. Enjoy your AL!

As I explained in the PR message, this basically converts the makefiles to a portable format. Both the old and new makefiles generate the exact same compiler flags, except that the new version doesn't depend on GNU make, as preferred by CRAN.

The 2nd commit defines `_REENTRANT` (by adding `-D_REENTRANT` to `CXXFLAGS`) to make reentrant version `lgamma_r` available from `cmath`. I needed this to fix the compilation of `EpiNow2` with development version of `rstan`/`stan`/`math`.

seabbs:
Awesome thanks - all sounds good and reasonable to me. Good to hear about the portability testing.

PR #125, opened 2020-10-17
Explore initialisation of infectiousness

seabbs:
As discussed with @jhellewell14 this PR explores moving from initializing infections prior to estimate Rt (only for unobserved infections) and instead initializes a single prior on the first summed infectiousness. 

There are two complexities here: 1. How indexing works in terms of what is reported and what is generated (because we have unobserved and observed time with unobserved time mapping to observed time). and 2. If initial infectiousness is correctly shared across early Rt estimates. 

This PR is a work in progress and currently, indexing is not correct and infectiousness is not properly updated for early Rt estimate .


seabbs:
Closing as done in #144

PR #144, opened 2020-10-29
Update initialisation of Rt infections

### LINKED ISSUE #120: Update initialisation of infections
1. Start with historic infectiousness need to think about a prior for this that will scale to any case load (so we can jumpstart the model in the middle of a timeseries). Shifted reported cases currently fills this role. Likely best to use an exponential prior. 
2. Estimate Rt
3. Decide when to start estimating Rt (either when reported data starts or earlier). Current default is when reported data has started.

seabbs:
This PR closes #120 and is based on discussion with @jhellewell14 and others. 

It updates the intialisation of infections when estimating Rt by using a single normal prior for log initial infections (with the logmean set as the log of the mean of cases observed in the first 7 days and the standard deviation as 10% of this). It then generates new infections over the period of the seeding time using a log linear model with a daily growth rate parameter (which itself has a standard normal prior). Once seeding time is complete (which is the mean of all input delays) then Rt estimates are updated as previously. 

Testing indicates some small increase in fit quality along with a slight reduction in fitting time. `tests/testthat/test-stan-infections.R` gives examples of the internal functions behaviour. 

PR also: 

* reporting overdispersion to returned output
* stan fit returned as default for `epinow` and `estimate_infections`. 

Note later PRs will look at making accessing key parameters easier.


seabbs:
Merging for now happy to discuss updating in another PR in an issue.

PR #147, opened 2020-10-31
stan: negbin: defer to poisson if phi is large

### LINKED ISSUE #143: `estimate_infections.stan` occasionally throws exceptions with large numbers in the negative binomial
This likely indicates an issue with the model set up and may be linked to a stan code bug. Alternatively it may just be due to model complexity and controls within the stan code should be implemented to deal with it.

hsbadr:
Defer to poisson distribution if the reciprocal overdispersion parameter
of the negative binomial distribution (phi) is large, to avoid overflows.

See https://github.com/stan-dev/math/pull/466

Fixes #143


seabbs (approved):
This all looks good and makes sense - thanks. I tuned down the switch to 1e4 as still saw a failing test with this issue. Hopefully can get to the root cause sooner rather than later.

PR #148, opened 2020-10-31
Convolve PRs and then map infections to reports once

seabbs:
This PR explores convolving all reporting delays into a single PMF and then using this to convolve reports to infections. The current master approach is to treat each report delay on its own and iteratively map reports to a latent variable. Unfortunately the approach implememented in this PR is numeric as an analytic solution for log normals does not appear to exist. If this approach was shown to be sufficiently worthwhile however it might be worth switching delay support to a distribution for which an analytic definition does exist (i.e Gamma). 

The approach implemented in this PR may or may not be numerically more stable but on test data does not appear to result in greatly improved performance (fitting is somewhat stochastic so there may be some improvement/cost that is masked by this). Potentially the convolution of the PMFs can be further optimized but it is likely that the majority of the compute time is the dot product of the convolution from infection to report (and hence repeating this iteratively or doing a larger one all at once may have a minimal impact). 

Prior to merging this implementation needs further testing on longer time-series to further explore whether it introduces any benefit. Ideally, the implementation of the convolution also needs review + and opinions shared on the advantages/disadvantages of this approach. Personally, I would default to the master branch iterative implementation in the absence of evidence as the implementation in this PR is more mathematically complex and may make understanding the implementation more difficult. Happy to hear opinions on this however.


R/estimate_infections.R
```diff
@@ -82,7 +82,7 @@
 #'                            stan_args = list(warmup = 200,
 #'                             control = list(adapt_delta = 0.95, max_treedepth = 15),
 #'                             cores = ifelse(interactive(), 4, 1)),
-#'                             verbose = interactive())
+#'                            verbose = interactive())
```
hsbadr, line 85: The old formatting was ok.

inst/stan/functions/generated_quantities.stan
```diff
@@ -14,7 +14,7 @@ int[] report_rng(vector reports, real[] rep_phi, int model_type) {
   int sampled_reports[t];
   if (model_type) {
     for (s in 1:t) {
-      sampled_reports[s] = neg_binomial_2_rng(reports[s] > 1e8 ? 1e8 : reports[s], rep_phi[model_type]);
+      sampled_reports[s] = neg_binomial_2_rng(reports[s] > 1e7 ? 1e7 : reports[s], rep_phi[model_type]);
```
hsbadr, line 17: I don't think that the `reports` causes the overflows/exceptions and `1e8` is well bellow `.Machine$integer.max`. I believe that `rep_phi` is the culprit (see #147).

R/estimate_infections.R
seabbs: It was actually out by a space (due to the stan_args secondary list). I went through and checked this (very fun job) so what is on master should now be correctly formatted.

inst/stan/functions/generated_quantities.stan
hsbadr, line 17: It'd be cleaner to define the upper and lower bounds globally so that it'll be consistent everywhere and could be easily tuned. I'm talking about `small = 1e-5` and `large = 1e8`, for example, or whatever a global variables will be named.

hsbadr: All global constants could be moved to a header file.

seabbs:
The approach used in this PR has been implemented in #261  so closing.

Issue #133, opened 2020-10-20
Allow forecasts to be generated from estimates


sbfnk:
It would be good to allow R trajectories (or samples thereof) and initial infections as an input that re-simulates infections and reported cases using these trajectories. This would make it possible to, e.g. explore "expert adjustments" to Rt trajectories or scenarios.

seabbs:
Agree also needed for validation work. This so currently managed via the simulate functions code but that is an R implementation that doesn't capture the full model complexity. The Stan refactor (#122) should enable this and as this is the next job on the list may happen fairly quickly.

sbfnk:
Isn't this functionality already provided by `forecast_infections`? It would be more robust for this to be handled stan-internally but it seems the functionality is there.

seabbs:
yes it is (and/or the internal functions).

PR #150, opened 2020-11-02
Simulate infections

sbfnk:
initial implementation of within-stan infection simulation, to explore scenarios and forecasting methods without a dependency on parallel R code in `EpiSoon`. Addresses https://github.com/epiforecasts/EpiNow2/issues/133.

This has resulted in some duplicate stan code which could probably be improved by slight changes to the function/data syntax, and also some inelegant calls to functions that were set up to interact with `estimate_infections`.

It works with the example and would allow response to recent requests from policy makers.


seabbs (commented):
This looks great. It is along the lines of what I was thinking we should be doing for all the forecasting so at some point it would be nice to move everything into this function that gets generated in future time. 

The only technical issue with this is the covariance matrix for the gaussian process but think this can just be recalculated with the posterior lengthscale and magnitude. 

Just waiting on #144 and then will merge (after a local walkthrough).

seabbs:
Merged the initialisation in from master and worked through all looks good to me. It is quite a bit slower than I would hope though - perhaps can be made parallel at some point. 

I think it would be a good idea to pull out as much as possible from estimate_infections into simulate_infections and start doing estimation/forecasting in separate steps (opening issue). What do you think?

seabbs (approved):
:+1:

PR #156, opened 2020-11-03
Speed up simulate infections

seabbs:
The new `simulate_infections` appears to have a massive IO cost from moving around data. For a real use case (UK scenarios) run time is at least 6 hours (still running). 

This PR attempts to reduce this cost and hence speed up the run times. It does this in two ways 1. by only passing samples that are needed to `simulate_infections.stan` and 2. by running samples in batches (with the user also now able to control the number of posterior samples used). 

Additionals features of this PR are: 

* Added a progress bar for `simulate_infections`
* Added parallel support for `simulate_infections`.
* Depreciates `burn_in` arg as no longer needed with new initialisation. 
* Adds support for `burn_in` to `simulate_infections` whilst still present.

Timing indicates that using a batch size of 10 speeds up the example from 9 minutes to 10s . I don't clearly understand why this is the case and perhaps there is a mistake.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,7 @@
 * Reworked the package logging system to improve the reporting of issues both in `epinow` and in `regional_epinow` for large batch runs.
 * Fix from @hsbadr to prevent overflow when overdispersion is larger (by switching to a Poisson approximation). Hitting this issue may indicate a bug in other model code that will need further work to explore.
 * Moved default verbosity for all functions (excepting `regional_epinow`) to be based on whether or not usage is interactive. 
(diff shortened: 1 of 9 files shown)


seabbs:
I still can't find what is causing this but it is definitely only here that I see this issue (and not in the main model etc). I think this should be merged as the main user (me) needs to be able to use this for UK scenarios and I don't think it is harmful (outside a presumably small increase in runtimes from looping everything). Using the new default gives a runtime of < a minute as expected on a longer run. 

In the latest commits I added support for the `burn_in` arg which was causing issues in the UK and depreciated that arg as a feature (as we can now stably estimate Rt as long as there are delays present due to the new initialisation approach).

seabbs:
Merging for now and opening issue #157 to look at as and when.

PR #164, opened 2020-11-10
Adds future susceptible population adjustment

seabbs:
This PR adds future susceptible population adjustment (with estimated Rt already being adjusted by definition). The remaining to do are:

* Review approach + theory. Approach ultimately implemented includes a scaling by infectiousness as without this more infections than people was possible. Based on 2.1 -> 2.4 [here](https://imperialcollegelondon.github.io/epidemia/articles/introduction.html). Also includes an ifelse gate to check the adjustment and scale to 0 if result would lead to negative cases (seen during sampling and likely due to the Rt already adjusted correction).
* Review implementation
* Check testing
* Work out how to pass the required parameters from `regional_epinow` to `estimate_infections`. Likely requires surfacing all `estimate_infections` args in `epinow` and `regional_epinow` and then passing an additional `region_opts` list to override defaults at the `run_region` stage. Open to other ideas as this is quite clunky.

Example implementation is in `estimate_infections` and produces the following output: 

Screenshot 2020-11-10 at 00 45 32


seabbs (commented):
👍

PR #168, opened 2020-11-13
Add support for truncation

seabbs:
This PR adds support for in model truncation. This is likely useful in the UK where truncation of data is common and the current approach is a filtering step. The implementation works as follows (with complete steps checked): 

- [x] Fit a truncation distribution to a data frame containing repeat observations around a target date (with the expectation this should be extended to n target dates in the future). Fitting here is something along the lines of `earliest date = NB(convolve(later date, reversed cmf))`.
- [x] Input the uncertain truncation distribution in the same log normal framework used for delays (i.e `trunc_opts`).
- [x] Use the fit from step 1 as the prior distribution for truncation (or skip if not present (the default)). This is the same step as used for reporting delays and the fraction observed and requires a heavily informed prior (as there is unlikely to be information in the data).
- [x] Apply the truncation distribution as a reversed CMF to the head of observed data after all other steps are complete. 
- [x] Add support to `simulate_infections`. **Not added as not clear how (as meaning changes) or what it would be used for if present. Open for discussion.**
- [x] Add evaluation plot
- [x] Make usable from `regional_epinow`
- [ ] Test on real (UK) data

The limitations/assumptions of this approach are as follows: 

- Current truncation is related to past truncation. 
- Truncation is a multiplicative scaling of underlying reported cases. 
- Truncation is log normally distributed. Other distributions are possible but supporting multiple may increase code complexity.  
- Truncation is fixed across all regions called in `regional_epinow`. This is an interface limitation that can be resolved when/if the interface for `regional_epinow` is updated to support between regions differences (this also currently prevents population adjustment being used at this level as well).

Example from the docs for `estimate_infections`: 

Screenshot 2020-11-13 at 16 04 40

Unrelated to this PR model investigation indicates that the posterior of the generation time and reporting delays are tighter (between 2 and 10 times) than the priors. This may indicate a bug with the weighting, the need for a rethink on what that weighting should be (current weighting is the number of time points which has some level of theory underpining it), or a sign something has changed in the model. Alternatively, we may be happy with this behaviour as it may indicate fitting based on some signal. This can be seen [here](https://raw.githubusercontent.com/epiforecasts/covid-uk/master/format-rt/figures/latest.png) where the generation time has an apparent SD ~ 0.05 but the input has an SD of 0.7.  For the testing examples this may be due to having too few samples by default but in the UK instance (with 4000 samples) this is unlikely ot be the case. 

Other changes: 

- Moves Rt priors into a function in order to shorten the model.
- Increases the number of default samples to 2000 from 1000 as 1000 is never enough and giving people this as a default seems like a bad idea.
- Fixes 0 lower bound on delays. In practise this should not have had an impact on any of our usage but it would have caused issues with very short delays (likely an error message and stan related crash).
- After discussion with @kathsherratt this PR adds support for region specific settings in `regional_epinow` for all aguments that accept a `_opts` function. This is done via a named list of `_opts` calls (with names = regions). To help construct this list two new helper functions have been added `opts_list` and `update_list` along with an example of their use in the documentation for `regional_epinow`. 

(@kathsherratt pinged to look at the interface changes if of interest)


seabbs:
Output from new `plot.estimate_truncation` showing the recovery of final observations (bars) from observations at the time (dots) with estimates shown as a ribbon. Note on best case simulation data + final plot "final" observations are truncated. 

Screenshot 2020-11-16 at 14 40 42

seabbs:
Merging - happy to PR changes in later. `estimate_truncation` marked as experimental so safe to add I think.

PR #170, opened 2020-11-14
explore scores/fits of R and infections

sbfnk:
Some initial look at evaluating the performance of the models against reported cases, infections and R


seabbs (approved):
:+1: Exciting stuff!

PR #172, opened 2020-11-17
mild clean up in report.R for summary plots function

pearsonca:
Fixes* some of the code noise and fixes* documentation error.

*did via github interface, hence the silly commit history and also not confirming that everything still runs. definitely >90% fixes the problems


seabbs (approved):
Great stuff thanks @pearsonca !

PR #176, opened 2020-11-18
make process_region work if out == NULL

### LINKED ISSUE #175: error in `process_output` if `out == NULL`
This line
https://github.com/epiforecasts/EpiNow2/blob/8eccc543da4e2791d0467b0b7cb97639a51ec2b0/R/regional_epinow.R#L345
fails with `Error in as.environment(where) : using 'as.environment(NULL)' is defunct` if out is `NULL` (e.g., if `return_output==FALSE`).

The same issue should affect subsequent lines. This occurs in https://github.com/epiforecasts/covid-rt-estimates/pull/96#issuecomment-729923329

sbfnk:
Fixes #175 (I think) - untested.


seabbs:
Ah makes sense. So I thought this should never be triggered as we always need there to be output (to get timings) but looking at https://github.com/epiforecasts/EpiNow2/blob/7d1883342cf761776667f0cef14af6d59cc1bb6c/R/regional_epinow.R#L148 this is not the case when not generating a summary (as just changed in the `covid-rt-estimates`. 

Nice spot! Not just on top of the forecasting leaderboard....

sbfnk:
Something went wrong in the last commit, I think...the first change.

PR #185, opened 2020-11-25
Enable stanfit object to be used to initialise `estimate_infections`.

seabbs:
This PR:

* Adds a new argument `init_fit` to `stan_opts()` that enables the user to pass in a `stanfit` to use to initialise a model fit in `estimate_infections()`. Optionally `init_fit = "cumulative"` can also be passed which first fits to cumulative data and then uses the result to initialise the full fit on incidence data. This approach is based on the approach used in [epidemia](https://github.com/ImperialCollegeLondon/epidemia/) authored by James Scott. Currently `stan` warnings from this initial fit are broadcast to the user and may cause concern as the short run time and approximate settings often lead to poor convergence. 

Currently found no benefit to either of these approaches but testing is need on problematic data (i.e the UK) which I will do shortly. Potentially initialising from fits has other use cases but care would be needed here as all parameters are currently extracted some of which should potentially not be shared across fits (i.e the starting seed cases for example).


seabbs (commented):
Well done me

PR #186, opened 2020-11-26
stan: Properly clip/limit the values of the parameters

hsbadr:
Make sure that all values of the parameters bellow a prespecified lower bound (`small = 1e-5`) or above a prescribed upper bound (`large = 1e9`) are clipped. Otherwise, the values in the range `0 - 1e-5` or `1e9 - Inf` are not limited.

For example, with `alpha = alpha <= 0 ? 1e-5 : alpha;`:
* if `alpha <= 0` → `alpha = 1e-5`
* if `alpha = 1e-500` → `alpha = 1e-500 < 1e-5`
* if `alpha = 1e-6` → `alpha = 1e-6 < 1e-5`
* if `alpha >= 1e-5` → `alpha >= 1e-5`

So, `0 < alpha < 1e-5` won't be clipped meaning that if `alpha = 1e-500`, `log(alpha) ~= -Inf`.

This PR fixes this behavior by clipping all values bellow the lower bound and above the upper bound.

### DIFF
--- a/inst/stan/functions/pmfs.stan
+++ b/inst/stan/functions/pmfs.stan
@@ -4,14 +4,17 @@ vector discretised_gamma_pmf(int[] y, real mu, real sigma, int max_val) {
   vector[n] pmf;
   real trunc_pmf;
   // calculate alpha and beta for gamma distribution
-  real c_sigma = sigma + 1e-5;
-  real alpha = ((mu)/ c_sigma)^2;
-  real beta = (mu) / (c_sigma^2);
-  //account for numerical issues
-  alpha = alpha <= 0 ? 1e-5 : alpha;
-  beta = beta <= 0 ? 1e-5 : beta;
-  alpha = is_inf(alpha) ? 1e8 : alpha;
-  beta = is_inf(beta) ? 1e8 : beta; 
+  real small = 1e-5;
+  real large = 1e9;
+  real c_sigma = sigma < small ? small : sigma;
+  real c_mu = mu < small ? small : mu;
+  real alpha = ((c_mu) / c_sigma)^2;
+  real beta = (c_mu) / (c_sigma^2);
+  // account for numerical issues
+  alpha = alpha < small ? small : alpha;
+  alpha = alpha > large ? large : alpha;
+  beta = beta < small ? small : beta;
+  beta = beta > large ? large : beta;
   // calculate pmf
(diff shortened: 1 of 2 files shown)


inst/stan/functions/pmfs.stan
```diff
@@ -4,14 +4,17 @@ vector discretised_gamma_pmf(int[] y, real mu, real sigma, int max_val) {
...
+  real alpha = ((c_mu) / c_sigma)^2;
+  real beta = (c_mu) / (c_sigma^2);
   //account for numerical issues
-  alpha = alpha <= 0 ? 1e-5 : alpha;
-  beta = beta <= 0 ? 1e-5 : beta;
-  alpha = is_inf(alpha) ? 1e8 : alpha;
-  beta = is_inf(beta) ? 1e8 : beta; 
+  alpha = alpha < small ? small : alpha;
+  beta = beta < small ? small : beta;
+  alpha = is_inf(alpha) ? large : alpha;
```
seabbs, line 16: Presumably the same logic applies here?

seabbs (changes requested):
Thanks for this @hsbadr - very sensible and a good catch. Can you add a small note to the news and see my comment on the treatment of large values (probably needs updating in the same way).

inst/stan/functions/pmfs.stan
hsbadr: Sounds good! Check f9a3226a157894b03db928fa1886d1a675e471ad !

hsbadr:
@seabbs With cases getting close to ~`100` million globally and a world population of ~ `7.8` billion, we probably need to increase the upper bound of `1e8` and check the limits everywhere.

seabbs:
The limit should be on daily cases though so probably not a possible situation?

hsbadr:
I don't mean in this specific file; it has nothing to do with cases. I'm just suggesting to check the limits of all variables everywhere because chains get stuck randomly probably in a wrong parameter space, and I see `1e8` in many places such as https://github.com/epiforecasts/EpiNow2/blob/e1a4577beb26910dc4969ebfbd37fef1ccb55576/inst/stan/functions/generated_quantities.stan#L59.

seabbs:
Yes I am very keen to resolve the chain sticking issue.

seabbs (approved):
Looks good now!

PR #190, opened 2020-11-28
Constrain initialisation and extreme areas of the parameter space

seabbs:
This PR expands and reviews the restrictions placed on initialisations in #0dce73b. It also adds boundary restrictions to prevent the mean generation time being greater than the maximum specified generation time and additional restrictions to reject if the standard deviations of parameters are less than 0.


seabbs:
@hsbadr yes potentially but I am not sure which options shouldn't be exposed (though with sensible defaults of course) in the current interface. The upper limit on R in this PR should only really come into play if the prior is extremely misspecified which I would hope would not be the case. Open to making this user settable though but I think perhaps it could be an option in `rt_opts`.

PR #202, opened 2020-12-11
Refactor of rt updating

seabbs:
Refactor of `update_rt` (an internal `stan` function found in `inst/stan/functions/rt.stan`) to be vectorised. This change reduces run times by approximately 20 ~ 40% (though only tested on a small subset of examples) and opens the way for future model extensions (such as additive rather than multiplicative random walks, introducing covariates etc).


inst/stan/functions/rt.stan
```diff
@@ -1,72 +1,36 @@
...
   }
-  // iteratively update Rt
-  for (s in index:t) {
-    R[s] = update_R(R, noise, noise_terms, s, i_stationary);
-    if (bp_n > 0) {
-      at_bp = bps[s];
-      bp_in += at_bp;
-      R[s] = update_breakpoints(R[s], bp_effects, bp_in, at_bp, i_stationary);
+  //initialise gaussian process
+  if (gp_n) {
```
sbfnk, line 24: suggest to settle on either `if (x)` or `if (x > 0)` and stay consistent.

sbfnk (approved):
All looks sensible to me.

inst/stan/functions/rt.stan
seabbs: yes good point - should really just be `if (x)` everywhere.

PR #207, opened 2020-12-15
change labels and labelling for increasing/decreasing/stable

sbfnk:
Suggestion for geting rid of "unsure" (to replace with "stable") as e.g. we could be really really sure that R=1. Also suggest a change in label assignment.

These are very much up for discussion.


sbfnk:
Interesting - why did you think "stable" could be misinterpreted (as R~1 does imply stable incidence / prevalence)?

sbfnk:
PS: really nice Lancet ID piece - hadn't seen that.

seabbs:
Yes that is nice 👍 

I like stable as it does correctly describe what we would expect to happen to cases in the short term. I think there are two minor issues. 1. Rt might be rapidly oscillating above and below 1 so it could be not stable over time and 2. People might misunderstand and think its a good category to be in (which it may or may not be depending on case load).

seabbs (approved):

seabbs:
Merging in for now (as an improvement over master) but open to further discussion of a better option.

hsbadr:
> Interesting - why did you think "stable" could be misinterpreted (as R~1 does imply stable incidence / prevalence)?

Building on @seabbs comment: "`stable`" implies a stabilizing pattern where the slope is ~zero, but this can be misleading especially if it's rapidly oscillating over a period of time. Let's say that we project the number of cases over 4 weeks and both the 1st and last weeks have the same number of cases `N` while in the 2nd and 3rd week the cases were fluctuating in a wide range `N * (1 ± ε)` where `ε` can be a large proportion, do you consider this "`stable`"? Same for `Rt` estimates. It depends on how you define the increasing/decreasing threshold, if any, and if you account for the temporal variability.

BTW, @seabbs: `EpiNow2` forecasts are extreme in many cases, especially for the units with relatively small number of infections. It seems that it's very sensitive to the most recent trend and doesn't learn from the historical pattern (timeseries memory).

medewitt:
Building on @hsbadr 's point regarding extreme for small units - I've seen this, too (needs some integral gain to dampen this effect). Weird reporting anomalies can be lumped into this as well (where no cases are reported Fri-Sun, the a dump of cases on Monday, but this isn't consistent so it is missed by the delay convolution and the day of week effect). No fixes in mind, but just wanted to pass as example.

PR #210, opened 2020-12-28
Test PR commands

seabbs:
This PR tests moving package styling (i.e. stan/C++ checking from @hsbadr and R checks from `styler`) into PR commands run using GitHub actions.


seabbs:
Testing new PR commands to remove whitespace, style and redocument (/style) and to style C++ code (/cpp).

seabbs:
retest /style

hsbadr:
@seabbs You may also want to restore the automatically-generated `configure` script:
```
# Generated by rstantools.  Do not edit by hand.

#! /bin/sh
"${R_HOME}/bin/Rscript" -e "rstantools::rstan_config()"
```
`rstantools` generally checks if the 1st line includes `Do not edit` before updating the file.

.github/workflows/pr-commands.yaml
```diff
@@ -53,9 +53,9 @@ jobs:
       - uses: r-lib/actions/pr-push@v1
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-  style-stan:
-    if: startsWith(github.event.comment.body, '/style-stan')
-    name: style-stan
+  stan:
+    if: startsWith(github.event.comment.body, '/stan')
+    name: stan
     runs-on: ubuntu-16.04
```
hsbadr, line 59: May need to add `sudo apt install clang-format`.

.github/workflows/pr-commands.yaml
```diff
@@ -53,9 +53,9 @@ jobs:
       - uses: r-lib/actions/pr-push@v1
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-  style-stan:
-    if: startsWith(github.event.comment.body, '/style-stan')
-    name: style-stan
+  stan:
```
hsbadr, line 56: I'd name it `format` instead of `stan` as it can be used for C/C++ code formatting in any project.

hsbadr: Also, we may use `ubuntu-20.04` instead of `ubuntu-16.04` in all workflows.

seabbs: I updated this to cpp on master (needed to do these changes on master in order to test the GitHub Actions (I am not aware of a better way to test GAs))

.github/workflows/pr-commands.yaml
seabbs: good point

.github/workflows/pr-commands.yaml
hsbadr: Why did you add it to `pr-commands.yaml` workflow? I'd test with `on push` events for easy and quick triggers. I'm testing a `Lint-Format.yaml` workflow [here](https://github.com/hsbadr/HiClimR).

PR #212, opened 2020-12-30
Update GHA workflows for code linting and formatting

hsbadr:
This adds `Lint-Format` GHA workflow for code linting and formatting.


hsbadr:
@seabbs These workflows should work as expected now:
- `Lint` for code linting (check the warnings if you'd like to fix them).
- `Format` for R and C/C++ code cleanup and formatting.
- `Render` for rendering the documents.

The `Render` workflow takes a long time for `EpiNow2` and could be disabled.

seabbs:
Thanks for this @hsbadr  looks good to me.  Just a few clarifying qs. 

1. Does this trigger on every commit to a PR (that isn't escaped using "[ci skip]" in the commit message). If that is the case will reformatting of the C++ code lead to the contributor having to recompile their stan code locally? I am not sure that is a problem just thinking through the workflow. 

2. Is there a way to skip the `inst` folder as that will likely continue to contain some less than ideal code.

hsbadr:
> 1. Does this trigger on every commit to a PR (that isn't escaped using "[ci skip]" in the commit message).

Yes.

> If that is the case will reformatting of the C++ code lead to the contributor having to recompile their stan code locally? I am not sure that is a problem just thinking through the workflow.

No. The `Format` workflow does code formatting and commit the changes, if needed, while the `Lint` workflow applies `lintr::lint_package()` to all `R` files. We can customize the linters and other configs by creating `.lintr` file.

> 2. Is there a way to skip the `inst` folder as that will likely continue to contain some less than ideal code.

Yes. We can exclude directories and customize the linters in `.lintr` config file, like:
```
linters: with_defaults(
    line_length_linter = line_length_linter(120)
  )
exclusions: list(
  "inst",
  "vignettes"
 )
```

seabbs:
Excellent - thanks for clearing those points up (you just beat me to adding the `.lintr` 😆).

seabbs (approved):
👍

hsbadr:
> you just beat me to adding the `.lintr` 😆.

You've the opportunity to fix the warnings [here](https://github.com/epiforecasts/EpiNow2/actions/runs/464045357), disable the related linters, or exclude files or subdirectories (e.g., `"tests"` - or call from the namespace like `testthat::expect_equal`) 😄

hsbadr:
@seabbs Here's the summary of the current linting warnings:
```
linters: with_defaults(
    line_length_linter = NULL, # 98
    trailing_whitespace_linter = NULL, # 11
    object_usage_linter = NULL, # 8
    object_name_linter = NULL, # 6
    commented_code_linter = NULL, # 3
    seq_linter = NULL, # 1
    single_quotes_linter = NULL, # 1
    trailing_blank_lines_linter = NULL, # 1
    dummy_linter = NULL
  )
```
If you add this to the `.lintr` config files, all current warnings will be disabled. But, I think we need to address some of them.

PR #213, opened 2020-12-30
Update growth calculation

seabbs:
This updates the output growth rate to be directly calculated from infections (r = log(inf_(t-1)) - log(inf_t)) rather than via the growth rate approximation using Rt. 

1. Adds `calculate_growth` and tests
2. Reduces growth rates output to be number of observations minus 1 (as can't be calculated for the first time step).

Everything looks as I would expect when unit testing and running the examples expect for this slightly odd artefact in the forecast portion that doesn't show up in the Rt or infections plot. 

Screenshot 2020-12-30 at 20 18 25

### DIFF
--- a/R/extract.R
+++ b/R/extract.R
@@ -119,7 +119,7 @@ extract_parameter_samples <- function(stan_fit, data, reported_dates, reported_i
   out$growth_rate <- extract_parameter(
     "r",
     samples,
-    reported_dates
+    reported_dates[-1]
   )
   if (data$week_effect == 1) {
     out$day_of_week <- extract_parameter(
--- a/inst/stan/estimate_infections.stan
+++ b/inst/stan/estimate_infections.stan
@@ -120,19 +120,16 @@ generated quantities {
   vector[estimate_r > 0 ? 0: ot_h] gen_R;
   real r[ot_h];
   vector[ot] log_lik;
-  if (estimate_r){
-    // estimate growth from estimated Rt
-    r = R_to_growth(R, gt_mean[1], gt_sd[1]);
-  }else{
+  if (estimate_r == 0){
     // sample generation time
     real gt_mean_sample = normal_rng(gt_mean_mean, gt_mean_sd);
     real gt_sd_sample = normal_rng(gt_sd_mean, gt_sd_sd);
     // calculate Rt using infections and generation time
     gen_R = calculate_Rt(infections, seeding_time, gt_mean_sample, gt_mean_sample,
                          max_gt, rt_half_window);
-    // estimate growth from calculated Rt
-    r = R_to_growth(gen_R, gt_mean_sample, gt_sd_sample);
   }
+  // estimate growth from infections
+  r = calculate_growth(infections, seeding_time);
   // simulate reported cases
   imputed_reports = report_rng(reports, rep_phi, model_type);
   // log likelihood of model
--- a/inst/stan/estimate_infections.stan
+++ b/inst/stan/estimate_infections.stan
@@ -118,7 +118,7 @@ model {
 generated quantities {
   int imputed_reports[ot_h];
   vector[estimate_r > 0 ? 0: ot_h] gen_R;
-  real r[ot_h];
+  real r[ot_h - 1];
   vector[ot] log_lik;
   if (estimate_r == 0){
     // sample generation time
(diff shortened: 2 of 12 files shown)


sbfnk:
Makes sense. Is this with forecasts keeping R constant? I could possibly see a non-smooth change in R translating into a discontinuity in growth.

seabbs:
Yes it is. Hmm.

R/extract.R
```diff
@@ -119,7 +119,7 @@ extract_parameter_samples <- function(stan_fit, data, reported_dates, reported_i
   out$growth_rate <- extract_parameter(
     "r",
     samples,
-    reported_dates
+    reported_dates[-1]
```
hsbadr, line 122: Instead of dropping the first time step, update `calculate_growth` to either get an approximate value from `Rt` or set it to `NA` .

inst/stan/estimate_infections.stan
```diff
@@ -118,21 +118,18 @@ model {
 generated quantities {
   int imputed_reports[ot_h];
   vector[estimate_r > 0 ? 0: ot_h] gen_R;
-  real r[ot_h];
+  real r[ot_h - 1];
```
hsbadr, line 121: This is related to approximating the growth rate for the 1st time step.

inst/stan/estimate_infections.stan
```diff
@@ -118,21 +118,18 @@ model {
...
-  }else{
+  if (estimate_r == 0){
     // sample generation time
     real gt_mean_sample = normal_rng(gt_mean_mean, gt_mean_sd);
     real gt_sd_sample = normal_rng(gt_sd_mean, gt_sd_sd);
     // calculate Rt using infections and generation time
     gen_R = calculate_Rt(infections, seeding_time, gt_mean_sample, gt_mean_sample,
                          max_gt, rt_half_window);
-    // estimate growth from calculated Rt
-    r = R_to_growth(gen_R, gt_mean_sample, gt_sd_sample);
```
hsbadr, line 134: Now that `calculate_growth` and `R_to_growth` would have the same return value and vector length, you might keep both with a switch, if there's a value in keeping the `Rt` growth rate approximation (e.g., for diagnostics, not necessarily as a user option).

inst/stan/functions/generated_quantities.stan
```diff
@@ -44,3 +44,11 @@ real[] R_to_growth(vector R, real gt_mean, real gt_sd) {
...
   return(r);
 }
+// Calculate growth rate
+real[] calculate_growth(vector infections, int seeding_time) {
+  int t = num_elements(infections);
+  int ot = t - seeding_time;
+  vector[t] log_inf = log(infections); 
+  vector[ot] growth = log_inf[(seeding_time + 1):t] - log_inf[seeding_time:(t - 1)];
+  return(to_array_1d(growth));
+}
```
hsbadr, line 54: Update, if needed, according to the above comments.

inst/stan/simulate_infections.stan
```diff
@@ -27,7 +27,7 @@ generated quantities {
   matrix[n, t] infections; //latent infections
   matrix[n, t - seeding_time] reports; // observed cases
   int imputed_reports[n, t - seeding_time];
-  real r[n, t - seeding_time];
+  real r[n, t - seeding_time - 1];
```
hsbadr, line 30: This is related to approximating the growth rate for the 1st time step.

inst/stan/simulate_infections.stan
```diff
@@ -48,6 +48,6 @@ generated quantities {
     }
    // simulate reported cases
    imputed_reports[i] = report_rng(to_vector(reports[i]), rep_phi[i], model_type);
-   r[i] = R_to_growth(to_vector(R[i]), gt_mean[i, 1], gt_sd[i, 1]);
+   r[i] = calculate_growth(to_vector(infections[i]), seeding_time + 1);
```
hsbadr, line 51: Could use a switch as discussed above.

tests/testthat/test-stan-generated_quantities.R
```diff
@@ -0,0 +1,16 @@
...
+
+
+test_that("calculate_growth works as expected", {
+  skip_on_cran()
+  expect_equal(calculate_growth(rep(1, 5), 1), rep(0, 4))
+  expect_equal(round(calculate_growth(1:5, 2), 2), c(0.41, 0.29, 0.22))
+  expect_equal(round(calculate_growth(exp(0.4*1:5), 2), 2), rep(0.4, 3))
+  expect_error(calculate_growth(1:5, 6))
+  expect_error(calculate_growth(1:5, 0))
+})
```
hsbadr, line 15: Could also add tests to compare the outputs from `calculate_growth` and `R_to_growth`, and to check if there's discontinuities in the forecasts (maybe by comparing the difference between the first forecase time step and the previous day within a reasonable threshold).

sbfnk:
Closing in favour of #307

PR #214, opened 2020-12-30
Standardises generation of infections

seabbs:
This PR standardises the generation of infections and the use of covariates (currently GP and BP only). Work in progress.

### DIFF
--- a/inst/stan/functions/covariates.stan
+++ b/inst/stan/functions/covariates.stan
@@ -1,15 +1,15 @@
-// update a vector of Rts
-vector update_Rt(vector input_R, real log_R, vector noise, int[] bps,
-                 real[] bp_effects, int stationary) {
+// update combined covariates
+vector update_covariate(real base_cov, vector noise, int[] bps,
+                        real[] bp_effects, int stationary,
+                        int t, int link) {
   // define control parameters
-  int t = num_elements(input_R);
   int bp_n = num_elements(bp_effects);
   int bp_c = 0;
   int gp_n = num_elements(noise);
   // define result vectors
   vector[t] bp = rep_vector(0, t);
   vector[t] gp = rep_vector(0, t);
-  vector[t] R;
+  vector[t] cov;
   // initialise breakpoints
   if (bp_n) {
     for (s in 1:t) {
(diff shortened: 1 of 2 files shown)


sbfnk:
Not having looked in detail can‘t this be simplified further by using the `calculate_Rt` function and separating out setting of the GP prior (whether on Rt or infections)?

sbfnk:
I.e., can't we write the model as in https://github.com/epiforecasts/EpiNow2/issues/211#issuecomment-752566612_ where R_t, I_t and r are all parameters (i.e., what's currently generate infections becomes `calculate_Rt`? The only difference between the models is then in the prior choice, e.g. R_t could be a GP prior or an uninformed prior (i.e., a backcalculation model in which case it will just be calculated from uncertain I_t as in the first equation in the comment referenced), and this could be done in separate functions. Or am I missing something?

If this can be done, it would have the nice effects of, e.g., being able to use breakpoints/covariates on Rt in a model that has a GP prior on I_t and uninformed prior on R_t.

sbfnk:
Looking again most of this is there already: all it would need is for all models to have R_t, r_t and I_t as parameters, and all models to use `generate_infections_with_infectiousness` (which I think should just be called `renewal_model`) whereas `generate_infections_with_infectiousness` would be re-named to reflect that all it does is set a prior on infections (which is all that determines infections if the priors on R_t and r_t are unconstrained - though in practice one might want to bound them an turn into uninformative proper priors even when not using a GP, e.g. uniform with some bound, that way their posteriors become proper).

Wouldn't change anything in practice but reduce code complexity and duplicate code (e.g. calculate_Rt vs generate_infections).

seabbs:
Hmm so in my mind it was a choice (i.e infections directly, infections via renewal equation, infections via growth. It is not totally clear to me that you could do all of them at once as the renewal equation is interactive - will stare at the code a bit.

sbfnk:
Hmmm, yeah so it might depend on whether Stan interprets an equals sign as a statement of equality or an assignment. If there's no way to have it interpret as equality than my suggestion won't work.

seabbs:
So my understanding of what it is doing is assignment - that could be incorrect however.

sbfnk:
Closing in favour of #307

PR #216, opened 2021-01-06
Update code linting and formatting

hsbadr:
This tries to address the linting warnings and test code formatting.


hsbadr:
@seabbs I've tested both `Lint` and `Format` workflows in my fork and they work as expected. The `Lint` workflow works fine here, but the `Format` workflow fails push the committed changes. This could be due to branch protection or GHA user permissions. Maybe, it'll work when you (or an admin) push to `master`.

seabbs (approved):
Thanks - looks good. I'll try a master commit.

PR #217, opened 2021-01-06
Update label in summary

sbfnk:
This is to match changes in #207 which are currently not feeding through to the website


sbfnk:
Beyond this change I'm a bit puzzled why so much on the web site (looking at UK) is labelled as Unsure where that now represents a relatively narrow probability range of 0.4-0.6. Has this not fed through somehow?

seabbs (approved):
Thanks. Having this manually defined in a bunch of places is really not ideal. 

I thought I had found the issue with the website last night and fixed it but clearly not. I'll try again with this update and then have another look if it doesn't resolve the issue.

PR #218, opened 2021-01-09
fix estimate of R <= 1

sbfnk:
I think this fixes the problem we've been having with the expected change, and also saves the probability of control in the summary for separate evaluation.


seabbs (approved):
Looks good to me.

PR #220, opened 2021-01-13
Support samples in simulate_infections

seabbs:
* Added supported to `simulate_infections` so that a `data.frame` of R samples can be passed in instead of a vector of R values.
* Added extraction of posterior samples to the summary method for `estimate_infections`.

This is being used in our German forecasting hub submission to convert an expert Rt forecast into a case forecast. It will also facilitate the overhaul and improvement of `forecast_infections` (which can now be recoded to forecast Rt and then call `simulate_infections`).


seabbs (commented):
🕺

PR #224, opened 2021-01-18
GHA: Cleanup workflow

hsbadr:
This line isn't needed anymore and should be removed. It causes silent `Error: there is no package called ‘curl’`.


seabbs (approved):
:+1:

PR #225, opened 2021-01-24
Update README.Rmd

pitmonticone:
Update URL for 

Gostic KM, McGough L, Baskerville EB, Abbott S, Joshi K, Tedijanto C, et al. (2020) Practical considerations for measuring the effective reproductive number, Rt. PLoS Comput Biol 16(12): e1008409. https://doi.org/10.1371/journal.pcbi.1008409


seabbs:
Thanks! Good spot.

PR #227, opened 2021-02-06
Don’t round summaries by default

sbfnk:
I think this is what’s needed to enable us to generate figures on the web site from markdown


seabbs:
:+1:

PR #230, opened 2021-02-12
Update README.md

pitmonticone:
Update with 

Gostic KM, McGough L, Baskerville EB, Abbott S, Joshi K, Tedijanto C, et al. (2020) Practical considerations for measuring the effective reproductive number, Rt. PLoS Comput Biol 16(12): e1008409. https://doi.org/10.1371/journal.pcbi.1008409

as in README.Rmd .


seabbs:
Thanks for this!

Issue #238, opened 2021-03-04
unexpected data.table environment side effects


pearsonca:
In some functions, EpiNow2 uses `setDTthreads(...)` to manage the interaction of `data.table` parallelism with other libraries approaches to parallelism. However, that invocation appears to have a global side effect in any given session, rather than a locally limited one.

This is unexpected (and undesirable) behaviour, since a script might use e.g. `estimate_infections` then later expect to use `data.table` parallelism when digesting those outcomes or simulating based on them etc.

PR #245, opened 2021-04-14
Reset data.table thread side effect resolution

medewitt:
The purpose of this pull request is to address #238 where the behaviour of data.table in the global environment is changed by some of the core EpiNow2 functions.  This pull request adds code that retrieves the number of threads assigned by data.table, sets it them to 1, then returns the initial number of threads on exit. 

No new package dependencies.


seabbs (changes requested):
This is a really nice change and definitely what should have been in from the first. As this code is repeated a few times and is now a few lines could you wrap this into a internal function? It would be great if you could add a line to the news describing this change and giving yourself credit! (apologies this took a while to get back to)

medewitt:
@seabbs sounds good. Initially, I was concerned about scoping of environments, but I think I have solution with a `set_dt_single_thread` function:

``` r
library(data.table)
setDTthreads(2)
set_dt_single_thread <- function() {
    a <- list2env(x = list(dt_previous_threads = getDTthreads()))
    assign(deparse(substitute(x)), a, env = parent.frame())
    setDTthreads(1)
    do.call("on.exit", 
                    list(quote(setDTthreads(x$dt_previous_threads))), 
                    envir = parent.frame()) 
    }

test_function <- function(){
    print(getDTthreads())
    set_dt_single_thread()
    print(getDTthreads())
    test <- data.table(n = 100, y = rnorm(100))
}
test_function()
#> [1] 2
#> [1] 1
getDTthreads()
#> [1] 2
```

<sup>Created on 2021-05-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>
```

**Comment by seabbs** [comment]
Ah I hadn't really considered the complexities here but that is a pretty clever solution (so nice in fact I wonder if they might be interested a version of it in `data.table` as it seems like a pretty common thing package devs using `data.table` might want to do.

**Comment by medewitt** [comment]
Thanks! And that's an interesting idea. I wonder, too, if `setDTthreads` could have an environment argument that defaults to the global, but could be set to the local environment for exactly the purposes this function supports.  As you say, the threading on data.table can get sticky when you don't anticipate it....

**Review by medewitt — COMMENTED** [review]
Consolidated extra lines of code to a single new utility function called `set_dt_single_thread `

Also looks like the checks failed installing gcc when running the test-coverage step (unrelated to modification). 

```
==> Installing gdal dependency: gcc
==> Pouring gcc--11.1.0.catalina.bottle.tar.gz
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink bin/gfortran
Target /usr/local/bin/gfortran
already exists. You may want to remove it:
  rm '/usr/local/bin/gfortran'
```

**Review by seabbs — CHANGES_REQUESTED** [review]
Looks great. Can you add a test for this allow the lines of your example and then good to merge. Thanks for flagging to CI issue will investigate.

**Comment by medewitt** [comment]
> Looks great. Can you add a test for this allow the lines of your example and then good to merge. Thanks for flagging to CI issue will investigate and thanks again for the nice contribution.

Ok, this should be good to go. I added the tests for functionality and fixed the CI issue with a Fortran symlink conflict. Plus I learned that data.table multithreading is not immediately enabled in CI for Macs. Just upgraded to R 4.0.5 on my mac and had to change compiler flags to get multithreading to work for data.table, so makes sense why the mac VM isn't configured to do so.

**Review by seabbs — APPROVED** [review]
Looks great! Ah interesting that makes sense. Nice fix for the CI.

**Comment by pearsonca** [comment]
very lazily going through my github notifications.

neat approach - conceptually like [RAII / SBRM](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) pattern.

one note: does it make sense to `@export` this? It's not meant to be used outside a function. As suggested earlier in this discussion, I could see this as a function exported by `data.table`, but it's a distraction in the "vocabulary" of this package.

Also: I went to look for whether `data.table` exports something like this and found:

```
data.table automatically switches to single threaded mode upon fork (the mechanism used by parallel::mclapply and the foreach package). Otherwise, nested parallelism would very likely overload your CPUs and result in much slower execution. As data.table becomes more parallel internally, we expect explicit user parallelism to be needed less often. The restore_after_fork option controls what happens after the explicit fork parallelism completes. It needs to be at C level so it is not a regular R option using options(). By default data.table will be multi-threaded again; restoring the prior setting of getDTthreads(). But problems have been reported in the past on Mac with Intel OpenMP libraries whereas success has been reported on Linux. If you experience problems after fork, start a new R session and change the default behaviour by calling setDTthreads(restore_after_fork=FALSE) before retrying. Please raise issues on the data.table GitHub issues page.
```

That suggests that they may have developed a more automatic way out of what EpiNow2 was trying to work around. Worth returning to investigate whether `data.table` latest now addresses the problem generally trying to be solved here.

PR #253, opened 2021-06-03
Minor stan optims

seabbs:
This PR adds some small optimisations to the stan code. These include: 

- Making use of `fmax` and `fmin` vs using if checks.
- Only using the `target` framework when strictly needed (this means a constant does not need to be calculated and so should reduce run times). 

This PR also adds a two additional observation model arguments: `likelihood` and `return_likelihood`. These allow the model to return information from only the prior and to default to not calculating the overall run time. 

I also explored removing the weighting used on the delays and generation time. Without the weights the posteriors differed from the input priors (but did not go to zero). Under the assumption the model really contains no info (or very little) on which to base these parameters I added the weights back in - this is not a programming choice though so can be thought about later. 

I used the examples built in to `esimate_infections` to compare performance against the CRAN version of the package. All results are comparable and the first example runs for me in around 80% of the time.


hsbadr:
@seabbs Is there any reason to stick to the current `lognormal` prior on the GP length scale? I get better stability with inverse Gamma (check [this](https://betanalpha.github.io/assets/case_studies/gp_part3/part3.html)), especially with long simulations, and discarding the `ls_mean` & `ls_sd` in `gp_opts()`.
```bash
# Use inverse-gamma prior on Gaussian process (GP) length-scale parameter
sed -i 's|rho ~ lognormal(ls_meanlog, ls_sdlog)|rho ~ inv_gamma(1.499007, 0.057277 * ls_max)|g' inst/stan/functions/gaussian_process.stan

File: R/estimate_infections.R [hunk 1]
```diff
@@ -52,7 +52,7 @@
 #'
 #' # default setting
 #' # here we assume that the observed data is truncated by the same delay as
-#' def <- estimate_infections(reported_cases,
+# ' def <- estimate_infections(reported_cases,
 #'   generation_time = generation_time,
```
hsbadr, line 56: This line is belongs to `estimate_infections()` in the previous line:
```r
Warning message:
When processing R/estimate_infections.R: <text>:18:36: unexpected ','
17: # ' def <- estimate_infections(reported_cases,
18:   generation_time = generation_time,
```

R/opts.R
```diff
@@ -248,6 +248,10 @@ gp_opts <- function(basis_prop = 0.2,
 #'  to map latent infections (convolved to date of report). If none empty a mean
 #'  (`mean`) and standard deviation (`sd`) needs to be supplied defining the normally
 #'  distributed scaling factor.
+#' @param likelihood Logical, defaults to `TRUE`. Should the likelihood be
+#'  included in the model
+#' @param return_likelihood Logical, defaults to `FALSE`. Should the likelihood
+#'  be returned by the model.
```
hsbadr, line 254: Couldn't we remove `return_likelihood` and add `"likelihood"` to the list of supported outputs?
By using the `output` argument like:
```r
    output = c(
      "fit",
      "latest",
      "likelihood",
      "plots",
      "regions",
      "samples",
      "summary",
      "timing"
    )
```

hsbadr: Also, the `create_obs_model` test fails due to the changed number of arguments:
https://github.com/epiforecasts/EpiNow2/blob/a582f210bb8a4b487aff1722f99777119fa83c3b/tests/testthat/test-create_obs_model.R#L6

seabbs: Thanks @hsbadr 

Can't add to output as that is only supported in `epinow` and not in the internal functions. That being said I am kicking off a round of changes with the first focus being some work on the stan model. Once that is done going to look at improving the interface where this might come in (and probably some fairly big other changes) - just kicking the tires with a couple of PRs to get back up to speed on the code base.

R/estimate_infections.R
seabbs: :+1: thanks good spot

R/opts.R
hsbadr: Sounds great. Thanks for heads up!

seabbs:
thanks for the mini-review @hsbadr (opened an issue to discuss the log normal prior).

seabbs:
PR also pulls in some documentation updates.

inst/stan/functions/observation_model.stan
```diff
@@ -68,64 +68,67 @@ void truncation_lp(real[] truncation_mean, real[] truncation_sd,
...
       target += poisson_lpmf(cases | reports) * weight;
-    } else {
-      target += neg_binomial_2_lpmf(cases | reports, sqrt_phi) * weight;
     }
   } else {
-    target += poisson_lpmf(cases | reports) * weight;
+    if (weight == 1) {
+      cases ~ neg_binomial_2(reports, sqrt_phi);
+    }else{
+      target += poisson_lpmf(cases | reports) * weight;
```
seabbs, line 89: should be negbin

seabbs (commented):
Bug in likelihood when weights are non-zero.

inst/stan/functions/observation_model.stan
seabbs: done

seabbs:
@sbfnk merging but keen to make changes if there issues.

Issue #255, opened 2021-06-03
Is there any reason to stick to the current `lognormal` prior on the GP length scale? 


seabbs:
@seabbs Is there any reason to stick to the current `lognormal` prior on the GP length scale? I get better stability with inverse Gamma (check [this](https://betanalpha.github.io/assets/case_studies/gp_part3/part3.html)), especially with long simulations, and discarding the `ls_mean` & `ls_sd` in `gp_opts()`.
```bash
# Use inverse-gamma prior on Gaussian process (GP) length-scale parameter
sed -i 's|rho ~ lognormal(ls_meanlog, ls_sdlog)|rho ~ inv_gamma(1.499007, 0.057277 * ls_max)|g' inst/stan/functions/gaussian_process.stan

_Originally posted by @hsbadr in https://github.com/epiforecasts/EpiNow2/issues/253#issuecomment-853911207_

seabbs:
thanks for the mini-review @hsbadr 

On your point about the length scale prior. No there is no need to stick to log normal. We went with it at the time as on simulations it worked well and it was easier to parameterise for non-expert users. Having said that since then I have only been using inverse gamma in other projects (i.e here: https://github.com/seabbs/inc2prev/blob/c293db869401f0aedc4d35a5e8a195dda54196ab/stan/model.stan#L65) So very open to moving back to inverse gamma. The key is that it is easy for users to spec.

hsbadr:
@seabbs Sounds good. I'll submit a PR for this.

PR #257, opened 2021-06-04
GP: lengthscale: Use inverse-gamma prior if `ls_sd = 0`

hsbadr:
This is a proposal to address #255. It allows using an inverse-gamma prior on Gaussian process (GP) length-scale parameter while keeping the user interface and the default log-normal prior unchanged. The parameters of the inverse-gamma prior are a function of `ls_max` and have been tested for both short and long simulations where the default prior makes the model unstable. The new prior is **more stable for long simulations** and adaptively change the distribution based on the simulation length (total number of days) without relying on the user inputs or the fixed defaults. **It can be tested by setting `ls_sd = 0` in `gp_opts()`.**

I haven't updated the docs (`roxygen2::roxygenize()`) or reconfigured because I use the development version on Stan.

Here's the summary plot from the example with `gp = gp_opts(ls_sd = 0)`:

image

@seabbs Feel free to make any changes to this proposal.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -485,7 +485,7 @@ create_initial_conditions <- function(data) {
       out$eta <- array(rnorm(data$M, mean = 0, sd = 0.1))
       out$rho <- array(rlnorm(1,
         meanlog = data$ls_meanlog,
-        sdlog = data$ls_sdlog * 0.1
+        sdlog = ifelse(data$ls_sdlog > 0, data$ls_sdlog * 0.1, 0.01)
       ))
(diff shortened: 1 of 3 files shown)


hsbadr:
@seabbs In the manual, `ls_min` defaults to 7 while it's actually `ls_min = 3`. I'd like to change the default to `ls_min = 0` and update the docs. What do you think?

hsbadr:
@seabbs I've tested this. It's ready to go.

mommamgaypay (approved):
#260

PR #258, opened 2021-06-08
Update README.md

seabbs:
Switches to use the universe from @Bisaloo - keeps github install instructions for the real keen beans but adds a warning .


seabbs (commented):
LGTM (self review)

PR #260, opened 2021-06-14
Add overdispersion as user facing parameter

seabbs:
This PR adds the overdispersion prior (which is scaled as 1 over the square root) as a user facing prior with no change to the package defaults. This allows users to update this based on prior knowledge etc. The aim of this additon is to be observation model agnostic so that future updates can use this parameter to set priors for other observation family parameters.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,7 +13,7 @@ This release is under development and the features outlined below may change bef
 
 * Added support for varying the length of the day of the week effect (see `obs_opts()`). This allows, for example, fitting to data with cases only reported every 3 days. 
 * Minor optimisations in the observation model by only using the `target` likelihood definition approach when required and in the use of `fmax` and `fmin` over using if statements.
-* Added support for users setting the overdispersion (parameterised as one over the square root of phi) of the reporting process. This is accessible via the `phi` argument of `obs_opts` wit the default of a normal distribution with mean 0 and standard deviation of 1 truncated at 0 remaining unchanged. 
+* Added support for users setting the overdispersion (parameterised as one over the square root of phi) of the reporting process. This is accessible via the `phi` argument of `obs_opts` with the default of a normal distribution with mean 0 and standard deviation of 1 truncated at 0 remaining unchanged. 
 
 ## Documentation
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,6 @@
 
 This release is under development and the features outlined below may change before release.
 
-
 ## New features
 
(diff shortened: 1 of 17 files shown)


NEWS.md
```diff
@@ -14,6 +13,7 @@ This release is under development and the features outlined below may change bef
 
 * Added support for varying the length of the day of the week effect (see `obs_opts()`). This allows, for example, fitting to data with cases only reported every 3 days. 
 * Minor optimisations in the observation model by only using the `target` likelihood definition approach when required and in the use of `fmax` and `fmin` over using if statements.
+* Added support for users setting the overdispersion (parameterised as one over the square root of phi) of the reporting process. This is accessible via the `phi` argument of `obs_opts` wit the default of a normal distribution with mean 0 and standard deviation of 1 truncated at 0 remaining unchanged.
```
hsbadr, line 16:
```suggestion
* Added support for users setting the overdispersion (parameterised as one over the square root of phi) of the reporting process. This is accessible via the `phi` argument of `obs_opts` with the default of a normal distribution with mean 0 and standard deviation of 1 truncated at 0 remaining unchanged. 
```

inst/stan/data/observation_model.stan
```diff
@@ -1,5 +1,7 @@
   int day_of_week[t - seeding_time]; // day of the week indicator (1 - 7)
   int model_type;                    // type of model: 0 = poisson otherwise negative binomial
+  int phi_mean;                      // Mean and sd of the normal prior for the
+  int phi_sd;                        // reporting process
```
hsbadr, line 4: Shouldn't `phi_mean` and `phi_sd` be `real`, not `int`?

inst/stan/functions/observation_model.stan
```diff
@@ -66,12 +66,12 @@ void truncation_lp(real[] truncation_mean, real[] truncation_sd,
 }
 // update log density for reported cases
 void report_lp(int[] cases, vector reports,
-               real[] rep_phi, int phi_prior,
+               real[] rep_phi, int phi_mean, int phi_sd,
```
hsbadr, line 69: Shouldn't `phi_mean` and `phi_sd` be `real`?

inst/stan/data/observation_model.stan
seabbs: Yes - haha this just flummoxed me a bit in testing

inst/stan/functions/observation_model.stan
seabbs: Yes good spot (and thanks for the review)

seabbs (commented):
LGTM (self-review)

PR #262, opened 2021-06-28
Add optional plot of forecast horizons. 

seabbs:
Adds optional plotting of model estimates by type (i.e forecast etc) and cascades this argument through higher level functions. Adds some basic unit tests for plotting (not failing vs actual ouput tests).


R/plot.R
```diff
@@ -166,7 +186,7 @@ plot_estimates <- function(estimate, reported, ylab = "Cases", hline,
   # plot estimates
   plot <- plot +
     ggplot2::geom_vline(
-      xintercept = estimate[type == "Estimate based on partial data"][date == max(date)]$date,
+      xintercept = orig_estimate[type == "Estimate based on partial data"][date == max(date)]$date,
```
sbfnk, line 189: Do we definitely want this line for all values of `estimate_type`? I'm thinking e.g. perhaps not if `estimate_type == "Estimate"`?

sbfnk (approved):
Super!

R/plot.R
seabbs: 🤷🏻  

The benefit to having it is you know when the model thinks it was run which is always useful but yes downside its not really anchored to anything if you limit what you are plotting.

sbfnk: Yeah makes sense, let's keep it in.

PR #272, opened 2021-09-24
Add flag for stanc3 compatibility

andrjohns:
Hi,

This PR adds a compiler flag to your `Makevars` files which is needed for compatibility with the upcoming version of `rstan`:  `-DUSE_STANC3`

The Stan-to-c++ transpiler (`stanc`) was refactored after 2.21, and so different c++ is used in the new `rstan` package (>= 2.26). Because of this, the flag `-DUSE_STANC3` needs to be added when a stan model is being run using the new rstan package (and the new stanc implementation).

This PR adds the compiler flag, which is conditionally set based on the installed `rstan` version.

Feel free to let me know if you need any more info.

Thanks!
Andrew


andrjohns:
Hi,

I'm closing this PR as we're implementing the fix the rstantools, so no changes to your package will be necessary

Issue #284, opened 2021-11-16
Adding an option for not filtering zeros at start of reported case time series


LloydChapman:
For my particular use case, it would be helpful to have an option not to filter zeros in the reported case time series before the first reported case as is currently done in `create_clean_reported_cases`, i.e. to add an option to `epinow` and `estimate_infections` to switch the filtering on and off.
I'll create a pull request with this feature.

Issue #283, opened 2021-11-16
Mistake in implementation of zero_threshold in create_clean_reported_cases()


LloydChapman:
I think there are a couple of mistakes in the implementation of the `zero_threshold` for smoothing zeros in the reported case time series when the average of the preceding week's cases are above a certain threshold. In line 37 of `create.R`, I think it should be `!is.infinite(zero_threshold)` rather than `is.infinite(zero_threshold)`, and in line 41, I think the current day's cases should be removed from the sum of the last 8 days cases before dividing by 7 to get the average cases over the last week, i.e.: 
```
  if (!is.infinite(zero_threshold)) {
    reported_cases <-
      reported_cases[
        ,
        `:=`(average_7 = (data.table::frollsum(confirm, n = 8) - confirm) / 7)
      ]
    reported_cases <- reported_cases[
      confirm == 0 & average_7 > zero_threshold,
      confirm := as.integer(average_7)
    ][
      ,
      c("average_7") := NULL
    ]
  }
```
I'll open a pull request with these edits.

PR #285, opened 2021-11-16
Add option for filtering leading zeros and correct zero_threshold

LloydChapman:
Suggested changes to address issues:
#283 - correcting implementation of zero_threshold 
#284 - adding option for filtering leading zeros
I'm not sure how the error handling in EpiNow2 works, so haven't added a warning about early Rt estimates being unreliable, but I think that would be good to add.


DESCRIPTION
```diff
@@ -63,6 +63,10 @@ Authors@R:
              family = "Monticone",
              role = "ctb",
              email = "pietro.monticone@edu.unito.it"),
+      person(given = "Lloyd",
```
seabbs, line 66: I've added you as a contributor will revert if not okay.

R/create.R
```diff
@@ -3,15 +3,20 @@
 #' Cleans a data frame of reported cases by replacing missing dates with 0 cases and applies an optional
 #' threshold at which point 0 cases are replaced with a moving average of observed cases. See `zero_threshold`
 #' for details.
-#' @param zero_threshold `r lifecycle::badge("experimental")` Numeric defaults to 50. Indicates if detected zero
-#' cases are meaningful by using a threshold of 50 cases on average over the last 7 days. If the average is
-#' above this threshold then the zero is replaced with the backwards looking rolling average. If set to infinity
-#' then no changes are made.
+#' @param filter_leading_zeros Logical, defaults to TRUE. Should zeros at the
```
seabbs, line 6: I've changed the name so its clear what it is for in upstream functions.

R/create.R
```diff
@@ -27,18 +32,20 @@ create_clean_reported_cases <- function(reported_cases, horizon, zero_threshold
...
-  # is greater than a threshold
-  if (is.infinite(zero_threshold)) {
+  # Check case counts preceding zero case counts and set to 7 day average if
+  # average over last 7 days is greater than a threshold
+  if (!is.infinite(zero_threshold)) {
     reported_cases <-
       reported_cases[
         ,
-        `:=`(average_7 = data.table::frollsum(confirm, n = 8) / 7)
+        `:=`(average_7 = (data.table::frollsum(confirm, n = 8)) / 7)
```
seabbs, line 48: I've reverted this as actually not a bug. This is because it is only used when `confirm = 0` and so both definitions are the same. I appreciate this is a little opaque but keeping for the the very slight efficiency gain.

seabbs (approved):
This LGTM - sorry for the very very long delay and thanks for the contribution. 
Good spot on the `is_infinite` usage.

PR #288, opened 2022-02-01
option for secondary baseline

sbfnk:
Adds a baseline level to the secondary model, i.e. a constant component that is not affected by the primary data stream (somewhat akin to the endemic component in `hhh4`. This was motivated by observations that the model currently seems quite a long way off the data in some countries (though not all) in the European Forecast Hub submissions. When inspecting some of the raw data on cases and admissions, the relationship seemed reasonably close to linear but with an intercept at >0. 

Visually, the updated model connects somewhat better to recent data (if not impressively so, possibly because still fitting to a mixture of Delta and Omicron), but it comes at the expense of less straightforward epidemiological interpretation of the estimated delay and scaling - happy to hear of other suggestions that could improve things.

From https://github.com/sophiemeakin/ecdc-covid19-admissions/blob/main/main.R

Without baseline:
image

With baseline:
image

### DIFF
--- a/R/estimate_secondary.R
+++ b/R/estimate_secondary.R
@@ -238,7 +238,8 @@ secondary_opts <- function(type = "incidence", ...) {
       historic = 1,
       primary_hist_additive = 1,
       current = 0,
-      primary_current_additive = 0
+      primary_current_additive = 0,
+      estimate_baseline = 0
     )
   } else if (type %in% "prevalence") {
     data <- list(
@@ -246,7 +247,8 @@ secondary_opts <- function(type = "incidence", ...) {
       historic = 1,
       primary_hist_additive = 0,
       current = 1,
-      primary_current_additive = 1
+      primary_current_additive = 1,
+      estimate_baseline = 0
     )
   }
   data <- update_list(data, list(...))
@@ -427,7 +429,8 @@ forecast_secondary <- function(estimate,
 
   # allocate empty parameters
   data <- allocate_empty(
-    data, c("frac_obs", "delay_mean", "delay_sd", "rep_phi"),
+    data, c("frac_obs", "delay_mean", "delay_sd", "rep_phi",
+            "secondary_baseline"),
     n = data$n
   )
   data$all_dates <- as.integer(all_dates)
--- a/inst/stan/estimate_secondary.stan
+++ b/inst/stan/estimate_secondary.stan
@@ -24,14 +24,15 @@ parameters{
   real truncation_mean[truncation];      // mean of truncation
   real truncation_sd[truncation];        // sd of truncation
   real<lower = 0> rep_phi[model_type];   // overdispersion of the reporting process
+  real<lower = 0> secondary_baseline[estimate_baseline]; // baseline for secondary component
 }
 
 transformed parameters {
   vector<lower=0>[t] secondary;
   // calculate secondary reports from primary
-  secondary = calculate_secondary(primary, obs, frac_obs, delay_mean,
-                                  delay_sd, max_delay, cumulative,
-                                  historic, primary_hist_additive,
+  secondary = calculate_secondary(primary, obs, frac_obs, secondary_baseline,
+                                  delay_mean, delay_sd, max_delay,
+                                  cumulative, historic, primary_hist_additive,
                                   current, primary_current_additive, t);
  // weekly reporting effect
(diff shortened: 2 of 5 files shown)


sbfnk:
Tagging @sophiemeakin

sophiemeakin:
This seems like a good temporary fix and the code changes seem fine - but I wonder why this is happening? Some thoughts:
My first q would be is the misalignment between observed and forecast admissions due to the fit of the secondary model, or the case forecasts? 
If the former - do we have an idea why is this happening, and so whether it's a temporary or permanent issue? If it's only a temporary change (e.g. change in variant distributions, reporting etc. etc.), then we should look at how adding the baseline will affect other locations where this temporary change isn't present (e.g. the UK forecast looks better without the baseline than with it).
If the latter (although I suspect it isn't as we are using the hub-ensemble case forecast), then I personally wouldn't include the baseline in the model - the issue is with the case forecasts and we should think about how we can correct those instead.

sbfnk:
> This seems like a good temporary fix and the code changes seem fine - but I wonder why this is happening? Some thoughts: My first q would be is the misalignment between observed and forecast admissions due to the fit of the secondary model, or the case forecasts? If the former - do we have an idea why is this happening, and so whether it's a temporary or permanent issue?

The fits of the secondary model to past data look poor so I think it's a combination of data and variants that break the underlying assumption of linear scaling over a 12-week period, even before case forecasts come in to complicate things further.

As an illustration, here's cases and 1-week shifted admissions, and a linear fit with setting intercept at 0 vs. leaving it free.

image

inst/stan/estimate_secondary.stan
```diff
@@ -42,6 +43,9 @@ transformed parameters {
 }
 
 model {
+  if (estimate_baseline) {
+    secondary_baseline[1] ~ normal(0, min(obs)) T[0, ];
```
seabbs, line 47: better to set min(obs) in transformed data.

seabbs: or better yet in R

R/estimate_secondary.R
```diff
@@ -238,15 +238,17 @@ secondary_opts <- function(type = "incidence", ...) {
       historic = 1,
       primary_hist_additive = 1,
       current = 0,
-      primary_current_additive = 0
+      primary_current_additive = 0,
+      estimate_baseline = 0
```
seabbs, line 242: user control to turn this on or off independently? Could think about doing so by passing the mean intercept and having 0 be the setting that turns it off (so endemic = c(100, 10) is a prior of N(100, 10), and endemic = 0 turns it off)

R/estimate_secondary.R
```diff
@@ -238,15 +238,17 @@ secondary_opts <- function(type = "incidence", ...) {
...
     )
   } else if (type %in% "prevalence") {
     data <- list(
       cumulative = 1,
       historic = 1,
       primary_hist_additive = 0,
       current = 1,
-      primary_current_additive = 1
+      primary_current_additive = 1,
+      estimate_baseline = 0
```
seabbs, line 251: better to add this to the list outside of the if-else step to avoid duplication.

seabbs (changes requested):
Interesting stuff.

 I think having this as a component of the model makes sense (with thinking about moving towards a renewal + convolution of upstream + endemic being a place to target). Not sure it is exactly the resolution to the current problem which I thought was more driven by rapid changes in the CFR, CHR etc due to Omicron replacing Delta and large scale boosting (we saw this early last year when vaccination was rolling out but it was less obvious as most places were on the downslope). Having it be turned of by default seems like a good call but I think exposing and documenting to the user would be a good idea. 

We looked at having that CHR being time-varying and it became quite involved as a project (#261). A simplified solution using a random walk on the scaling parameter might help quite a bit? 

In terms of this PR. I think probably need the following: 

- [ ] Expose setting to the user
- [ ] Document
- [ ] Expose prior setting to the user (optional)
- [ ] Add news update 
- [ ] Add version bump
- [ ] Test addition (optional).

seabbs:
Are you still interested in having this merged?

sbfnk:
Possibly but it needs to be revisited and your bullet points addressed. I've converted it to a draft for now.

sbfnk:
I'm not convinced this is still needed. Closing for now.

PR #303, opened 2022-07-21
don't pass R to update_Rt

sbfnk:
This slightly simplifies the call to `update_R` (also reducing the amount of memory copying required) - the function never uses the contents of `R`, just its length, which is stored in a separate variable.


seabbs (approved):
LGTM

PR #304, opened 2022-07-21
only call truncate if truncation is modelled

sbfnk:
This simplifies the call to `truncate` in `estimate_infections`, only calling it when truncation is actually required. This seems more intuitive as the call to `truncate` was previously required even if no truncation was applied, but only to create an `obs_report` variable not containing the forecasts. This now follows the same logic as e.g. only applying scaling if `obs_scale` is set to 1, such that the decision of whether truncation is applied is now happening outside the `truncate` function, which now always truncates.

As a (positive, in my opinion) side effect this does away with having to pass vectors to the `truncate` function requiring declarations of vectors of size 1 in `estimate_truncation`.


seabbs (approved):
Looks good to me. The downside is more if/else shenanigans in the main model code potentially making it harder to track what is happening.

PR #307, opened 2022-07-27
switch to directly calculated growth rate and extend model options

sbfnk:
This PR is meant to supersede #213 and #214. It 
- finalises the work commenced in #214 to support [three different models](https://github.com/epiforecasts/EpiNow2/pull/214#issuecomment-753024893)
- merges #214 into #213 and implements most of the suggestions made there by @hsbadr 
- rebases #213 onto current `master`
- works with default options and a simple example (the README)

Things left to do:
- [ ] expose and document the model options and ensure the stan model options are set consistently
- [ ] more extensive testing
- [ ] update NEWS


sbfnk:
Any reason this work in progress shouldn't continue?

PR #322, opened 2022-10-16
Final changes for 1.3.3

seabbs:
*WIP - tracking down a 10 times performance reversion.*

This PR adds final changes for 1.3.3. In particular, it includes some extensions to documentation, improved test coverage, and performance tweaks.

Currently, I am tracking down a significant reversion in performance meaning this is not ready for merging. I assume this was a result of the optimisations to the discrete distributions but I had very poor performance both before and after meaning it may be something else.


seabbs:
I am seeing these performance issues in Docker but with a local install this PR runs faster than main. Investigating now if its an issue with my development environment vs a package issue.

seabbs:
I've tested this on Mac and Linux and seeing comparable results for all models compared to `main` (with this PR being 10% or so faster). On the strength of that, I think we should merge, iron out the last few changes etc as they crop up, and push to CRAN today?

PR #333, opened 2022-11-02
bug fix: return_likelihood check

sbfnk:
`return_likelihood` is never greater than 1 - so at the moment setting `return_likelihood = TRUE` results in failure because `log_lik` has size 0.


seabbs (approved):
LGTM

PR #343, opened 2023-01-18
Develop

seabbs:
Merges in development changes


sbfnk:
This isn't ready for merging, there is at least one bug in `develop`.

seabbs:
Do you know what it is?

sbfnk:
It's [this](https://github.com/epiforecasts/EpiNow2/issues/344) (posted with what I was planning to do to solve it)

inst/stan/functions/convolve.stan
```diff
@@ -1,48 +1,70 @@
...
-vector convolve(vector cases, vector rev_pmf) {
-    int t = num_elements(cases);
-    int max_pmf = num_elements(rev_pmf);
-    vector[t] conv_cases = rep_vector(1e-5, t);
-    for (s in 1:t) {
-        conv_cases[s] += dot_product(cases[max(1, (s - max_pmf + 1)):s],
-                                     tail(rev_pmf, min(max_pmf, s)));
+// convolve two vectors
+// length of x
+vector convolve(vector x, vector y, int len) {
```
seabbs, line 3: I think we have introduced a bug here as this convolves/adds two probability mass functions vs a backward rolling convolution of a vector and an associated PMF (as in the previous version). I think we likely need both. 

Trying to recreate synthetic data confirms there is likely a problem. 

gp_nuts

I am just compiling main to check to see if this issue was actually introduced elsewhere (though due to the lack of commits touching the stan code I hope this isn't likely).

seabbs: i.e we have this: https://package.epinowcast.org/reference/add_pmfs.html (setting up some tests to check)

sbfnk: Ah yes this looks like a bug and agree it's unlikely to be in `main`. Might be good to think about whether there's any test we can add (e.g. based on your example above) to catch instances where the model is being completely broken as seems to be the case here.

seabbs: yes thinking about it now. Just reading through model changes on `main` as well to check nothing small has been missed.

seabbs: So I looked at this on `main` and found that this behaviour was actually being driven by a bug in `simulate_infections()` which I have now corrected. Updating `develop` can reproduce synthetic data though it isn't entirely clear to me that it should be able to given the convolution implementation (perhaps it is equivalent and I am just not quite seeing it?). 

Regardless, I see that the synthetic validation takes around 2 hours on `develop` (https://github.com/epiforecasts/EpiNow2/actions/runs/3971839196) vs 35 minutes for `main` (https://github.com/epiforecasts/EpiNow2/actions/runs/3971832809) so it seems likely there is at the very least still a performance regression. 

gp_nuts

seabbs: So looking at the current vs old versions of `convolve` I think they are the same (just testing this explicitly now) and its just a matter of efficiency.

sbfnk: Interesting - would you expect `dot_product` to be more efficient than the double loop?

seabbs: yes, I would with the sparse matrix version being the fastest based on experience with `epinowcast`/early versions of this code.

seabbs: I've added some tests to confirm things are working as expected. CI should show any performance difference.

inst/stan/estimate_infections.stan
```diff
@@ -29,6 +29,34 @@ transformed data{
...
+    sum(delay_max) - num_elements(delay_max) + 1);
+  vector[gt_fixed[1] ? gt_max[1] : 0] gt_fixed_pmf;
+  vector[truncation && trunc_fixed[1] ? trunc_max[1] : 0] trunc_fixed_pmf;
+  vector[delay_max_fixed] delays_fixed_pmf;
+
+  if (gt_fixed[1]) {
+    gt_fixed_pmf = discretised_pmf(gt_mean_mean[1], gt_sd_mean[1], gt_max[1], gt_dist[1], 1);
+  }
+  if (truncation && trunc_fixed[1]) {
+    trunc_fixed_pmf = discretised_pmf(
```
seabbs, line 45: why is no truncation being used here?

inst/stan/estimate_infections.stan
```diff
@@ -29,6 +29,34 @@ transformed data{
...
+      trunc_mean_mean[1], trunc_sd_mean[1], trunc_max[1], trunc_dist[1], 0
+    );
+  }
+  if (n_fixed_delays) {
+    delays_fixed_pmf = combine_pmfs(
```
seabbs, line 50: Similarly here why is not truncation being used?

inst/stan/estimate_infections.stan
sbfnk: to allow zero delays (which we can support unlike 0 generation times)

seabbs: which are happening in the indexing already? This is to truncate the mean vs right shifting?

seabbs: I've added a news note indicating this is a breaking change and updated the arg to have a clearer name (`left_truncate`).

inst/stan/functions/pmfs.stan
```diff
@@ -1,52 +1,36 @@
...
+    } else {
+      reject("Unknown distribution function provided.");
     }
-  }else{
-    pmf[1] = upper_cdf[1];
-    pmf[2:n] = upper_cdf[2:n] - upper_cdf[1:(n-1)];
+    // discretise
+    pmf = upper_cdf[2:(n + 1)] - upper_cdf[1:n];
+    // normalize
+    pmf = pmf / (upper_cdf[n + 1] - upper_cdf[1]);
```
seabbs, line 28: This normalisation is accounting for left and right truncation. In the case of the left truncation, it includes probability mass on day 1 as well as day zero so it is not clear to me that it is correct?

inst/stan/functions/convolve.stan
```diff
@@ -1,48 +1,88 @@
...
+  vector[len] pmf = rep_vector(0, len);
+  if (n_fixed > 0) {
+    pmf[1:n_fixed] = fixed_pmf;
+  } else if (n_variable > 0) {
+    pmf[1] = 1;
+  }
+  for (s in 1:n_variable) {
+    vector[pmf_n[s]] variable_pmf;
+    variable_pmf = discretised_pmf(pmf_mu[s], pmf_sigma[s], pmf_n[s], dist[s], truncate);
+    pmf = convolve(pmf, variable_pmf, len);
```
seabbs, line 85: For delay distributions we don't appear to be reversing the pmf anywhere. I'd suggest we add to here vs having in each function.

seabbs: I've surfaced these all into `combine_pmfs` which comes with a slight reduction in overhears but perhaps slightly harder to read code.

inst/stan/estimate_infections.stan
```diff
@@ -96,18 +96,27 @@ transformed parameters {
...
+    infections = generate_infections(
+      R, seeding_time, gt_pmf, initial_infections, initial_growth, pop,
+      future_time
+    );
   } else {
     // via deconvolution
-    infections = deconvolve_infections(shifted_cases, noise, fixed, backcalc_prior);
+    infections = deconvolve_infections(
+      shifted_cases, noise, fixed, backcalc_prior
+    );
```
sbfnk, line 109: Can we separate style commits vs code commits? It takes some looking to see what exactly has changed here in functionality and it's easy to miss something.

inst/stan/functions/convolve.stan
sbfnk: Perhaps we could call `_pmf` variables `_rev_pmf` when they're reversed? I found it quite hard to keep track of what was forward and what backward when reading the code in the past.

sbfnk: > I'd suggest we add to here vs having in each function.

Is this for readability or efficiency? Personally I found it easier to follow the model when `pmf` definitions were forward but that might of course just me me.

seabbs: its more efficient but also because it centralises where these transformations happen. Unless I missed something the changes in this branch led to delay distributions not being reversed and I think that was because there wasn't water tight logic on where this transformation should be happening. 

Yes can update the naming scheme.

inst/stan/functions/pmfs.stan
seabbs: So as this is referenced to 0 it is correct as is I think 🥳

sbfnk: I think the confusion here comes from the fact the `truncate` is probably not a great name for this argument, which boils down to how the discrete pmf is interpreted to arise from the continuous parameters, in particular whether the `1` bin reflects the interval `[0, 1]` (interpretation for delays, `truncate == 0`) or `[1, 2]` (interpretation for generation times, `truncate == 1`, out of necessity because 0 causes issues with the iterative approach of updating the renewal equation). Of course both aren't optimal and either way the distinction should probably be made clear somewhere to the user.

inst/stan/functions/convolve.stan
```diff
@@ -12,6 +13,23 @@ vector convolve(vector x, vector y, int len) {
...
 
+// convolve two vectors as a backwards dot product
+// y vector shoud be reversed
+// limited to the length of x and backwards looking for x indexes
+// designed for use convolve a case vector and a delay pmf
+vector convolve_dot_product(vector x, vector y, int len) {
+    int ylen = num_elements(y);
+    vector[len] z;
+    for (s in 1:len) {
+        z[s] = 1e-5 + // add a small value to ensure numerical stability
```
sbfnk, line 24: Is there a specific need for this? I noticed in testing that removing this can change outcomes a fair bit.

NEWS.md
```diff
@@ -7,13 +7,12 @@ Thanks to @seabbs, and @sbfnk and for [SACEMA](https://sacema.org) for hosting @
 ## Breaking changes
 
 * To enable enhance functionality `trunc_opts()` now takes a single argument (`dist`) which defines the truncation delay rather than a arbitary list of arguments (which was previously used to define the distribution.
-
+* Updated the handling of generation times in the renewal equation to be left truncation adjusted for the zeroth day. This more accurately the typical approach taken to estimate generation times but may slightly impact estimates vs those produced using previous versions.
```
sbfnk, line 10: As mentioned in [this comment](https://github.com/epiforecasts/EpiNow2/pull/343#discussion_r1084398737) I think it might be more accurate to say it's indexed such that the first bin of the discretised distribution reflects the range [1,2] of the continuous version (but then also extends to [n, n + 1] as the last bin).

inst/stan/functions/convolve.stan
seabbs: the `1e-5`? That was only there for numeric stability so may be able to remove. Not great if it has a large impact.

NEWS.md
seabbs: so that is true but there is also an adjustment for left truncation in there as well no?

sbfnk: yes, true

seabbs: Very happy to take a suggested reword. It also bears thinking about whether the left truncation adjust in particular is safe given it likely isn't used by most approaches that produce these estimates.

inst/stan/functions/convolve.stan
seabbs: I've dropped in [c63d92f](https://github.com/epiforecasts/EpiNow2/pull/343/commits/c63d92f36cb759097c04948cc638ddef7182f6da). Will play around with tomorrow.

seabbs: Played around with this a bit and it doesn't seem to make a major difference in terms of stability. Probably just a hang over of the bad old days code.

seabbs:
>  Make bounding on frac_obs optional and able to be specified by the user. Have the default be [0, 1] bounding and flag this breaking change in the NEWS.md.

I am currently thinking we should just flag this as a breaking change. I really don't think many of the use cases for it being > 1 are that great and in the case of the COVID pipeline I don't think it will break anything but will instead very slightly alter one of the models which doesn't seem like the biggest cost in the world.

R/opts.R
```diff
@@ -74,7 +74,7 @@ generation_time_opts <- function(..., disease, source, max = 15L,
       dot_options$dist <- "gamma"
     }
     ## set max
-    if (!("dist" %in% names(dot_options))) {
+    if (!("max" %in% names(dot_options))) {
```
sbfnk, line 78: don't think you need the if statement - `max` is a named argument so will never be in `names(dot_options)`

sbfnk:
> > Make bounding on frac_obs optional and able to be specified by the user. Have the default be [0, 1] bounding and flag this breaking change in the NEWS.md.
> 
> I am currently thinking we should just flag this as a breaking change. I really don't think many of the use cases for it being > 1 are that great and in the case of the COVID pipeline I don't think it will break anything but will instead very slightly alter one of the models which doesn't seem like the biggest cost in the world.

I agree (it will alter estimates like this one but only marginally)
https://github.com/epiforecasts/covid19-uk-nowcasts-projections/blob/main/format-forecast/posteriors/all/occupancy/2023-01-24.csv#L4

R/opts.R
seabbs: It will when you supply a list which is what happens if you use `get_generation_time()` and then use `generation_time_opts()` on that distribution. That might be an interface to phase out but can see it being common as it stands.

seabbs:
I think this is ready now and aside from checking #349 can go to CRAN without further changes.

R/opts.R
seabbs: (as an aside I think ever allowing the user to specify a free list is kind of a bad idea - making them call `dist_spec` (i.e your dist fn) might be a good idea to enforce the right structure etc from the get go.

sbfnk: Yes I was thinking similar given that I managed to confuse myself over my own code here. Could `trunc_opts`, `generation_time_opts` and `delay_opts` perhaps even eventually be axed and replaced by `dist_spec` calls (with a deprecation warning in this version)? From what I can see the only things preventing that are the backwards compatibility with specifying `disease` and `source` and the `prior_weight` option (which could perhaps go somewhere else or equally be scrapped).

seabbs (commented):
Changes in d3b45935c5246033da54da5009a82add119caece LGTM - nice.

R/estimate_truncation.R
```diff
@@ -2,47 +2,66 @@
...
+#' the adjusted latest observations in the time period of the earlier data set.
+#' These data sets are then compared to the earlier observations assuming a
+#' negative binomial observation model with an additive noise term to deal with
+#' zero observations.
 #'
 #' This model is then fit using `stan` with standard normal, or half normal,
-#' prior for the mean, standard deviation, 1 over the square root of the over dispersion
-#' and additive noise term.
+#' prior for the mean, standard deviation, 1 over the square root of the over
+#' dispersion and additive noise term.
```
sbfnk, line 29:
```suggestion
#' prior for the mean, standard deviation, 1 over the square root of the
#' overdispersion and additive noise term.
```
I think

sbfnk (commented):
Looks good. Good idea to add the `@author` field to give credit. Is there anyone else who may have contributed to any of the functions and would need to be acknowledged? 

My only hesitation is in the set up for the delay distributions. The arguments to be passed to `generation_time`, `delays`, and `truncation` are all of the same type (a delay distribution) but at the moment need to be invoked using different function, which could be avoided. See [here]() for an alternative implementation, removing some of the `_opts` model (a downside perhaps) but reducing these three options to all involve calls to `dist_spec` (a benefit). If we go further with it it needs a bit more checking, possibly improved code organisation (e.g. the code of munging delays is duplicated at the moment and could be shipped out into a function)  and updating the documentation. Related discussion at
https://github.com/epiforecasts/EpiNow2/pull/343#discussion_r1086458803

I'm also not 100% convinced we really need to support the mixture of fixed and uncertain delay distributions (which causes a lot of code overhead). It might be simpler to only support one or the other, in which case we could combine multiple delays in R instead of stan.

NEWS.md
```diff
@@ -1,26 +1,47 @@
...
+
+## Breaking changes
+
+* To enable enhance functionality `trunc_opts()` now takes a single argument (`dist`) which defines the truncation delay rather than a arbitrary list of arguments (which was previously used to define the distribution.
+* Updated the handling of generation times in the renewal equation to be left truncation adjusted for the zeroth day. This more accurately the typical approach taken to estimate generation times but may slightly impact estimates vs those produced using previous versions.
+* The range of the `frac_obs` parameter has restricted with an upper bound of 1 to reflect its name and description. This may impact a small number of edge case uses with the majority being models fit using `estimate_secondary()`. By @sbfnk in #340.
+
+## Features
+
+* Adds a new function `simulate_secondary()` for simulating secondary observations under the generative process model assumed by `estimate_secondary`.
```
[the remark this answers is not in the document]
seabbs: Yes good point.

seabbs:
I really like the dist_spec usage as an input and I think the s3 c() method is a good idea. I also like having seeding time be it’s own function. I am struggling a little with the amount of internal code replication and complexity this adds. We could obviously reduce this with some internal functions etc but they would still have to be reliably called on all inputs. I would also be a bit worried about future extensions using this approach (for example what is the input if you had time varying distributions) 

The main upside (beyond dist_spec usage as input which as I said is nice) is that the user doesn’t have to write an of the opts functions but the cost is more opaqueness in what is happening?

I’ll do some cleaning up of what you have implemented and see how it feels as I do that. 

I agree on the distribution mixing not being incredibly useful but perhaps as it is implemented we should stick with it and see if a use if found and if not depreciate?

sbfnk:
Yes agree on all of these. Perhaps a half way house would be to bring the _opts functions back but have all of them take an S3 delay_dist object as first argument? The post processing of e.g. delays could then move back into these functions. This would also remove the bed for dot arguments anywhere (as we can specify multiple delays with the c function).

seabbs:
> Yes agree on all of these. Perhaps a half way house would be to bring the _opts functions back but have all of them take an S3 delay_dist object as first argument? The post processing of e.g. delays could then move back into these functions. This would also remove the need for dot arguments anywhere (as we can specify multiple delays with the c function).

Yes I think we should definitely do this as a minimum. The `...` stuff is/was really not ideal

NEWS.md
```diff
@@ -1,26 +1,47 @@
...
 * Switched to `linewidth` in `plot_CrIs` rather than `size` to avoid issues with `ggplot2` 3.4.0.
+* Set up validation against synthetic data to run as a CI check.
+* Added tests for internal stan convolution functions.
+* Update all `get_` distribution functions to return the distribution as well as summary
+ parameters.
+
+## Documentation
+
+* Slight edits to the model outline for `estimate_infections()`.
+* Updated examples to make use of fixed distributions to improve run-times where appropriate.
```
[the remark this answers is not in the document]
seabbs: Dropped.

seabbs:
It would be nice if we can work out what needs to be done to close this out ASAP so we can release `1.3.4` and start making more modular additions for the big `2.0.0` release.

seabbs:
From slack chats my conclusion is we are happy to merge here and move towards a release today.

PR #345, opened 2023-01-18
 switch to directly calculated growth rate and extend model options, take 2

sbfnk:
Attempts to resurrect #307 as work in progress. Copying text from there:

This PR is meant to supersede #213 and #214. It 
- finalises the work commenced in #214 to support [three different models](https://github.com/epiforecasts/EpiNow2/pull/214#issuecomment-753024893)
- merges #214 into #213 and implements most of the suggestions made there by @hsbadr 
- rebases #213 onto current `master`
- works with default options and a simple example (the README)

Things left to do:
- [ ] expose and document the model options and ensure the stan model options are set consistently
- [ ] Update the README so that it works. 
- [ ] Remove `backcalc_opts()` from all input
- [ ] Remove `rt_opts()` from all input
- [ ] Move `generation_time_opts()` inside of `process_opts()` as an argument. 
- [ ] Check postprocessing works as expected
- [ ] Check that all examples and docs work as expected
- [ ] Document model based on update specification
- [ ] Get passing CRAN check if not already
- [ ] more extensive testing
- [ ] update NEWS


sbfnk:
To do for being able to revisit this: 
- [ ] rebase to `main`
- [ ] reduce the number of models to 2 (growth rate model is the same as renewal model with fixed generation time of 1)
and then go back to the items in the original issue

sbfnk:
Closing as most suggestions in here have now made it into `main` one way or the other (and we have diverged to extract any use from this)

PR #358, opened 2023-01-31
cmdstanr version

seabbs:
Just using this to test the new benchmarking implementation. Will close once that has run.


seabbs:
testing.

Issue #357, opened 2023-01-30
Update to use `dist_spec` as a common distribution specification device.


seabbs:
Yes agree on all of these. Perhaps a half way house would be to bring the _opts functions back but have all of them take an S3 delay_dist object as first argument? The post processing of e.g. delays could then move back into these functions. This would also remove the need for dot arguments anywhere (as we can specify multiple delays with the c function).

_Originally posted by @sbfnk in https://github.com/epiforecasts/EpiNow2/issues/343#issuecomment-1406305408_

sbfnk:
Addressed in #363

Issue #313, opened 2022-09-22
Fit to multiple time series and fold in secondary model


sbfnk:
It might be nice to add functionality to have multiple observation time series assuming that they are all reported from the same infection proces with a delay and scale. This would potentially lead to more informed reproduction number / incidence estimation, but also enable doing what is currently done in `estimate_secondary.stan` within the existing model in `estimate_infections.stan` (e.g. interpreting the ratio of `obs_scale` between observed time series of cases and deaths as CFR) without treating the primary observation time series as fixed.

sbfnk:
If https://github.com/epiforecasts/EpiNow2/pull/307 is eventually merged on could then perhaps also have covariates or a GP prior on e.g. the scaling factor.

sbfnk:
Proposed interface for specifying multiple observation processes (building on #363):

Multiple delays can already be named with
```r
delay = c(incubation_period = dist_spec(...), reporting_delay = dist_spec(...))
```

Observation processes could then similarly be concatenated
```r
obs = obs_opts(c(hospitalisations = obs_spec(...), deaths = obs_spec(...)
```
where `obs_spec` includes some of the existing `obs_opts`, i.e. distribution family, week effect etc. In addition it would have something like `delays = c("incubation_period", "reporting_delay")` referring to the relevant specified delays by name. The `scaling` option should probably be incorporated into `dist_spec` so they can be appropriately chained. Ideally this would also include some of the existing observation model of `estimate_secondary`, e.g. incidence/cumulative. It could also include the option of not specifying an observation process, in which case it wouldn't be used to constrain the model but still generated (e.g. if one wanted a forecast of a secondary outcome from a known delay/scaling but where no observation exists, or explore scenarios).

seabbs:
> In addition it would have something like delays = c("incubation_period", "reporting_delay") referring to the relevant specified delays by name.

This part seems key. The alternative would be to make `delay`  a list of lists but with each top level entry being named after the observation model. The upside of this is making it clear when the model is specified how the modules connect but the big obvious downside is lots of code would need to change. 

I am happy with the proposal of identifying delays in the observation module. Some thought is needed about what this would mean for identifying when delays are constant and when they are not and how you could do this if keeping the current processing of delays (i.e a list with features that describe that list vs a list of lists that has these features). 

> It could also include the option of not specifying a data set, in which case it wouldn't constrain the model but still generated 

[...]

PR #363, opened 2023-02-08
implement `dist_spec` interface

### LINKED ISSUE #356: Distribution interface
See https://github.com/epiforecasts/EpiNow2/pull/343#pullrequestreview-1269957931

sbfnk:
Updates the distribution interface to use the `dist_spec` S3 class. In the process also implements nonparametric distribution specifications (indicated `_np` vs. `_p` for parametric) and multiple generation times (e.g. for vector-bourne diseases). All fixed delay distributions are now calculated in R using the existing functionality in `dist_skel`.

Open questions:
- is `dist_spec` a good title (maybe clearer as `delay_dist` or perhaps just `dist`)?
- should `mean`, and `sd` for the lognormal be interpreted as the natural means/sd? I think this would be less confusing, and we could add `logmean` and `logsd` or `mu` and `sigma` as additional ways of specifying this in `lognormal_dist_def`, just as it is in `gamma_dist_def`.

This fixes #356 and #357. It should also pave the way to implement #313 in the future (as delay distributions can be named and could then be referenced by different observation specifications.


sbfnk:
Remaining error is the one that would be fixed by #366 but this didn't occur before so it may point to an inadverted change to the model.

seabbs:
Now all the explained things are resolved I'll have a proper look through + see if I can identify what is leading to the change in behaviour.

sbfnk:
> Now all the explained things are resolved I'll have a proper look through + see if I can identify what is leading to the change in behaviour.

Should be all fixed now.

sbfnk:
Looks like performance is worsening with this - will need to investiage.

seabbs:
> is dist_spec a good title (maybe clearer as delay_dist or perhaps just dist)?

I like `dist` and I can't think of a reason that using it would be a problem. 

> should mean, and sd for the lognormal be interpreted as the natural means/sd? I think this would be less confusing, and we could add logmean and logsd or mu and sigma as additional ways of specifying this in lognormal_dist_def, just as it is in gamma_dist_def.

Yes I think this is fine to add but it needs to be very clearly communicated that this change has happened or it will break lots of code. As I have said many times I think the default should be to use the actual parameters of the distribution as the default  as this will limit the information loss in a standard modelling pipeline. This should also be the case for the gamma distribution as well really and perhaps we can make that transition in this release as well. 

> pdates the distribution interface to use the dist_spec S3 class. In the process also implements nonparametric distribution specifications (indicated _np vs. _p for parametric) and multiple generation times (e.g. for vector-bourne diseases). All fixed delay distributions are now calculated in R using the existing functionality in dist_skel.

This is all nice. I need to go through the code in detail and kick some tires but in principle I am a big fan of the proposed changes.

R/create.R
```diff
@@ -500,37 +498,23 @@ create_stan_data <- function(reported_cases, generation_time,
...
-        out$truncation_sd <- array(
-          truncnorm::rtruncnorm(1,
-            a = 0,
-            mean = data$trunc_sd_mean,
-            sd = data$trunc_sd_sd * 0.1
-          )
-        )
-      }
-    }
+    out$delay_mean <- array(purrr::map2_dbl(
```
seabbs, line 501: Multiple repeated chunks here now that could/should really be coming from a small function.

R/dist.R
```diff
@@ -811,8 +852,13 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
 #' @param max Numeric, maximum value of the distribution. The distribution will
 #' be truncated at this value.
 #'
+#' @param pmf Numeric, defaults to an empty vector corresponding to a parametric
```
seabbs, line 855: This doesn't mention it being non-parametric etc.

R/dist.R
```diff
@@ -821,57 +867,242 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
 #' @author Sam Abbott
 #' @export
 dist_spec <- function(mean, sd = 0, mean_sd = 0, sd_sd = 0,
-                      dist = c("lognormal", "gamma"), max = NULL,
-                      fixed = FALSE) {
-  dist <- match.arg(dist)
+                      dist = c("lognormal", "gamma"), max,
```
seabbs, line 870: we may as well call this arg distribution (or change to family) as we aren't really getting much by shortening it

R/dist.R
```diff
@@ -821,57 +867,242 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
...
+          pmf <- dist_skel(
+            n = seq_len(max) - 1, dist = TRUE, cum = FALSE, model = dist,
+            params = params$params[[1]], max_value = max,
+            discrete = TRUE
+          )
+        }
+      }
+    } else { ## nonparametric fixed
+      if (!missing(max) && max > length(pmf)) {
+        pmf <- pmf[1:(max + 1)]
```
seabbs, line 954: This looks like it will cause an error from sight read.

R/dist.R
```diff
@@ -821,57 +867,242 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
...
+##' combined in the passed [dist_spec()].
+##'
+##' @param x The [dist_spec()] to use
+##' @param ... Not used
+##' @return A vector of means.
+##' @author Sebastian Funk
+##' @method mean dist_spec
+##' @importFrom utils head
+##' @export
+mean.dist_spec <- function(x, ...) {
```
seabbs, line 1034: This is nice to have.

R/estimate_infections.R
```diff
@@ -63,22 +63,26 @@
 #' reported_cases <- example_confirmed[1:60]
 #'
 #' # set up example generation time
-#' generation_time <- generation_time_opts(
+#' generation_time <- get_generation_time(
```
seabbs, line 66: this code line is probably getting whiplash

R/estimate_infections.R
```diff
@@ -63,22 +63,26 @@
...
 #' incubation_period <- get_incubation_period(
+#'   disease = "SARS-CoV-2", source = "lauer", fixed = TRUE
+#' )
+#' # delays between infection and case report, with uncertainty
+#' incubation_period_uncertain <- get_incubation_period(
 #'   disease = "SARS-CoV-2", source = "lauer"
 #' )
-#' reporting_delay <- list(
+#' reporting_delay <- dist_spec(
 #'   mean = convert_to_logmean(2, 1), mean_sd = 0,
```
seabbs, line 78: if you have changed mean and sd to be on the natural scale then this is likely wrong.

R/opts.R
```diff
@@ -32,151 +19,93 @@
...
+#' # A generation time sourced from the ligerature
+#' dist <- get_generation_time(disease = "SARS-CoV-2", source = "ganyani")
+#' generation_time_opts(dist)
+generation_time_opts <- function(dist, prior_weight = NULL) {
 
-  ## check to see if options have been passed in as a list
-  if (length(dot_options) == 1 && is.list(dot_options[[1]])) {
-    dot_options <- dot_options[[1]]
+  if (missing(dist)) {
+    dist <- dist_spec(mean = 1)
```
seabbs, line 39: why not just have this as the default?

R/opts.R
```diff
@@ -32,151 +19,93 @@
...
+         "to `dist_spec` or `get_generation_time`. ",
+         "This behaviour has changed from previous versions of `EpiNow2` and ",
+         "any code using it may need to be updated. For examples and more ",
+         "information, see the relevant documentation pages using ",
+         "`?generation_time_opts`")
   }
-  gt$weight <- prior_weight
-
-  names(gt) <- paste0("gt_", names(gt))
+  data <- unclass(dist)
```
seabbs, line 49: share code here with delay_opts that could easily be a helper function

R/opts.R
```diff
@@ -201,11 +130,20 @@ delay_opts <- function(..., fixed = FALSE) {
 #' trunc_opts()
 #'
 #' # truncation dist
-#' trunc_opts(dist = list(mean = 3, sd = 2))
-trunc_opts <- function(dist = list()) {
-  data <- do.call(dist_spec, dist)
+#' trunc_opts(dist = dist_spec(mean = 3, sd = 2, max = 10))
+trunc_opts <- function(dist = dist_spec()) {
+  if (!"dist_spec" %in% class(dist)) {
```
seabbs, line 135: shared code with delay_opts and generation_time_opts - could be a helper function

inst/stan/data/delays.stan
```diff
@@ -1,14 +1,13 @@
...
-  real delay_mean_mean[delays];// prior mean of mean incubation period
-  real delay_sd_mean[delays];  // prior sd of sd of incubation period
-  real delay_sd_sd[delays];    // prior sd of sd of incubation period
-  int<lower = 1> delay_max[delays];       // maximum incubation period
-  int delay_dist[delays];       // 0 = lognormal; 1 = gamma
+  int<lower = 0> delay_n;                  // number of delay distribution distributions
+  int<lower = 0> delay_n_p;                // number of parametric delay distributions
+  int<lower = 0> delay_n_np;                // number of nonparametric delay distributions
+  real delay_mean_mean[delay_n_p]; // prior mean of mean delay distribution
+  real<lower = 0> delay_mean_sd[delay_n_p];   // prior sd of mean delay distribution
```
seabbs, line 5: rather than differentiating between delays and gts we could just pass them all together and then have indices that pull out the ones we need for any given step. This would make things a bit shorter and more concise + prepare the ground for fitting to multiple observations. 

It would mean that the generation time wasn't present as a clear parameter in the model output (with this also being true for the other delays) and this might make development harder and diagnosing issues with the model. That could be dealt with by having better post-processing which we would also need if we were going to have multiple observations.

seabbs: Same point about truncation distributions etc..

seabbs: Doing this would require changing how pre-processing works and the design of that could be so painful as to not make doing this worthwhile.

inst/stan/data/observation_model.stan
```diff
@@ -3,17 +3,22 @@
...
   int obs_scale;                     // logical controlling scaling of observations
   real obs_scale_mean;               // mean scaling factor for observations
   real obs_scale_sd;                 // standard deviation of observation scaling
   real obs_weight;                   // weight given to observation in log density
   int likelihood;                    // Should the likelihood be included in the model
   int return_likelihood; // Should the likehood be returned by the model
+
+  int<lower = 0> trunc_n;                  // number of trunc distribution distributions
+  int<lower = 0> trunc_n_p;                // number of parametric trunc distributions
+  int<lower = 0> trunc_n_np;                // number of nonparametric trunc distributions
```
seabbs, line 15: Noting that this would also have to be changed to support multiple observations (see #313) and either be specified as a nested list or given to the observation model. 

A much bigger potential redesign would be to make delays and truncations be nested in each observation model. This could have quite a few benefits but would come at the cost of needing quite a bit of redesign work.

inst/stan/data/simulation_delays.stan
```diff
@@ -1,5 +1,11 @@
-  int delays;                           // no. of delay distributions
-  real delay_mean[n, delays];           // mean of delays
-  real delay_sd[n, delays];             // sd of delays
-  int delay_max[delays];                // maximum delay
-  int delay_dist[delays];                // 0 = lognormal, 1 = gamma
+  int<lower = 0> delay_n;                  // number of delay distribution distributions
```
seabbs, line 1: we could consider moving to a chunk based model for some of these repeated scripts to avoid all this copy and pasting - comes at the cost of model readability. Could also make a simple generator that we run to build the stan code but this would easily lead to dev problems I think (with things getting out of sync).

inst/stan/estimate_infections.stan
```diff
@@ -30,29 +30,31 @@ transformed data{
...
-    );
-  }
-  if (n_fixed_delays) {
-    delays_fixed_pmf = combine_pmfs(
-      to_vector([ 1 ]), delay_mean_mean[fixed_delays],
-      delay_sd_mean[fixed_delays], delay_max[fixed_delays], 
-      delay_dist[fixed_delays], delay_max_fixed, 0, 0
-    );
-  }
+  vector[gt_max_fixed] gt_fixed_pmf;
```
seabbs, line 43: this is quite a simple step - we could be doing this in the `c` method of dist vs in the stan model..

inst/stan/estimate_infections.stan
```diff
@@ -187,28 +188,32 @@ generated quantities {
...
+    real gen_gt_var = pmf_var(gt_rev_pmf, 1, 1, gen_gt_mean);
+    r = R_to_growth(R, gen_gt_mean, gen_gt_var);
   } else {
     // sample generation time
-    real gt_mean_sample[1];
-    real gt_sd_sample[1];
-    vector[gt_max[1]] gt_rev_pmf;
+    real gt_mean_sample[gt_n_p];
+    real gt_sd_sample[gt_n_p];
+    vector[gt_max_total] gen_rev_pmf;
```
seabbs, line 198: I am bit lost why we are being inconsistent with prefixs here? Happy to just call everything `gen_` vs switching?

inst/stan/functions/pmfs.stan
```diff
@@ -54,10 +53,24 @@ vector combine_pmfs(vector fixed_pmf, real[] pmf_mu, real[] pmf_sigma, int[] pmf
...
+    }
+  }
+  if (left_truncate) {
+    real norm;
+    for (i in 1:left_truncate) {
+      pmf[i] = 0;
+    }
+    norm = sum(pmf);
+    if (left_truncate < len) {
+      for (i in left_truncate:len) {
```
seabbs, line 70: this seems like it could be causing a slow down can it now be vectorised using `rep_vector` and `{}` to construct a new combined vector?

inst/stan/simulate_infections.stan
```diff
@@ -23,7 +23,23 @@ data {
 }
 
 transformed data {
-  int delay_max_total = sum(delay_max) - num_elements(delay_max) + 1;
+  int gt_max_fixed =
+    num_elements(gt_np_pmf) - num_elements(gt_np_pmf_groups) + 1;
+  int gt_max_total = gt_max_fixed + sum(gt_max) - num_elements(gt_max);
+  int delay_max_fixed =
```
seabbs, line 29: arrrrrhhh code duplication (agree hard to avoid)

seabbs: I have no idea if chunks are only compiled once. I assume not but if they are moving to chunks could drastically reduce compile times

R/create.R
pearsonca: `truncnorm` is vectorized, and this is generating a scalar anyway, so the `array(purrr::map2_dbl(...))` seems overkill.

imo, the nice syntax here would be `with(X, truncnorm(1, a = 0, mean = mean, sd = sd))`. that would mean structuring data differently. more like `data$delay$mean$mean, data$delay$mean$sd, ...`.

that sort of approach would also be in the direction of something like `with(data$delay$mean, do.call(rdistro, c(args, list(n=1)))` which is the sort of maximum flexibility option.

pearsonca: given how systematically names are structured currently, this should be an `apply`-family-able operation, if structured correctly. i.e. "for all of these set of names (delay, trunc, gt): for their sub-elements (mean, sd), use the sub-sub elements (mean, sd) this way"

R/opts.R
```diff
@@ -32,151 +19,93 @@
...
-      dot_options$max <- max
-    }
-    ## set default of mean=1 for backwards compatibility
-    if (!("mean" %in% names(dot_options))) {
-      dot_options$mean <- 1
-    }
-    dot_options$fixed <- fixed
-    gt <- do.call(dist_spec, dot_options)
+  if (!"dist_spec" %in% class(dist)) {
+    stop("The generation time distribution must be of given either using a call ",
```
pearsonca, line 42: grammar issue here.

pearsonca: also seems clearer to use `is()` or `inherits()` etc here than `... %in% class(...)` idiom

sbfnk:
> As I mention in the comments below we could go further in unifying distributions across delays, generation times, and truncations etc and if we did this now it could save us pain in the future.

I agree, this would be nice. And since we'll have to somehow reference to delays by index (and then pre-combine for certain combination of indices) if we want to address #313 perhaps the additional pain from doing so isn't too great.

sbfnk:
> I also really think we want to be combining fixed PMFs in R vs in stan now we have code to generate them as this will reduce model complexity and make it easier to test everything is working as expected (is this already happening? I think not but may have missed).

I agree that this would have both of these advantages. The potential downside is that if we implement #313 the ability to do so may change for each observation model so this information would have to be accessible somehow at the point of combination (and thus couldn't be done e.g. in `c.dist()`. 

As an example, say we have observations of hospitalisations and deaths, with a fixed distribution incubation period, a fixed distribution of the symptom onset -> hospitalisation delay and an uncertain symptom onset -> death delay. The delays could be combined in advance for hospitalisations but not for deaths.

R/estimate_infections.R
sbfnk: For now they're unchanged but I think we should rename them logmean/logsd or mu/sigma here (with mean/sd acquiring their natural meanings).

R/create.R
seabbs: This sounds like someone would love to write a little helper function to improve our DRY  🤡

R/estimate_infections.R
seabbs: very much agree on renaming and agree we should/can support the natural scale input with an internal transform. I think we should doc that that isn't an ideal approach though.

seabbs:
> I agree that this would have both of these advantages. The potential downside is that if we implement https://github.com/epiforecasts/EpiNow2/issues/313 the ability to do so may change for each observation model so this information would have to be accessible somehow at the point of combination (and thus couldn't be done e.g. in c.dist().

Ah that is only the case if we stick with a list of delays and then link by name in the observation model. If we take one of the other design approaches then we can do this (and this perhaps indicates it might have some benefits).

R/create.R
pearsonca: ...trying to get me fired for distraction by shiny objects.

seabbs: ah it will just be a formal reprimand - got to be worth it for all the fame and adulation this will bring.

sbfnk: removed `map2_dbl` call in https://github.com/epiforecasts/EpiNow2/pull/363/commits/145c3f8a20492fd35b5b1059f0d84d81be083a9b

R/dist.R
sbfnk: Resolved in https://github.com/epiforecasts/EpiNow2/pull/363/commits/a3815887b399e1a72d886a3f7a590ccb99dcb22f

R/dist.R
sbfnk: Resolved in https://github.com/epiforecasts/EpiNow2/pull/363/commits/c99a07f178afda437b19c82b7090439b57e8eff9

R/dist.R
sbfnk: Fixed in https://github.com/epiforecasts/EpiNow2/pull/363/commits/25232c7da75861f497dda68dbb2e580be2da3cf4

R/opts.R
sbfnk: Done in https://github.com/epiforecasts/EpiNow2/pull/363/commits/6977eaa09ec1c9d764d9ed2d0293ed24a9ecba03

R/opts.R
sbfnk: Done in https://github.com/epiforecasts/EpiNow2/pull/363/commits/bf455251037ccea9f31ec2196fc9304a3e218d89

inst/stan/functions/pmfs.stan
sbfnk: Something like this? https://github.com/epiforecasts/EpiNow2/pull/363/commits/d4e89c2674e3c929c7554c74c249292164cca30d

R/opts.R
sbfnk: Done in https://github.com/epiforecasts/EpiNow2/pull/363/commits/c81b39f1b8efb17b2d0d392c644f20c7c7204775

R/opts.R
sbfnk: Done in https://github.com/epiforecasts/EpiNow2/pull/363/commits/bf455251037ccea9f31ec2196fc9304a3e218d89

inst/stan/simulate_infections.stan
sbfnk: no, from looking at `src` after compilation I think unfortunately these are not compiled into separate object but instead it's separate compilation per model in `inst/stan`

inst/stan/data/simulation_delays.stan
sbfnk: I agree, still quite a repetition. Will put in a separate issue.

inst/stan/data/observation_model.stan
sbfnk: This now has a single delay vector as a first step towards this.

R/estimate_infections.R
sbfnk: This would potentially require a larger rethink - see #382.

inst/stan/estimate_infections.stan
sbfnk: Done now in a new `+` method for `dist_spec` (extending the now internal `c` method that won't convolute - needed to put different delays together that shouldn't be combined, e.g. generation time and reporting delays).

inst/stan/functions/pmfs.stan
seabbs: yes

sbfnk:
Touchstone [failing](https://github.com/epiforecasts/EpiNow2/actions/runs/4830073966/jobs/8605861778?pr=363#step:2:7370) presumably because of the syntax change in https://github.com/epiforecasts/EpiNow2/blob/eeec1e62ded8b6374dd79506c4efb6f71bddcce3/touchstone/setup.R#L20.

Not sure what the solution is for this, if any.

seabbs:
> It's masking a stats function which seems like a bad idea.

That does seem like a bad idea

sbfnk:
Synthetic validation currently looking like this suggesting there's a problem.
image

sbfnk:
> Synthetic validation currently looking like this suggesting there's a problem (apart from lots of infections).

A bit more investigation has revealed that this is a problem shared between this PR and `main` so I think is beyond the scope of the PR. It seems to be related to issues with identifiability of the `frac_obs` parameter.

Here's a fig comparing estimated infections between PR and main, and GP and backcalc models. Black is the truth from the simulation model.

image

These are the posteriors, black dots are the truth from the simulation model.

image

sbfnk:
This is ready for review. Any suggestions on further mitigating the loss in speed would be particularly welcome.

inst/stan/functions/convolve.stan
```diff
@@ -1,29 +1,16 @@
...
+    int xlen = num_elements(x);
     int ylen = num_elements(y);
     vector[len] z;
     for (s in 1:len) {
-        z[s] = dot_product(
-            x[max(1, (s - ylen + 1)):s], tail(y, min(ylen, s))
-          );
+      z[s] = dot_product(
+        x[max(1, (s - ylen + 1)):min(s, xlen)],
+        y[max(1, ylen - s + 1):min(ylen, ylen + xlen - s)]
```
seabbs, line 12: does this slow down vs tail? Can tail not still be used here?

seabbs: Yeah I like the extra features. Trying to understand why it's the models without delays that suffer the big performance hits. Well small but non-trivial.

sbfnk: [Profiling](https://github.com/epiforecasts/EpiNow2/blob/cmdstanr/inst/stan/estimate_infections.stan) suggests that everything involving convolutions is quite drastically slowed down.

PR:
```
                    name total_time
 1:                    R       4.30
 2:           assign max       0.00
 3:      day of the week       1.43
 4:               delays       3.63
 5:            delays lp       0.10
 6: generated quantities       0.01
 7:                gp lp       0.51
 8:                   gt       3.45
 9:           infections      10.31
10:            report lp       5.47
11:              reports      14.75
12:                rt lp       0.69
13:            update gp       4.47
```

main:
```
                    name total_time
 1:                    R       3.59
 2:     assign fixed pmf       0.00
 3:           assign max       0.00
 4:      day of the week       1.14
 5:               delays       0.59
 6:            delays lp       0.11
 7: generated quantities       0.01
 8:                gp lp       0.30
 9:                   gt       0.52
10:                gt lp       0.07
11:           infections       8.33
12:            report lp       4.58
13:              reports       6.01
14:                rt lp       0.53
15:             trunc lp       0.08
16:            update gp       3.36
```

sbfnk: No delays model.

PR:
```
                    name total_time
 1:                    R       6.70
 2:           assign max       0.00
 3:      day of the week       2.37
 4:            delays lp       0.16
 5: generated quantities       0.01
 6:                gp lp       0.66
 7:                   gt       5.27
 8:           infections      16.54
 9:            report lp       8.84
10:              reports       0.24
11:                rt lp       1.08
12:            update gp       6.53
```

Main:
```
                    name total_time
 1:                    R       6.00
 2:     assign fixed pmf       0.00
 3:           assign max       0.00
 4:      day of the week       2.19
 5:               delays       0.19
 6:            delays lp       0.19
 7: generated quantities       0.01
 8:                gp lp       0.77
 9:                   gt       0.87
10:           infections      13.43
11:            report lp       7.90
12:              reports       0.47
13:                rt lp       0.73
14:             trunc lp       0.13
15:            update gp       6.18
```

sbfnk: > But I can check if a reformulation yields performance benefits.

Doesn't look like it so will revert to previous version and try something else.

seabbs: Not ideal at all. Will take a close look on Friday if not before

seabbs: So the key targets are the delay and infections profiling block with some commonality between them (i.e. likely convolutions) leading to the slowdown. I wonder how if any of this is just allocation costs etc and the difference decreases as you run larger models (not ideal but better)

seabbs:
>  be related to issues with identifiability of the frac_obs parameter.

So this has never been identifiable has it? It has historically just drawn from the prior. Unless I am confused here?

inst/stan/functions/convolve.stan
sbfnk: It's a little bit faster with the latest changes - will profile again.

sbfnk:
> So this has never been identifiable has it? It has historically just drawn from the prior. Unless I am confused here? If its now all over the place in the GP model that doesn't seem ideal.

Yes though what is a bit worrying is that it doesn't return the prior (mean 0.1) and especially the posteriors from the initial fit to the example return really tight posteriors a long way from the prior (the black dots in the fig). Can you remember/remind me where `example_confirmed` comes from?

seabbs:
> Yes though what is a bit worrying is that it doesn't return the prior (mean 0.1) and especially the posteriors from the initial fit to the example return really tight posteriors a long way from the prior (the black dots in the fig). Can you remember/remind me where example_confirmed comes from?

Yes this is worrying indeed and hasn't historically been the case. 

`example_confirmed` is early Italy data from 2020 COVID-19

seabbs:
Good progress on the speed ups. Narrowing the gap. 

We could include the in model profiling as a `touchstone` benchmark. Would have to do it running key functions from R which is a pain vs the integrated cmdstanr profiling but still might be useful

inst/stan/functions/convolve.stan
sbfnk: Here's another comparison of time spent in different parts of the code (using cmdstan profiler) based on 100 chains. Not clear to me at all why e.g. more time seems would be spent calculating the reporting log-prior which hasn't been touched at all.

```
                    name mean.main mean.PR sd.main sd.PR
 1:                total     22.00   29.05    2.70  4.50
 2:            report lp      6.28    8.04    0.76  1.25
 3:           infections      6.07    8.83    0.75  1.36
 4:                    R      3.44    4.44    0.43  0.70
 5:            update gp      3.14    3.92    0.39  0.62
 6:      day of the week      1.11    1.41    0.14  0.21
 7:                   gt      0.55    1.41    0.07  0.22
 8:              reports      0.37      NA    0.05    NA
 9:                gp lp      0.37    0.46    0.05  0.08
10:                rt lp      0.33    0.47    0.04  0.07
11:            delays lp      0.17    0.08    0.03  0.01
12:               delays      0.10      NA    0.01    NA
13:                gt lp      0.05      NA    0.01    NA
14: generated quantities      0.01    0.01    0.00  0.00
15:     assign fixed pmf      0.00      NA    0.00    NA
16:           assign max      0.00    0.00    0.00  0.00
```

sbfnk:
Having done some more internal profiling I think the loss in speed comes from the overhead of passing more arguments to the function, i.e.

```
vector get_delay_rev_pmf(
  int delay_id, int len, int[] delay_types_p, int[] delay_types_id,
  int[] delay_types_groups, int[] delay_max,
  vector delay_np_pmf, int[] delay_np_pmf_groups,
  real[] delay_mean, real[] delay_sigma, int[] delay_dist,
  int left_truncate, int reverse_pmf, int cumulative
)
```

vs.

```
vector combine_pmfs(
  vector fixed_pmf, real[] pmf_mu, real[] pmf_sigma, int[] pmf_n, 
  int[] dist, int len, int left_truncate, int reverse_pmf
)
```

as well as some of the added overhead in code from allowing a mixture of fixed/uncertain delays in any order (and order would matter in case we ever wanted time-varying delays).

I don't see any obvious way to speed this up further so the decision may have to be whether the loss in performance is worth the added flexibility, avoidance of duplication and potential for future functionality (unless someone else finds a way to further speed up the model in this PR).

seabbs:
Ah okay this makes sense. Is the loss of speed absolute or does it scale as inference time increases?

pearsonca:
Why would passing more arguments have such an impact? Cursory inquiry into stan's semantics suggests this should be pass-by-reference (https://mc-stan.org/docs/2_18/reference-manual/parameters-are-constant.html) - can't imagine that minting the additional references is that expensive.

R/dist.R
```diff
@@ -829,67 +874,306 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
...
+##'
+##' This combines the parameters so that they can be fed as multiple delay
+##' distributions to [epinow()] or [estimate_infections()].
+##'
+##' @param ... The delay distributions (from calls to [dist_spec()]) to combine
+##' @return Combined delay distributions (with class [dist_spec()]`)
+##' @author Sebastian Funk
+##' @method c dist_spec
+##' @importFrom purrr transpose map
+c.dist_spec <- function(...) {
```
seabbs, line 1074: Have you added tests for this? In particular the convolution of PMFs

R/dist.R
```diff
@@ -829,67 +874,306 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
...
+##' combine
+##' @param e2 The second delay distributions (from a calls to [dist_spec()])
+##' to combine
+##' @return Delay distributions representing the sum of the two delays
+##' (with class [dist_spec()]`)
+##' @author Sebastian Funk
+##' @method + dist_spec
+##' @importFrom stats convolve
+##' @export
+`+.dist_spec` <- function(e1, e2) {
```
seabbs, line 1040: Have you added tests of this?

R/estimate_infections.R
```diff
@@ -28,6 +28,10 @@
 #' options. See the documentation of `delay_opts()` and the examples below for
 #' details.
 #'
+#' @param truncation A casll to `trunc_opts()` defining the truncation of
```
seabbs, line 31: typo casll to call

R/get.R
```diff
@@ -301,3 +301,27 @@ get_regions_with_most_reports <- function(reported_cases,
...
+
+##' Estimate seeding time from delays and generation time
+##'
+##' The seeding time is set to the mean of the specified delays, constrained
+##' to be at least the maximum generation time
+##' @param delays Delays as specified using `dist_spec`
+##' @param generation_time Generation time as specified using `dist_spec`
+##' @return An integer seeding time
+##' @author Sebastian Funk
+get_seeding_time <- function(delays, generation_time) {
```
seabbs, line 313: tests to cover this?

seabbs:
Added a basic dist plotting function in [7bbda02](https://github.com/epiforecasts/EpiNow2/pull/363/commits/7bbda02edc00b286a78e7385ed22af19447add4e) to help the review process

seabbs (approved):
LGTM to me but I think @sbfnk should also now review given the substantial number of changes I have made. 

The key change is to add a tolerance arg for +.dist_spec so that combined pmfs are optimised to remove very small numbers. 

Not very happy about our performance hit here but as not clear where its coming from we should make an issue and address that there.

seabbs:
@sbfnk this is waiting on your review now. 

Note I had to bump up fitting settings for `estimate_secondary` in order to get the tests to pass which again indicates something about the new implementation is likely a little unstable/non-optimal.

seabbs:
@sbfnk waiting on you now I think. I have had to crank adapt delta on some of the unit tests which isn't a great sign in terms of numerical stability. Did you see that in your testing?

R/create.R
```diff
@@ -675,3 +635,53 @@ create_stan_args <- function(stan = stan_opts(),
...
+##' Create delay variables for stan
+##'
+##' @param ... Named delay distributions specified using `dist_spec()`.
+##' The names are assigned to IDs
+##' @param ot Integer, number of observations (needed if weighing any priors)
+##' with the number of observations
+##' @return A list of variables as expected by the stan model
+##' @importFrom purrr transpose map
+##' @author Sebastian Funk
+create_stan_delays <- function(..., ot) {
```
sbfnk, line 648:
```suggestion
##' @author Sebastian Funk
##' @internal
create_stan_delays <- function(..., ot) {
```

R/estimate_infections.R
```diff
@@ -695,17 +707,20 @@ format_fit <- function(posterior_samples, horizon, shift, burn_in, start_date,
...
-      is.na(date), NA_character_,
-      default = "estimate"
+      data.table::fifelse(
+        date > (max(date, na.rm = TRUE) - horizon - shift),
+        "estimate based on partial data",
+        "estimate"
       )
+    )
   ]
+  # nolint end
```
sbfnk, line 723: unrelated - left over from merging/rebasing `main`?

R/get.R
sbfnk, line 313:
```suggestion
##' @author Sebastian Funk
##' @internal
get_seeding_time <- function(delays, generation_time) {
```

R/opts.R
```diff
@@ -119,65 +59,27 @@ generation_time_opts <- function(..., disease, source, max = 15L,
...
-#' # A single delay where we override the uncertainty assumption
-#' delay_opts(delay, fixed = TRUE)
-#'
-#' # A delay where uncertainty is implict
-#' delay_opts(list(mean = 1, mean_sd = 0, sd = 0.5, sd_sd = 0, max = 15))
+#' # A single delay without uncertainty
+#' delay <- dist_spec(mean = 1, sd = 0.5, max = 15)
+#' delay_opts(delay)
 #'
 #' # Multiple delays
```
sbfnk, line 69:
```suggestion
#' # Multiple delays (in this case twice the same)
```

R/dist.R
```diff
@@ -829,67 +880,515 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
...
+    group_starts <- c(1L, cumsum(delays$np_pmf_length) + 1L)
+    for (i in seq_len(length(group_starts) - 1L)) {
+      new_pmf <- stats::convolve(
+        new_pmf,
+        rev(delays$np_pmf[seq(group_starts[i], group_starts[i + 1L] - 1L)]),
+        type = "open"
+      )
+    }
+    if (!is.infinite(tolerance)) {
+      if (length(new_pmf) > 1) {
```
sbfnk, line 1092:
```suggestion
    if (tolerance > 0 && length(new_pmf) > 1) {
```

R/dist.R
```diff
@@ -829,67 +880,515 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
...
+#'
+#' This combines the parameters so that they can be fed as multiple delay
+#' distributions to [epinow()] or [estimate_infections()].
+#'
+#' @param ... The delay distributions (from calls to [dist_spec()]) to combine
+#' @return Combined delay distributions (with class [dist_spec()]`)
+#' @author Sebastian Funk
+#' @method c dist_spec
+#' @importFrom purrr transpose map
+#' @export
```
sbfnk, line 1151:
```suggestion
#' @internal
```
so as to avoid confusion with `+` (which is the external-facing function)

sbfnk:
> @sbfnk waiting on you now I think. I have had to crank adapt delta on some of the unit tests which isn't a great sign in terms of numerical stability. Did you see that in your testing?

Not really - does this cause an error or just a warning somewhere? I've looked through this again and it looks good to me, apart of course from the performance hit.

sbfnk:
> I'm also very unhappy about the synthetic recovery of infections as that is definitely a new issue vs legacy. That indicates that a bug has crept in somewhere to me and that is not ideal.

I agree it's not ideal - though I think could also be down to the specific scenario I've used for testing this. Definitely worth exploring further.

R/dist.R
seabbs: This prevents users being able to do things like plot(c(lognormal, lognormal + lognormal)) which does seem handy.

R/estimate_infections.R
seabbs: Fixed in https://github.com/epiforecasts/EpiNow2/pull/363/commits/967d3cdb29b71daff6488d2c1529f53fb56e35f5

R/dist.R
seabbs: Dropped in https://github.com/epiforecasts/EpiNow2/pull/363/commits/749dc8e03b0cedbd990406662929bd80c2f80a61

seabbs:
> does this cause an error or just a warning somewhere? 

No it doesn't.

PR #364, opened 2023-02-08
Set C++17 as a system requirement

seabbs:
As per the new CRAN requirements this PR sets C++17 as a package requirement. The adjustment made here is based on recommendations in [this](https://discourse.mc-stan.org/t/cran-note-for-package-using-stan-suggest-c-17/30247/17) thread.


seabbs:
Checking devel I see that the C++14 note has been resolved and that all CI tests etc are passing as expected. Just waiting on synthetic validation then will merge and resubmit to CRAN.

seabbs:
(@sbfnk tagged you for forms sake but I think this doesn't really need a review if all checks are passing)

seabbs:
Synthetic validation all looks good to me.

seabbs:
This looks good to me. Will merge into main now and resubmit tomorrow evening as long as no issues show up.

PR #366, opened 2023-02-08
allow edge case of negative binomial with 0 mean

sbfnk:
This is how R handles it and I see no reason not to support this here instead of throwing an error

```r
> rnbinom(n = 1, mu = 0, size = 1)
[1] 0
```


seabbs:
I think stan sees a reason (for whatever reason):  "Error : Exception: Exception: neg_binomial_2_rng: Location parameter is 0, but must be > 0!  (in 'functions/observation_model.stan' at line 117; included from 'model_estimate_infections' at line 7)"

seabbs:
Sorry, my point was why does it do this and not follow the behaviour of R.

seabbs (approved):
This looks good I think. Just needs a news file entry before it can be merged..

PR #370, opened 2023-02-20
deal with zeroes in generation times

sbfnk:
Bulding on #363 to give the user options for what to do with the generation time.


seabbs (approved):
This looks good. We discussed extending this to the delays more generally but after seeing the implementation I am not sure how that would work. 

Only thing I think is missing here is a news item flagging this change.

sbfnk:
I'm tempted to ditch this and instead keep a flexible option for left truncation (which seems the most principled approach to me compared to shifting or loading). The other option would be to keep this and have an additional argument that indicates amount of left truncation/shifting/loading.

seabbs:
I agree regarding left truncation though am not sure I have a good enough handle on how people are best practice estimating Gts and hence what assumptions they may be making

seabbs:
Are we closing in favour of adaptive left truncation?

sbfnk:
> Are we closing in favour of adaptive left truncation?

I'm so inclinced but don't feel particularly strongly.

PR #372, opened 2023-03-08
1.3.5 minor release

seabbs:
This is a minor release to deal with compilation issues on CRAN. Currently testing on `rhub`.


seabbs:
@sbfnk any idea why touchstone is failing now? I see some minor differences between main and develop for the touchstone configuration with I think those on develop being the more recent?

seabbs:
😱

sbfnk (approved):
checks, rhub, windevel and touchstone all look fine now - time to get back to CRAN?

seabbs:
`rstan` is in the middle of being updated on CRAN and it sounds like they expect it to be soon. Perhaps we should wait for this?

sbfnk:
I think we should go ahead now - who knows how long they'll take.

seabbs:
> I think we should go ahead now - who knows how long they'll take.

This has happened it seems. It looks like they now support `2.26`. This might mean they are on the pathway to being up to date.

sbfnk:
Exciting!!

PR #374, opened 2023-04-14
do not merge -- try to fix failing R CMD CHECK

sbfnk:
by stabbing in the dark


sbfnk:
It seems to be the [R commands](https://github.com/epiforecasts/EpiNow2/blob/048c71d61cb241dfab137c374451edc024db168b/configure#L7) in the `configure` file that break this - specifically `src/Makevars.win` not being found which to me is a complete enigma because it would imply that `src/Makevars` is found.

@seabbs did this solution come from somewhere else? And given that this is a shell script already, is there any reason not to do the editing of `Makeavars` using shell commands here in a random attempt to fix this?

seabbs:
So it came from a thread on the CRAN requirements change. It looks like however that this could be cause by the `rstantools` update which should have fixed the original issue (and now looks to be on CRAN)

seabbs:
well damn - that didn't fix it.

sbfnk:
Closing as all seems fine now (🤞).

PR #387, opened 2023-05-11
Fix example in `create_backcalc_data()`

### LINKED ISSUE #385: Fix examples in `create_backcalc_data()`
The current examples use `create_gp_data()` but should be `create_backcalc_data()`.

jamesmbaazam:
This PR replaces the previous example in `create_backcalc_data(). The current example was calling `create_gp_data()` instead of `create_backcalc_data()`.


jamesmbaazam:
This was a mistake as I mistakenly branched off my previous branch rather than main. Closing!

PR #389, opened 2023-05-11
change the default value of sample in `dist_fit()`

jamesmbaazam:
In the current design of `dist_fit()`, the argument `sample` is supplied with a default value of `NULL`, which is checked and replaced with 1000 if it's `NULL` or < 1000. I think that means it should have a default value of 1000 with a check to see if it's < 1000 when the default is overridden with a supplied value. 

In this PR, I change the default value of `samples` to 1000. If the value of `samples` is supplied, the function checks if `samples` is <1000, and if so, sets it to 1000 and throws a warning to inform the user.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,10 @@ This release is in development. For a stable release install 1.3.5 from CRAN.
   instead of `create_gp_data()`. By @jamesmbaazam in #388 and reviewed by @seabbs.
 * Improved compilation times by reducing the number of distinct stan models and deprecated `tune_inv_gamma()`. By @sbfnk in #394 and reviewed by @seabbs.
 * Changed touchstone settings so that benchmarks are only performed if the stan model is changed. By @sbfnk in #400 and reviewed by @seabbs.
+* `dist_fit()`'s `samples` argument now takes a default value of 1000 instead
+  of NULL. If a supplied `samples` is less than 1000, it is changed to 1000
+  and a warning is thrown to indicate the change. By @jamesmbazam in #389 and
+  reviewed by @seabbs.
 
 # EpiNow2 1.3.5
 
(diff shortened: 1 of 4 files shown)


R/dist.R
```diff
@@ -197,16 +198,16 @@ dist_skel <- function(n, dist = FALSE, cum = TRUE, model,
...
                      chains = 2, dist = "exp", verbose = FALSE) {
-  if (is.null(samples)) {
-    samples <- 1000
-  }
-
   if (samples < 1000) {
     samples <- 1000
   }
 
+  warning(sprintf("%s %s", "`samples` must be at least 1000.",
```
seabbs, line 207: This needs to be inside 203 to 205 to it will always trigger.

seabbs (changes requested):
Thanks, James, 

See the note about the warning message.

This needs:

- [ ] To be rebased on main
- [ ] To be redoc'd
- [ ] To have a news item
- [ ] To have a dev version increment.

Then good to merge.

R/dist.R
jamesmbaazam: That's true. Will change asap.

sbfnk:
Last missing task I think is

> [ ] To have a dev version increment.

jamesmbaazam:
I believe this was done here 58db941bf093ccfed1d86e15e792c2cc72475b8b.

seabbs:
That isn't in this PR?

jamesmbaazam:
> Last missing task I think is
> 
> > [ ] To have a dev version increment.

This has been addressed in e91568b75d66ffc9d9b01ad1b95dd343b03f371c.

NEWS.md
```diff
@@ -16,6 +16,10 @@ This release is in development. For a stable release install 1.3.5 from CRAN.
   instead of `create_gp_data()`. By @jamesmbaazam in #388 and reviewed by @seabbs.
 * Improved compilation times by reducing the number of distinct stan models and deprecated `tune_inv_gamma()`. By @sbfnk in #394 and reviewed by @seabbs.
 * Changed touchstone settings so that benchmarks are only performed if the stan model is changed. By @sbfnk in #400 and reviewed by @seabbs.
+* `dist_fit()`'s `samples` argument now takes a default value of 1000 instead
```
seabbs, line 19: this needs to be over a single line vs having linebreaks. I corrected this in the previous PRs as only just noticed

seabbs (changes requested):
Nearly there. Just a small formatting issue.

NEWS.md
jamesmbaazam: I've fixed this here 07e137b007f40f632f9e7cbbaa21f7e9c9e4f9a0.

seabbs (approved):
LGTM!

PR #394, opened 2023-05-17
reduce the number of stan models

### LINKED ISSUE #393: reduce number of models to be compiled
Installing the package from source takes quite a long time, partly due to the number of models being compiled. At the very least, the `exp`, `lnorm` and `gamma` models used by `dist_fit` could be combined, and `tune_inv_gamma` (which I don't think is used any more) be removed.

sbfnk:
fixes #393 by combining the models from `dist_fit` into one stan file and ditching the (unused) `tune_inv_gamma` model (which could of course be brought back if needed).


sbfnk:
> I assume you have checked this isn't used?

Yes - I'm not sure it was ever used, at least I couldn't find a use traversing through git blames.

seabbs (commented):
This looks good. I think we need a clean depreciation notice here so ideally that would use lifecycle to report that tune_inv_gamma has been deprecated if called. Even more ideally this could give a link to a gist etc where the code still exists (or the git history) and a note saying to let the authors know if using.


Otherwise this is good to go once it has a news update and a Dev version update. 

Nice to speed up package compilation!

seabbs (approved):
LGTM.

PR #400, opened 2023-05-24
limit touchstone to stan model changes

### LINKED ISSUE #399: Trigger touchstone only when stan model is changed?
We're currently wasting a lot of computation in benchmarking irrelevant changes to the code. We could consider only triggering touchstone when the stan model is changed (additionally perhaps when touchstone itself is changed), given that it will determine the vast majority of run time. This could be implemented by providing a more refined set of paths here:
https://github.com/epiforecasts/EpiNow2/blob/dee8c4f7a438553b44c9b187728e8c9075342305/.github/workflows/touchstone-receive.yaml#L10

sbfnk:
Fixed #399.

We could add a manual trigger but I'm not sure we'd be able to see the results without being embedded in a PR.

### DIFF
--- a/.github/workflows/touchstone-receive.yaml
+++ b/.github/workflows/touchstone-receive.yaml
@@ -7,15 +7,11 @@ concurrency:
 on:
   pull_request:
     paths:
-      # Directories with source code and benchmarking code
-      - "inst/**"
-      - "R/**"
-      - "src/**"
+      # Directories with stan code and benchmarking code
+      - "inst/stan/**"
       - "touchstone/**"
       # Benchmarking config file
       - ".github/workflows/touchstone-*.yaml"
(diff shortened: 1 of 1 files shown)


seabbs:
> We could add a manual trigger but I'm not sure we'd be able to see the results without being embedded in a PR.

Can we add it to the commit message?

sbfnk:
These are both great suggestions but I think would either require a separate workflow file or some fairly involved logic inside the workflow body to check whether the benchmark was triggered by a change in stan code or a trigger commit message. 

As a simpler solution how about we add a `path` trigger for a file called e.g. `.benchmark` which could be temporarily added to a PR to trigger touchstone? We’ll just have to remember to always remove it before merging.

PR #405, opened 2023-05-29
Fix create_clean_reported_cases() error

smkerr:
This pull request aims to address an error I receive when using `estimate_infections()` which appears to be caused by the underlying `create_clean_reported_cases()` function.

Error message:

```
Error in seq.grates_period(min(date), max(date) + horizon, by = "days") : 
  `by` must be an integer of length 1.
```

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -24,7 +24,7 @@ create_clean_reported_cases <- function(reported_cases, horizon,
                                         zero_threshold = Inf) {
   reported_cases <- data.table::setDT(reported_cases)
   reported_cases_grid <- data.table::copy(reported_cases)[,
-   .(date = seq(min(date), max(date) + horizon, by = "days"))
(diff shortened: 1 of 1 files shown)


sbfnk:
Hi, @smkerr, and thanks for your first-time contribution! We very much welcome people fixing bugs - ideally first raising them as an Issue with a reproducible example, see also the [contributing guide](https://github.com/epiforecasts/EpiNow2/blob/0c1aef7056c6592a8bd232bd7762df738ef6b057/.github/CONTRIBUTING.md).

In your case it's not completely clear where the issue comes from but it seems you're using a non-standard date format (`grates_period`?) that can't deal with step sizes of days - perhaps the simplest solution would be to first convert them to be of the base `Date` class?. Your proposed solution would create issues for other standard formats such as `POSIXct`.

I'll close this for now but feel free to re-open if you have a better suggestion, or to provide an issue with a fully reproducible example.

PR #415, opened 2023-07-04
Not for merging: benchmarking generate infections

sbfnk:
Two experiments:

1. Replace [exponential](https://github.com/epiforecasts/EpiNow2/blob/b9bab0810541ca35cfaf39cccee586f28dd6779f/inst/stan/functions/infections.stan#L37) with [multiplications](https://github.com/epiforecasts/EpiNow2/blob/b9bab0810541ca35cfaf39cccee586f28dd6779f/inst/stan/functions/infections_exp.stan#L36) (`exp`)
2. Initialise vectors to zero and replace [sums](https://github.com/epiforecasts/EpiNow2/blob/b9bab0810541ca35cfaf39cccee586f28dd6779f/inst/stan/functions/infections.stan#L46) with [assignments](https://github.com/epiforecasts/EpiNow2/blob/b9bab0810541ca35cfaf39cccee586f28dd6779f/inst/stan/functions/infections_zero.stan#L46) (`zero`) - enabled by [`cd2e738` (#366)](https://github.com/epiforecasts/EpiNow2/pull/366/commits/cd2e738ea283cd7921303d54a22cf5b3f597fc38)

```{r}
> summary %>% filter(name == "infections")
# A tibble: 3 × 7
  model name        mean   low  high lower higher
  <chr> <chr>      <dbl> <dbl> <dbl> <dbl>  <dbl>
1 main  infections  8.18  7.93  8.39  7.86   8.53
2 exp   infections  7.85  7.61  8.10  7.56   8.30
3 zero  infections  6.26  6.11  6.44  6.01   6.50
```

Both appear to lead to speed ups and, I think, would be good changes to implement.

PR #416, opened 2023-07-04
speed up generation of infections

sbfnk:
implements suggestions from #415

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Type: Package
 Package: EpiNow2
 Title: Estimate Real-Time Case Counts and Time-Varying
     Epidemiological Parameters
-Version: 1.3.6.8000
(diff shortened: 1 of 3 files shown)


seabbs (approved):
LGTM. 

Note this may reduce stability in contexts with problematic data which were the motivation for adding these padding values.

sbfnk:
I was wondering about that, thinking that it would have been mitigated by [`cd2e738` (#366)](https://github.com/epiforecasts/EpiNow2/pull/366/commits/cd2e738ea283cd7921303d54a22cf5b3f597fc38) but I might be missing something.

seabbs:
Potentially though it was largely issues in warm-up from the likelihood vs in the generated quantities. I think given pathfinder will soon be out many of these issues with stability are likely soon to be resolved.

Issue #424, opened 2023-07-18
Action warnings


sbfnk:
Some GitHub actions casue deprecation warnings. These should be updated to the latest versions.

PR #425, opened 2023-07-18
update R CMD check action

sbfnk:
- bumps actions/checkout to the latest version (v3) closing #424
- replaces cancel-previous-runs with a concurrency group


seabbs (approved):
LGTM

PR #426, opened 2023-07-18
qualify internal functions in tests

### LINKED ISSUE #422: Tests fail with standard library loading
Some functions in the test suite are not exported. It would be good to qualify them with `EpiNow2:::` so that the tests can be run using just the `testthat` package but without `devtools` or the like.

sbfnk:
Closes #422


seabbs (approved):
LGTM

Issue #429, opened 2023-07-20
Failing samplers during tests


sbfnk:
Tests show sampling failures such as
```
[1] "Error in sampler$call_sampler(args_list[[i]]) : "
[2] "  reached elapsed time limit"                    
[1] "error occurred during calling the sampler; sampling not done"
[1] "Error in sampler$call_sampler(args_list[[i]]) : "
[2] "  reached elapsed time limit"                    
[1] "error occurred during calling the sampler; sampling not done"
[1] "Error in sampler$call_sampler(args_list[[i]]) : "
Execution halted
[2] "  reached elapsed time limit"                    
[1] "error occurred during calling the sampler; sampling not done"
[1] "Error in sampler$call_sampler(args_list[[i]]) : "
[2] "  reached elapsed time limit"                    
[1] "error occurred during calling the sampler; sampling not done"
```

or

```
Warning ('test-estimate_secondary.R:24:1'): (code run outside of `test_that()`)
There were 3402 divergent transitions after warmup. See
https://mc-stan.org/misc/warnings.html#divergent-transitions-after-warmup
to find out why this is a problem and how to eliminate them.
Backtrace:
    ▆
 1. └─EpiNow2::estimate_secondary(...) at test-estimate_secondary.R:24:0
 2.   ├─rstan::sampling(...)
 3.   └─rstan::sampling(...)
 4.     └─rstan (local) .local(object, ...)
 5.       └─rstan:::throw_sampler_warnings(nfit)
```

They occur on all platforms so don't seem to be behind the specific issues on windows (#420) or Ubuntu (#428) but it would probably still be good to avoid theses.

PR #432, opened 2023-07-20
don't defer negative binomial to Poisson at low levels of overdispersion

### LINKED ISSUE #349: estimate_secondary prevalence model does not recover the synthetic parameters
Currentl the example for `estimate_secondary()` does not recover parameters when the prevalence model is used. This could indicate a bug or it could indicate a general identifiability issue.

### LINKED ISSUE #431: Identifiability issues due to approximation to negative binomial
We currently approximate the negative binomial with a poisson for very small levels of overdispersion. This causes a discontinuity in the likelihood surface and renders the overdispersion parameter unidentifiable at such levels. This is behind some of the issues mentioned in #424 and the recovery failures identified in #349.

sbfnk:
Existing comment in the stan code refers to avoiding overflows so will have to monitor that.

This also reinstates the tests for parameter recovery with the secondary model and fixes a bug where observation weights weren't used in the negative binomial model.

Closes #431 
Addresses part of #429 
Fixes #349

### DIFF
--- a/inst/stan/functions/observation_model.stan
+++ b/inst/stan/functions/observation_model.stan
@@ -54,28 +54,22 @@ void truncation_lp(real[] truncation_mean, real[] truncation_sd,
 void report_lp(int[] cases, vector reports,
                real[] rep_phi, real phi_mean, real phi_sd,
                int model_type, real weight) {
-  real sqrt_phi = 1e5;
   if (model_type) {
-    // the reciprocal overdispersion parameter (phi)
+    real sqrt_phi; // the reciprocal overdispersion parameter (phi)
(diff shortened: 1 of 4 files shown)


inst/stan/functions/observation_model.stan
```diff
@@ -85,11 +79,11 @@ vector report_log_lik(int[] cases, vector reports,
   real sqrt_phi = 1e5;
```
seabbs, line 79: Can also drop this.

seabbs (commented):
LGTM. Version and news please otherwise good to go - depending on CI

sbfnk:
Two checks failing for apparently unrelated reasons.

PR #435, opened 2023-07-24
fix logmean bound

sbfnk:
https://github.com/epiforecasts/EpiNow2/commit/d0bea23c4d505f4478786790a6f040e3f30acfb4 introduced a bug where the logmean of delays is lower-bounded by 0 whereas it shouldn't be (for the gamma distribution this is done in the `delays_lp` function).


sbfnk:
No NEWS item needed IMO as no change to previous minor version.

seabbs (approved):
LGTM. Good spot.

PR #436, opened 2023-07-24
fix regional summary in case of 0/0

### LINKED ISSUE #392: Edge case breaking summary stats output
When both the upper and lower bound of the latest estimate are 0, log_cases is set to NA in [summarise.R](https://github.com/epiforecasts/EpiNow2/blob/main/R/summarise.R) (lines 311-312; 0/0 > 1000 returns NA, instead of a logical). 

This produces an error when generating summary statistics: "Error in if (log_cases) {: missing value where TRUE/FALSE needed"

sbfnk:
Fixes #392

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Type: Package
 Package: EpiNow2
 Title: Estimate Real-Time Case Counts and Time-Varying
     Epidemiological Parameters
-Version: 1.3.6.9004
+Version: 1.3.6.9005
 Authors@R: 
     c(person(given = "Sam",
              family = "Abbott",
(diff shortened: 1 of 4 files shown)


sbfnk:
PR will be rebased to `main` once #437 has been merged.

DESCRIPTION
```diff
@@ -2,7 +2,7 @@ Type: Package
 Package: EpiNow2
 Title: Estimate Real-Time Case Counts and Time-Varying
     Epidemiological Parameters
-Version: 1.3.6.9002
+Version: 1.3.6.9003
```
seabbs, line 5: will need to bump in line with other PRs

seabbs (approved):
Nice LGTM.

PR #438, opened 2023-07-25
allow snapshots shorter than trunc_max

sbfnk:
estimate_truncation currently fails when the any snapshot is shorter than `trunc_max`. This PR fixes that by provding explicit bounds and vector elements in the appropriate place.

### DIFF
--- a/inst/stan/estimate_truncation.stan
+++ b/inst/stan/estimate_truncation.stan
@@ -11,6 +11,14 @@ data {
   int trunc_max;
   int trunc_dist;
 }
+transformed data{
+  int<lower = 1> end_t[obs_sets];
+  int<lower = 1> start_t[obs_sets];
+  for (i in 1:obs_sets) {
+    end_t[i] = t - obs_dist[i];
+    start_t[i] = max(1, end_t[i] - trunc_max + 1);
+  }
+}
 parameters {
   real logmean;
   real<lower=0> logsd;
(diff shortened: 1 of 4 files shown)


inst/stan/estimate_truncation.stan
```diff
@@ -45,23 +48,32 @@ model {
...
 generated quantities {
-  matrix[trunc_max, obs_sets] recon_obs;
+  matrix[trunc_max, obs_sets] recon_obs = rep_matrix(0, trunc_max, obs_sets);
   matrix[trunc_max, obs_sets - 1] gen_obs;
   // reconstruct all truncated datasets using posterior of the truncation distribution
   for (i in 1:obs_sets) {
     int end_t = t - obs_dist[i];
-    int start_t = end_t - trunc_max + 1;
-    recon_obs[, i] = truncate(to_vector(obs[start_t:end_t, i]), rev_cmf, 1);
+    int start_t = max(1, end_t - trunc_max + 1);
```
seabbs, line 65: in an ideal world it seems like these would be calculated in R as they are the same across the 3 blocks.

seabbs (approved):
All looks good. I would probably phrase this as a new feature vs fixing a bug as the function was never meant to work when snapshots < delays (though this was of course not communicated). 

Just needs a dev version bump.

inst/stan/estimate_truncation.stan
sbfnk: moved to transformed data as an intermediate solution in https://github.com/epiforecasts/EpiNow2/pull/438/commits/63b6fd50bd915087292159c82f72b0e447016561

seabbs (approved):
LGTM

sbfnk:
> I would probably phrase this as a new feature vs fixing a bug

Done in https://github.com/epiforecasts/EpiNow2/pull/438/commits/4bfc01e943f0b91f2606e8d12bec30b99036f437

seabbs:
Ruh Roh

PR #439, opened 2023-07-25
update estimate truncation vignette

sbfnk:
Adds some content to the `estimate_truncation` vignette and updates the `estimate_infections` vignette to match it.

### DIFF
--- a/vignettes/estimate_truncation.Rmd
+++ b/vignettes/estimate_truncation.Rmd
@@ -17,7 +17,7 @@ knitr::opts_chunk$set(
 ```
 **This is a work in progress. Please consider submitting a PR to improve it.** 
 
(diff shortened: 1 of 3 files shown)


vignettes/estimate_truncation.Rmd
```diff
@@ -16,3 +16,33 @@ knitr::opts_chunk$set(
 )
 ```
 **This is a work in progress. Please consider submitting a PR to improve it.** 
+
+This model deals with the problem of _nowcasting_, or adjusting for right-truncation in reported count data. This occurs when the quantity being observed, for example cases, hospitalisations or deaths, is reported with a delay, resulting in an underestimation of recent counts. The `estimate_truncation()` model attempts to infer parameters of the underlying delay distributions from multiple snapshots of past data. It is designed as a simple and imperfect model. For a more principled approach to nowcasting please consider using the [epinowcast](https://package.epinowcast.org) package.
```
seabbs, line 20: I'm not sure its designed as an imperfect model (i.e. we didn't set out to not do a great job...). Maybe reword - it is designed to be a simple model and therefore may not be ideal for all uses (or similar).

seabbs (approved):
LGTM. Very nice. Only point was about the phrasing "designed to be .. imperfect.." which is not quite true that was more of an emergent property.

vignettes/estimate_truncation.Rmd
sbfnk: Done in https://github.com/epiforecasts/EpiNow2/pull/439/commits/a8314419396f2b25fed9e617bd4e9ef2b2fecad1

seabbs (approved):

PR #443, opened 2023-09-01
Fix 442

### LINKED ISSUE #442: `estimate_secondary` fails with `prior_weight = 0`
Doesn't fail tests because `estimate_secondary` sets `prior_weight = 1` by default

``` r
library(data.table)
library(EpiNow2)

#### Incidence data example ####

# make some example secondary incidence data
cases <- example_confirmed
cases <- as.data.table(cases)[, primary := confirm]
# Assume that only 40 percent of cases are reported
cases[, scaling := 0.4]
# Parameters of the assumed log normal delay distribution
cases[, meanlog := 1.8][, sdlog := 0.5]

# Simulate secondary cases
cases <- simulate_secondary(cases, type = "incidence")
cases[
  ,
  c("confirm", "scaling", "meanlog", "sdlog", "index", "scaled", "conv") :=
    NULL
]
#
# fit model to example data specifying a weak prior for fraction reported
# with a secondary case
delays <- dist_spec(mean = 2.5, mean_sd = 0.5, sd = 0.47, sd_sd = 0.25, max = 30)

inc <- estimate_secondary(
  cases[1:60], delays = delays,
  obs = obs_opts(scale = list(mean = 0.2, sd = 0.2), week_effect = FALSE),
  verbose = FALSE
)
#> Error in create_stan_delays(delay = delays, trunc = truncation): argument "ot" is missing, with no default
```

<sup>Created on 2023-09-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

sbfnk:
fixes #442 

also fixes a related typo in the documentation and adds a new test that would have caught the issue

### DIFF
--- a/R/dist.R
+++ b/R/dist.R
@@ -890,7 +890,7 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
 #' assigned to \code{mean_sd} and \code{sd_sd} by setting them to zero.
 #' reduces compute requirement but may produce spuriously precise estimates.
 #' @param prior_weight Integer, weight given to the generation time prior.
-#' By default (prior_weight = 00) the priors will be weighted by the number of
+#' By default (prior_weight = 0) the priors will be weighted by the number of
 #' observation data points, usually preventing the posteriors from shifting
 #' much from the given distribution. Another sensible option would be 1,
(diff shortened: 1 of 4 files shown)


seabbs (approved):
LGTM

PR #445, opened 2023-09-01
move synthetic validation to macos

### LINKED ISSUE #444: Synthetic validation jobs fail
For no obvious reason as far as I can see. Seems to occur always at the same step (building vdiffr)

sbfnk:
testing whether this fixes #444


seabbs (approved):
Looks like that did the job.

PR #446, opened 2023-09-01
bug fix: make simulated delay mean unbounded

sbfnk:
This can be the mu parameter of the lognormal which should be unbounded


sbfnk:
merging as hot fix

PR #450, opened 2023-09-08
move prior weights to model arguments

### LINKED ISSUE #447: Default values of `prior_weight`
The slighly broader issue is is that we would probably want a default of 0 for `estimate_infections` (weigh by number of data points) and 1 for `estimate_secondary` (don't weigh) but if we specify the prior weights with the distributions we don't know at the time of declaration which model this will be fed into and as what kind of distribution (gt/delay/etc.). Perhaps this should be a model option.

_Originally posted by @sbfnk in https://github.com/epiforecasts/EpiNow2/issues/442#issuecomment-1705417844_

sbfnk:
Closes #447

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Type: Package
 Package: EpiNow2
 Title: Estimate Real-Time Case Counts and Time-Varying
     Epidemiological Parameters
-Version: 1.3.6.9007
+Version: 1.3.6.9008
 Authors@R: 
     c(person(given = "Sam",
              family = "Abbott",
--- a/R/create.R
+++ b/R/create.R
@@ -640,12 +640,11 @@ create_stan_args <- function(stan = stan_opts(),
 ##'
 ##' @param ... Named delay distributions specified using `dist_spec()`.
 ##' The names are assigned to IDs
-##' @param ot Integer, number of observations (needed if weighing any priors)
-##' with the number of observations
+##' @param weight Numeric, weight associated with delay priors; default: 1
 ##' @return A list of variables as expected by the stan model
 ##' @importFrom purrr transpose map
 ##' @author Sebastian Funk
-create_stan_delays <- function(..., ot) {
+create_stan_delays <- function(..., weight = 1) {
   dot_args <- list(...)
   ## combine delays
   combined_delays <- unclass(c(...))
@@ -673,9 +672,7 @@ create_stan_delays <- function(..., ot) {
   ## map pmfs
   ret$np_pmf_groups <- array(c(0, cumsum(combined_delays$np_pmf_length)) + 1)
   ## assign prior weights
-  if (any(ret$weight == 0)) {
-    ret$weight[ret$weight == 0] <- ot
-  }
+  ret$weight <- array(rep(weight, ret$n_p))
   ## remove auxiliary variables
   ret$fixed <- NULL
   ret$np_pmf_length <- NULL
--- a/R/estimate_infections.R
+++ b/R/estimate_infections.R
@@ -35,8 +35,9 @@
 #' @param horizon Numeric, defaults to 7. Number of days into the future to
 #' forecast.
 #'
-#' @param weigh_prior_delays Logical. If TRUE (default), all delay distribution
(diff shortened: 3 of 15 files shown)


sbfnk:
Not adding a deprecation warning as `dist_skel` was only introduced in this same version.

DESCRIPTION
```diff
@@ -2,7 +2,7 @@ Type: Package
 Package: EpiNow2
 Title: Estimate Real-Time Case Counts and Time-Varying
     Epidemiological Parameters
-Version: 1.3.6.9007
+Version: 1.3.6.9008
```
seabbs, line 5: I have decided this is OTT if we want to stop doing this?

R/estimate_infections.R
```diff
@@ -245,6 +251,7 @@ estimate_infections <- function(reported_cases,
                                 CrIs = c(0.2, 0.5, 0.9),
                                 filter_leading_zeros = TRUE,
                                 zero_threshold = Inf,
+                                weigh_prior_delays = TRUE,
```
seabbs, line 254: Is this clear enough? As an alternative could cal it `weight_delay_prior_by_time` or similar? I think this doesn't matter hugely so happy as is.

R/estimate_infections.R
```diff
@@ -35,6 +35,11 @@
 #' @param horizon Numeric, defaults to 7. Number of days into the future to
 #' forecast.
 #'
+#' @param weigh_prior_delays Logical. If TRUE (default), all delay distribution
+#' priors will be weighted by the number of observation data points, usually
+#' preventing the posteriors from shifting. If FALSE, no weight will be applied,
```
seabbs, line 40: Could add: This approximates placing an independent prior on each time point but with less computational overhead.

R/estimate_secondary.R
```diff
@@ -40,6 +40,11 @@
 #' use for estimation but not to fit to at the beginning of the time series.
 #' This must be less than the number of observations.
 #'
+#' @param weigh_prior_delays Logical. If TRUE, all delay distribution
```
seabbs, line 43: Rather than duplicating we likely want to inherit from one of the functions.

seabbs (commented):
This looks good. A few small comments about naming and things. 

Am I missing a stan code change for `ot` to `weight` or am I confused?

DESCRIPTION
sbfnk: Sounds good to me. Whilst we're there shouldn't we really follow version 1.3.5 by dev version 1.3.5.9000 so that we don't pre-determine whether the next release will be major, minor or patch? That's what `usethis::use_dev_version()` does.

R/estimate_secondary.R
sbfnk: We can only do that if we don't want to mention the default in the man page. I felt this trumped the non-repetition. That said, one of `estimate_truncation` and `estimate_secondary` could still inherit from each other.

R/create.R
```diff
@@ -673,9 +672,7 @@ create_stan_delays <- function(..., ot) {
   ## map pmfs
   ret$np_pmf_groups <- array(c(0, cumsum(combined_delays$np_pmf_length)) + 1)
   ## assign prior weights
-  if (any(ret$weight == 0)) {
-    ret$weight[ret$weight == 0] <- ot
-  }
+  ret$weight <- array(rep(weight, ret$n_p))
```
sbfnk, line 675: @seabbs on your question of variable name change: in stan it's still a per-delay weight (called `weight`) which is set here. It's only that we don't pass `ot` to this but rather an explicit weight (as it's only used for that) which is effectively set to `ot` in the `estimate_` functions.

DESCRIPTION
seabbs: yes agree

R/create.R
seabbs: Got it thanks

R/estimate_infections.R
sbfnk: As a compromise I renamed it to `weigh_delay_priors` which I think is better.

seabbs (approved):
LGTM. Happy once passing checks (which it seems to be about to be).

PR #452, opened 2023-09-08
make estimate_truncation return a dist_spec

### LINKED ISSUE #449: `estimate_truncation` should return a call to `dist_spec`
This might make more sense than returning a list

https://github.com/epiforecasts/EpiNow2/blob/d9f128dc756f577fb14caf395b1ff9106c5bc996/R/estimate_truncation.R#L251

though potentially exacerbates #447 which should therefore be fixed first.

sbfnk:
Fixes #449


seabbs (approved):
LGTM - nice.

PR #453, opened 2023-09-08
Update deprecated syntax for future rstan compatibility

andrjohns:
Now that rstan 2.26 is available on CRAN we need to update the deprecated syntax in your package's Stan models, otherwise it will fail to install with an upcoming version of RStan. 

The following updates have been made:

- New array syntax
- `_cdf` functions using conditional `|` notation (i.e., `_cdf(y |x)`)
    - The current CRAN rstan has a bug in recognising this in _`cdf` functions, so I've updated your syntax to work on the log scale with the respective `_lcdf` function. This will let your package be compatible with the current and future parser, and is also more numerically stable with extreme values (bonus!)

More information about the deprecated and removed syntax in Stan can be found here:
- https://mc-stan.org/docs/functions-reference/deprecated-functions.html
- https://mc-stan.org/docs/functions-reference/removed-functions.html

If you could merge and submit to CRAN soon, it would be greatly appreciated.

Let me know if you have any questions about these changes.

Thanks!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -109,7 +109,7 @@ Imports:
     R.utils (>= 2.0.0),
     Rcpp (>= 0.12.0), 
     rlang (>= 0.4.7),
-    rstan (>= 2.21.1),
+    rstan (>= 2.26.0),
     rstantools (>= 2.2.0),
     runner,
     scales,
@@ -135,8 +135,8 @@ LinkingTo:
     Rcpp (>= 0.12.0),
     RcppEigen (>= 0.3.3.3.0),
     RcppParallel (>= 5.0.1),
-    rstan (>= 2.21.1),
-    StanHeaders (>= 2.21.0-5)
+    rstan (>= 2.26.0),
+    StanHeaders (>= 2.26.0)
 Biarch: true
 Config/testthat/edition: 3
 Encoding: UTF-8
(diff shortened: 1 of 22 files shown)


sbfnk:
Thanks! Exciting to see the `rstan` update on CRAN. I had a look at the failing checks and as far as I can see it seem to be due to this [issue](https://github.com/stan-dev/rstan/issues/1093) I just opened on rstan.

andrjohns:
> Thanks! Exciting to see the `rstan` update on CRAN. I had a look at the failing checks and as far as I can see it seem to be due to this [issue](https://github.com/stan-dev/rstan/issues/1093) I just opened on rstan.

Thanks for flagging! I've opened a [PR with a fix](https://github.com/stan-dev/rstan/pull/1094) and I'll let you know when it's made it to CRAN

seabbs:
As we aren't testing stan code on CRAN (just on our CI) and that is what we use `expose_stan_functions()` for we should be able to update on CRAN/merge this without waiting on that PR fix. We would need to temporarily pause CI checks of the stan code in order to do this. 

That does leave on outstanding issue which is causing `R-CMD-as-CRAN-check` to fail which is the `estimate_truncation` example - will check this out today to see what the issue is.

sbfnk:
@andrjohns can you rebase to main / merge to see if the as-cran-check is fixed by #454?

sbfnk:
CRAN checks passing so merging for CRAN submission as discussed

PR #455, opened 2023-09-18
Improve deprecation

sbfnk:
Adds soft deprecation for the old distribution interface so as to avoid nasty surprises for unsuspecting CRAN users.

This is in anticipation of a 1.4.0 CRAN release.


sbfnk:
merging as CRAN checks are passing and others expected to fail

PR #457, opened 2023-09-27
fix gt pmf dimensions for non-parametric model

sbfnk:
The gt pmf gets calculated in the generated quantities in that case.


seabbs (approved):
Looks good. I have always done this with an if else in the past but not sure why as this way is shorter and just as clear.

PR #458, opened 2023-09-27
workflow/options vignettes

sbfnk:
Adds two vignettes, one on the general workflow for working with the package (which could be potentially expanded with more examples showing how the different models can feed into each other) and one to illustrate the various options for running `estimate_infections` in preparation from moving them out of the examples to address #430.

Vignettes are precompiled so as to not require re-rendering all the time in GitHub actions.

PR #459, opened 2023-09-27
Remove examples with long runtimes

### LINKED ISSUE #430: Long example run times
Some of the examples have very long run times - especially the ones for `estimate_infections` which run a number of model variants that take about an hour on github actions. It might be good to keep this to a single example and move the variants to a vignette explaining/showing differences and possible advantages/disadvantages.

sbfnk:
Partly by storing some model results as data. See https://github.com/epiforecasts/EpiNow2/pull/458#issue-1915664824

Closes #430.


seabbs (commented):
Nice. Do you think CRAN are going to be okay with this? In the past they have not been which is why `epinowcast` just `extdata` and loads in the example from a function.

sbfnk:
Good question - also the increased size of `data/` might raise eyebrows. 

This also reminded me that we have example outputs in `inst` in `rbi` which didn't cause issues with CRAN. Will move it there.

seabbs (dismissed):
LGTM

sbfnk:
Source tarball is now 6.6 MB which should be fine for CRAN.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # EpiNow2 1.4.9000
 
+## Package
+
+* Reduced the number of long-running examples.
```
seabbs, line 5: Missing the standard PR, author and reviewer attribution we have used elsewhere.

PR #464, opened 2023-10-03
Improve contribution guide

jamesmbaazam:
This PR improves the contribution guide by adding information on pre-compiled vignettes and how to request features and report bugs.


.github/CONTRIBUTING.md
```diff
@@ -10,9 +10,21 @@ You can find the `.R` file that generates the `.Rd` by reading the comment in th
 
 ## Bigger changes
 
-If you want to make a bigger change, it's a good idea to first file an issue and make sure someone from the team agrees that it’s needed. 
-If you’ve found a bug, please file an issue that illustrates the bug with a minimal 
-[reprex](https://www.tidyverse.org/help/#reprex) (this will also help you write a unit test, if needed).
+If you want to make a bigger change, it's a good idea to first file an issue and make sure 
+someone from the team agrees that it’s needed. 
+
+## New features & Bugs
```
seabbs, line 16: maybe this should be a ### subtitle?

.github/CONTRIBUTING.md
```diff
@@ -10,9 +10,21 @@ You can find the `.R` file that generates the `.Rd` by reading the comment in th
...
+
+If you have an idea for a feature or have found a bug, please file an issue.
```
seabbs, line 18: maybe we should split these into their own sections. Could potentially better link the text from bigger changes

.github/CONTRIBUTING.md
```diff
@@ -10,9 +10,21 @@ You can find the `.R` file that generates the `.Rd` by reading the comment in th
...
+For bugs, illustrate them with a minimal [reprex](https://www.tidyverse.org/help/#reprex)
+(this will also help you write a unit test, if needed). 
+
+## Vignettes
+
+If you have an idea for a vignette, please file an issue with an outline of the vignette
```
seabbs, line 24: What about if you what to change a current vignette?

seabbs (changes requested):
Looks good. Needs a news item before can merge. Also had a very comments on the contributing guide content.

.github/CONTRIBUTING.md
jamesmbaazam: True. Will fix that in the incoming commits.

NEWS.md
```diff
@@ -1,9 +1,10 @@
-# EpiNow2 1.4.9000
+# EpiNow2 1.4.9001
```
seabbs, line 1: we have stopped doing this (and if this isn't flagged in the contributing guide (or rather not flagged) it should/shouldn't be.

jamesmbaazam: Oh, really? Not sure I understand. Do you mean incrementing by 1 or incrementing in general? What should/shouldn't be flagged?

sbfnk: Broadly following the [Tidyverse package versioning](https://r-pkgs.org/lifecycle.html) (which perhaps isn't mentioned in the contribution guide)

> Increment the development version, e.g. from 9000 to 9001, if you’ve added an important feature and you (or others) need to be able to detect or require the presence of this feature. For example, this can happen when two packages are developing in tandem. This is generally the only reason that we bother to increment the development version. This makes in-development versions special and, in some sense, degenerate. Since we don’t increment the development component with each Git commit, the same package version number is associated with many different states of the package source, in between releases.

sbfnk: I think the flagging was referring to the whether this is mentioned in the contribution guide or not

jamesmbaazam: Thanks, I'll remove the increment. Are we going to determine on a case-by-case basis when to do the increment? Based on the snippet above, it seems something like changes to a vignette or the addition of one shouldn't warrant an increment. The guide already mentions that we ask for a NEWS item to be added.

sbfnk: Yes I think case-by-case with a strong default to no and only in exceptions if there is a specific need (which I currently can't imagine what it would be).

jamesmbaazam: I've made the change here 517968b5fef7805826cd39f6422e63d1733a9b34.

seabbs (approved):
LGTM - thanks @jamesmbaazam!

PR #468, opened 2023-10-04
Clearer distinguish length and max

### LINKED ISSUE #410: Distribution max is really the length
The `max` argument of probability distributions in `dist_spec` is not really the maximum, but the length of the PMF. Since its first index corresponds to 0, the maximum is actually `max - 1`.

There could be two solutions:
1. changing the interpretation of `max` to correctly reflect the maximum
2. changing the name of the argument to `length`

Since (1) is probably more intuitive to the user, I think it's the preferred option but would require more code changes.

sbfnk:
Fixes #410

Also updates examples and tests to have their max value reduced by 1 (as the previous interpretation of max parameters was to be the length).

This required a slight change to what `dist_skel` to return NA if asking for the probability mass of a value that is greater than the max.


sbfnk:
LGTM (self-review).

@seabbs would you like to review or happy for it to be merged as is?

seabbs (approved):
All looks good to me.

PR #469, opened 2023-10-04
testing with updated rstan

sbfnk:
can't use this for any CRAN version but could be an intermediate solution for getting all the tests/checks up and running again


sbfnk:
Can't do as `pak` doesn't seem to support submodules.

PR #472, opened 2023-10-12
cmdstanr version

sbfnk:
This currently replaces the use of `rstan` in the package with `cmdstan`. Eventually it should be turned into an option to have either.


sbfnk:
Closing in favour of #537

PR #474, opened 2023-10-12
fix bounds of initial conditions for delays

sbfnk:
as the mean currently incorrectly has a lower bound of 0 in all cases whereas it should only have it for gamma (for lognormal this parameter is interpreted as mu which is unbounded).


sbfnk:
LGTM (self-review for minor change). Will merge once tests have succeeded following rebase.j

PR #467, opened 2023-10-03
Keep CITATION and `citation.cff` in sync with DESCRIPTION

jamesmbaazam:
This PR closes #466 by adding a GitHub Actions workflow to auto-generate `citation.cff` whenever the citation fields in CITATION and DESCRIPTION change. It also adds code to keep `inst/CITATION` and `DESCRIPTION` in sync. Together, this will ensure that the package citation information is always up to date.

PR #476, opened 2023-10-12
Add input validation

### LINKED ISSUE #337: Add input validation
I think many errors (both in my own usage and in some bug reports such as #267) seem to come from invalid input. 

I understand that what constitutes a valid input is documented but I believe some extra hand-holding by failing early with clear error messages on invalid input would go a long way in helping users.

A main topic to discuss is probably what would be the best framework to do this input checking. Some options are presented in [the 'Checking the inputs of your R functions' blog post](https://blog.r-hub.io/2022/03/10/input-checking/).

jamesmbaazam:
This PR:

* Adds input checking to 5 core functions: `estimate_infections()`, `estimate_secondary()`, `estimate_truncation()`, `simulate_infections()`, and `epinow()`.

* A function `check_reports_valid()` is added to validate the `reports` dataset passed to the functions. Tests are added to check its functionality.

* The various `*_opts()` functions are assigned subclasses of the same name as the functions. These are tested against the assigned arguments to ensure the right `*_opts()` is passed to the right argument. For example, the `obs` argument in `estimate_secondary()` is expected to only receive arguments passed through `obs_opts()` and will error otherwise. 

This PR closes #337.

### DIFF
--- a/R/opts.R
+++ b/R/opts.R
@@ -533,8 +533,7 @@ obs_opts <- function(family = "negbin",
 #'
 #' @param ... Additional parameters to pass to `rstan::sampling`.
 #' @importFrom utils modifyList
-#' @return An `<rstan_sampling_opts>` object of arguments to pass to
-#' `rstan::sampling`.
+#' @return A list of arguments to pass to `rstan::sampling`.
 #' @author Sam Abbott
 #' @export
 #' @examples


jamesmbaazam:
@seabbs @sbfnk I've implemented a check for the input data to the various `estimate_*()` functions. It expects the non-date columns to be integer. Tests are currently failing locally because `example_data.rds` has a "numeric" confirm column. I would prefer to make that conversion at the source but can't find where is the data created. Does it exist?

sbfnk:
I'm not sure we want this to be strictly integer as the method should work in principle with e.g. proportions.

The check to me looks good, only think missing I think is a corresponding test.

In the rest of the package we have mostly (although I don't think strictly) taken the approach of importing functions with `@importFrom` and then not qualifying them with e.g. `checkmate::` in the code except when it might be ambiuous.

jamesmbaazam:
> The check to me looks good, only think missing I think is a corresponding test.

Thanks. It's still a work in progress, hence the PR is a draft.

sbfnk:
Noting that the test coverage PR was cancelled after 6 hours (didn't fail) - might be worth re-running to see if it's a transient GHA issue.

R/checks.R
```diff
@@ -0,0 +1,46 @@
...
+#' * three columns: date, primary, and secondary, if to be used by
+#'  estimate_secondary().
+#' @param for_estimate_secondary Logical; whether the data is being passed by
+#' estimate_secondary() or the other estimate_*() functions.
+#' @importFrom checkmate assert_data_frame assert_date assert_names
+#' assert_numeric
+#' @return Called for its side effects.
+#' @author James M. Azam
+#' @keywords internal
+check_reports_valid <- function(reports, for_estimate_secondary) {
```
sbfnk, line 21: Following the argument at https://design.tidyverse.org/boolean-strategies.html (as pointed out by @seabbs) I think it might be better to have this as an enum, something along the lines of `for = c("estimate_infections", "estimate_secondary")`

R/estimate_infections.R
```diff
@@ -131,6 +133,24 @@ estimate_infections <- function(reported_cases,
                                 weigh_delay_priors = TRUE,
                                 id = "estimate_infections",
                                 verbose = interactive()) {
+  # Validate inputs
+  check_reports_valid(reported_cases, for_estimate_secondary = FALSE)
+  assert_class(generation_time, "generation_time_opts", null.ok = TRUE)
```
sbfnk, line 138: just flagging that I think this should be `null.ok = FALSE` and the default argument for `generation_time` should be `generation_time_opts()` as in `estimate_infections()`, but that is an issue outside this PR.

R/estimate_infections.R
```diff
@@ -131,6 +133,24 @@ estimate_infections <- function(reported_cases,
...
+  assert_class(delays, "delay_opts", null.ok = TRUE)
+  assert_class(truncation, "trunc_opts", null.ok = TRUE)
+  assert_class(rt, "rt_opts", null.ok = TRUE)
+  assert_class(backcalc, "backcalc_opts", null.ok = TRUE)
+  assert_class(gp, "gp_opts", null.ok = TRUE)
+  assert_class(obs, "obs_opts", null.ok = TRUE)
+  assert_class(stan, "stan_opts", null.ok = TRUE)
```
sbfnk, line 145: I might be wrong (and it would be easy to test) but I don't think any of these except `rt` should have `null.ok = TRUE` - where setting `rt = NULL` perhaps somewhat confusingly indicates to not use a renewal model (which will eventually be addressed by #345).

R/estimate_secondary.R
```diff
@@ -148,6 +150,19 @@ estimate_secondary <- function(reports,
                                weigh_delay_priors = FALSE,
                                verbose = interactive(),
                                ...) {
+  # Validate the inputs
+  check_reports_valid(reports, for_estimate_secondary = TRUE)
+  assert_class(secondary, "secondary_opts", null.ok = TRUE)
+  assert_class(delays, "delay_opts", null.ok = TRUE)
+  assert_class(truncation, "trunc_opts", null.ok = TRUE)
+  assert_class(obs, "obs_opts", null.ok = TRUE)
```
sbfnk, line 158: as above re `null.ok = TRUE` I think

R/estimate_secondary.R
```diff
@@ -275,6 +290,7 @@ secondary_opts <- function(type = "incidence", ...) {
     )
   }
   data <- update_list(data, list(...))
+  attr(data, "class") <- c("secondary_opts", class(data))
```
sbfnk, line 293: return tag in roxygen header needs to be updated.

Beyond this PR but this should probably live in `opts.R` for easier findability.

R/checks.R
jamesmbaazam: I've made this change here b4c6ecb46882fdc54b40a0be7a0ea2feb32b0636.

R/estimate_infections.R
jamesmbaazam: Yes, you're right. `rt` and `gp` can be null but the others can't be.

R/estimate_secondary.R
jamesmbaazam: Thanks.

R/estimate_infections.R
jamesmbaazam: I've fixed the `null.ok = FALSE` issue. For the `generation_time` default, I'm guessing you meant to comment under `regional_epinow()`?

R/epinow.R
```diff
@@ -105,6 +108,17 @@ epinow <- function(reported_cases,
                    plot_args = list(),
                    target_folder = NULL, target_date,
                    logs = tempdir(), id = "epinow", verbose = interactive()) {
+  # Check inputs
+  assert_logical(return_output)
+  stopifnot("target_folder is not a directory" =
```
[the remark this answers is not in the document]
jamesmbaazam: I'm not sure I understand. This only checks that the specified target folder is a valid directory string.

R/estimate_secondary.R
jamesmbaazam: I've updated the docs to mention the returned class.

R/epinow.R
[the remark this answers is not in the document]
sbfnk: Ah my bad, sorry.

seabbs:
> Please add yourself as an author to any function where you've made substantial additions.


If this is `EpiNow2` dev policy we should update the contributing guide.

jamesmbaazam:
I'm not sure this test-coverage failure is related to the PR content.

jamesmbaazam:
> > Please add yourself as an author to any function where you've made substantial additions.
> If this is `EpiNow2` dev policy we should update the contributing guide.

Sure. We can have a section on giving credit. We recently did similar for Epiverse [here](https://epiverse-trace.github.io/blueprints/contribution-acknowledgements.html). I'm actually not sure what constitutes a substantial addition. I guess I would define it as a change without which the function would not do as promised? I would imagine that I would qualify as a co-author of these functions when I work on the S3 class structure for each of them. In that case, the return value of the function is the class with certain attributes, which I implemented, so I can consider myself a co-author.

R/estimate_secondary.R
```diff
@@ -237,7 +252,8 @@ estimate_secondary <- function(reports,
 #' options that can be passed.
 #'
 #' @seealso estimate_secondary
-#' @return A list of binary options summarising secondary model used in
+#' @return A `<secondary_opts>` object of binary options summarising secondary
+#' model used in
```
seabbs, line 256: line ending in a weird place

R/opts.R
```diff
@@ -538,6 +548,7 @@ rstan_sampling_opts <- function(cores = getOption("mc.cores", 1L),
                                 future = FALSE,
                                 max_execution_time = Inf,
                                 ...) {
+  dot_args <- list(...)
```
seabbs, line 551: better code style?

seabbs (changes requested):
This all looks really nice. Looking forward to having this merged. Just a few comments about line breaks (exciting times) and we need a news update for this.

sbfnk:
> I'm not sure this test-coverage failure is related to the PR content.

I think it's related but I'm not sure how - the corresponding action hasn't been failing on other PRs, e.g.
https://github.com/epiforecasts/EpiNow2/pull/467

That said I could not reproduce the timeouts running `covr::codecov` locally and they're not timing out within the R CMD check so I'm not sure what's going on.

R/checks.R
```diff
@@ -0,0 +1,55 @@
...
+#' `check_reports_valid()` checks that the supplied data is a `<data.frame>`,
+#' and that it has the right column names and types. In particular, it checks
+#' that the date column is in date format and does not contain NA's, and that
+#' the other columns are numeric.
+#'
+#' @param reports A data frame with either:
+#' * a minimum of two columns: `date` and `confirm`, if to be
+#'  used by [estimate_infections()] or [estimate_truncation()], or
+#' * a minimum of three columns: `date`, `primary`, and `secondary`, if
+#' to be used by
```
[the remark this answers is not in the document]
jamesmbaazam: Removed.

R/checks.R
```diff
@@ -0,0 +1,55 @@
...
+#' @importFrom checkmate assert_data_frame assert_date assert_names
+#' assert_numeric
+#' @return Called for its side effects.
+#' @author James M. Azam
+#' @keywords internal
+check_reports_valid <- function(reports, model) {
+  # Check that the case time series (reports) is a data frame
+  assert_data_frame(reports)
+  # Perform checks depending on the model to the data is meant to be used with
+  model <- match.arg(
```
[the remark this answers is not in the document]
jamesmbaazam: Good point. I'll raise an issue. I suggested elsewhere that we could also add it to the `testthat` setup like we've done [here](https://github.com/epiverse-trace/packagetemplate/blob/main/tests/testthat/setup-options.R).

R/estimate_secondary.R
jamesmbaazam: I've removed it.

R/opts.R
jamesmbaazam: I catch it here for modification later in the function because `iter` has two values in the final list if it is passed through `...` and also recalculated below.

jamesmbaazam:
> That said I could not reproduce the timeouts running `covr::codecov` locally and tests are not timing out within the R CMD check so I'm not sure what's going on.

Same. It passes locally but with a warning and 3 notes.

```r
❯ checking top-level files ... WARNING
  A complete check needs the 'checkbashisms' script.
  See section ‘Configure and cleanup’ in the ‘Writing R Extensions’
  manual.

❯ checking installed package size ... NOTE
    installed size is 12.3Mb
    sub-directories of 1Mb or more:
      doc       1.1Mb
      extdata   2.1Mb
      help      2.1Mb
      libs      6.1Mb

❯ checking for GNU extensions in Makefiles ... NOTE

❯ checking files in ‘vignettes’ ... NOTE
  The following directory looks like a leftover from 'knitr':
    ‘figure’
  Please remove from your package.

0 errors ✔ | 1 warning ✖ | 3 notes ✖
```

sbfnk:
Could you try temporarily removing test-checks.r and see if that fixes it?

sbfnk:
> Could you try temporarily removing test-checks.r and see if that fixes it?

Turns out this didn't do the trick. Next steps could be:
- roll back more changes (can always reset to 157b5c4 and force push to revert to original state later) if that's even possible (and I realise it might not be)
- address #492 (complete stab in the dark but who knows)
- ignore (but then we lose the codecov action which would be a bit of a shame)

jamesmbaazam:
I'll create a backup branch and try the fixes there. Whichever works will then be applied here.

seabbs:
Could this be a problem with the cache for this action? I just had this issue in `epinowcast` (but for another action) and resolved it by removing the related cache using `gh cache`

jamesmbaazam:
I've removed them and re-run the action. Let's see if it fixes it. 🤞🏾

sbfnk:
Maybe temporarily changing the action here to run `covr::codecov` with `quiet = FALSE` will help identify the issue?

jamesmbaazam:
> Maybe temporarily changing the action here to run `covr::codecov` with `quiet = FALSE` will help identify the issue?

You mean with `quiet =. TRUE`? I'm trying what was suggested [here](https://github.com/r-lib/covr/issues/435#issuecomment-638793132) in d1fa59d5b7220546bd589abc7188f96848d1a5cb.

sbfnk:
I thought `quiet = TRUE` was the [default](https://covr.r-lib.org/reference/codecov.html) so `quiet = FALSE` would be to get a few more messages that maybe would help pinpointing when it hangs.

jamesmbaazam:
> I thought `quiet = TRUE` was the [default](https://covr.r-lib.org/reference/codecov.html) so `quiet = FALSE` would be to get a few more messages that maybe would help pinpointing when it hangs.

It was set to FALSE already in the workflow.

jamesmbaazam:
This issue has persisted after several attempts so I'm going to split this PR into smaller parts and see if that helps:
- One to add the function that checks the reports argument.
- Separate PR's with input checks for each functions.

sbfnk:
I've been able to reproduce the issue using https://github.com/nektos/act (which is a great little tool for investigating GHA issues locally). 

Just looking at some runs I started before the weekend it looks like all the tests in `test-estimate_infections.R` were slowed down to snail's pace, apparently because of an issue in the interaction between `codecov` and `stan_opts` objects being passed to `do.call()` calls. Changing the code so that only pure lists are passed to `do.call` has resolved the issue. This is implemented in https://github.com/epiforecasts/EpiNow2/commit/db0bfce6c70fe0e6b1430ec0d056755a9c7273e2 which is based off the last commit of the original PR before we started investigating this issue.

To apply this fix in this PR you can:
1. `git reset --hard` to [5410759](https://github.com/epiforecasts/EpiNow2/pull/476/commits/5410759aa627aa0583ec3db523f3a0a4aef51392)
2. `git merge` the remote `fix-codecov` branch
3. force push

jamesmbaazam:
Amazing! Thanks for taking a look @sbfnk.  The `stan_opts` change is my fault. I made that change in ba00fc12a98250b9a6d485198f215ee47fcacd0f.

sbfnk:
It's still a really strange issue that didn't occur when either of us ran `codecov::covr`. Fingers crossed this fixes it.🤞

jamesmbaazam:
🥳

jamesmbaazam:
This should be ready for merge now.

R/estimate_infections.R
```diff
@@ -455,26 +455,15 @@ fit_model_with_nuts <- function(args, future = FALSE, max_execution_time = Inf,
...
     args$cores <- 1
     fits <- future.apply::future_lapply(1:chains, fit_chain,
-      stan_args = args,
-      max_time = max_execution_time,
-      catch = TRUE,
-      future.seed = TRUE
+                                        stan_args = args,
+                                        max_time = max_execution_time,
+                                        catch = TRUE,
+                                        future.seed = TRUE
```
sbfnk, line 466: This is deviating from our standard indenting style in the package (two spaces).

jamesmbaazam: This has been resolved in 841d34068d7b23f2979874fe0dd201da4e5c0792.

sbfnk (approved):

sbfnk (approved):

PR #484, opened 2023-10-24
stan model chunks

### LINKED ISSUE #381: Put duplicated stan code in chunks
Some code is duplicated between the various models in `inst/stan`. It might be good to move these into chunks that are pulled in. See https://github.com/epiforecasts/EpiNow2/pull/363#discussion_r1112398687

sbfnk:
Testing whether we could move duplicated stan code into chunks.

I like the way it simplifies the main models and highlights commonalities. It also makes coding more fault tolerant.

On the other hand it will become even harder to trace code having to navigate, at times, to chunks that then call functions contained in other files. This could be mitigated by, e.g., making functions that are only called once chunks (e.g. the GP functions); or by moving each stan function into a separate files (though would complicate ensuring all relevant functions are loaded for all chunks) or all into one file.

In conclusion, I'm on the fence regarding whether this is a good idea. If going ahead with this there are a few bits of code that could potentially still be moved into chunks.

A few processing functions had to be updated in order to make this work. Also the stan code is probably less efficient in places - touchstone will tell us how much.

Closes #381

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -654,8 +654,7 @@ create_stan_delays <- function(..., weight = 1) {
   ## number of different non-empty types
   type_n <- unlist(purrr::transpose(dot_args)$n)
   ## assign ID values to each type
-  ids <- rep(0L, length(type_n))
-  ids[type_n > 0] <- seq_len(sum(type_n > 0))
+  ids <- seq_along(type_n)
   names(ids) <- paste(names(type_n), "id", sep = "_")
 
   ## start consructing stan object
--- a/R/dist.R
+++ b/R/dist.R
@@ -978,13 +978,7 @@ dist_spec <- function(mean, sd = 0, mean_sd = 0, sd_sd = 0,
     )
     if (length(pmf) == 0) {
       if (missing(mean)) { ## empty
-        ret <- c(ret, list(
-          n = 0,
-          n_p = 0,
-          n_np = 0,
-          np_pmf = numeric(0),
-          fixed = integer(0)
-        ))
+        pmf <- 1
       } else { ## parametric fixed
         if (sd == 0) { ## delta
           pmf <- c(rep(0, mean), 1)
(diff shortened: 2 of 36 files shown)


jamesmbaazam:
On the issue of navigating the code, can we use prefixes to differentiate between the stan code in `./inst/stan/functions/` directory and those in the `./inst/stan/data/` directory? We could prefix the functions with `fnc_*` and those in data with "data_". I find that some stan files have the same name in both folders and can be confusing when opened side by side.

sbfnk:
> the stan code is probably less efficient in places

That turned out to be of no concern.

seabbs:
>  I find that some stan files have the same name in both folders and can be confusing when opened side by side.

So this is intentional in order to show which data chunks link with which function chunks etc? In the viewer I look at these from the tree structure of them being in folders then provides the distinction. Interesting to hear this isn't working out for you though. I would ideally like a solution that doesn't duplicate the info in the folder label in the filename though.

seabbs:
> In conclusion, I'm on the fence regarding whether this is a good idea. If going ahead with this there are a few bits of code that could potentially still be moved into chunks.

Yeah, I feel the same way for all the reasons you mention. I think I would much prefer doing this manually (i.e having everything in chunks and then providing R code to generate the complete models (this would need to be done at build time/whenever the chunks were modified (likely in CI to be sure)).

sbfnk:
> > In conclusion, I'm on the fence regarding whether this is a good idea. If going ahead with this there are a few bits of code that could potentially still be moved into chunks.
> 
> Yeah, I feel the same way for all the reasons you mention. I think I would much prefer doing this manually (i.e having everything in chunks and then providing R code to generate the complete models (this would need to be done at build time/whenever the chunks were modified (likely in CI to be sure)). In this mists of time I did some work on this in a branch of `idbrms` that never went anywhere - I think that wouldn't be that hard to revive.
> 
> The issue there would be how we maintain the code for generating the models of course...

Isn't there also an issue that this would require management of between-snippet dependencies, inflating complexity? Unless perhaps we did away with functions.

seabbs:
> Isn't there also an issue that this would require management of between-snippet dependencies, inflating complexity? Unless perhaps we did away with functions.

Hmm I'm not sure how much complexity it would really add to not duplicate functions? Or do you mean something more than this?

sbfnk:
> > Isn't there also an issue that this would require management of between-snippet dependencies, inflating complexity? Unless perhaps we did away with functions.
> 
> Hmm I'm not sure how much complexity it would really add to not duplicate functions? Or do you mean something more than this?

Perhaps I'm misunderstanding what you're suggesting. If we compose a model of chunks then chunks will depend on other chunks (defining functions or declaring variables) which themselves might have dependencies.

seabbs:
Yes but can't you define those in a list? We aren't saying we are going to have a system for composing models - just the models we already have?

sbfnk:
I think we concluded that this wasn't a good idea so I'm closing the issue.

Issue #485, opened 2023-10-24
phi = 1 / sqrt(sqrt_phi) -> phi = 1 / sqrt_phi ^ 2


seabbs:
Epinowcast uses https://github.com/epinowcast/epinowcast/blob/8eff560d1fd8305f5fb26c21324b2bfca1f002b4/inst/stan/epinowcast.stan#L314

`phi = 1 / sqrt_phi ^ 2` and this is also the recommend prior here: [Prior Choice Recommendations · stan-dev/stan Wiki (github.com)](https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations#story-when-the-generic-prior-fails-the-case-of-the-negative-binomial)

seabbs:
Resolved in #487

PR #487, opened 2023-10-25
Update phi parameterisation

seabbs:
renamed PR closing #485


seabbs:
No change in speed but then I would really expect the change to be 1. very small and 2. only show up when the data needs stabilising

sbfnk (approved):

PR #501, opened 2023-11-07
Fix partial matching

### LINKED ISSUE #499: Safeguards against partial matching
Do we want to keep using match.arg or switch to the rlang version which doesn't allow partial matching? This is probably something that requires a new issue but just noting here whilst I think of it.

_Originally posted by @seabbs in https://github.com/epiforecasts/EpiNow2/pull/476#discussion_r1377963069_

jamesmbaazam:
This PR closes #499 by replacing all `match.arg()` calls with `rlang::arg_match()` because the latter prevents partial matching by throwing an error. 

The PR also fixes previously partially matched arguments.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -156,15 +156,16 @@ create_shifted_cases <- function(reported_cases, shift,
 #' time.
 #'
 #' @param delay Numeric mean delay
-#'
+#' 
+#' @importFrom rlang arg_match
 #' @return A list containing a logical called fixed and an integer called from
 #' @author Sam Abbott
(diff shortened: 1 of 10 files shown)


jamesmbaazam:
@seabbs @sbfnk 
Fixing these lints will be out of the scope of this PR. I've actually been meaning to raise an issue about them. Would you recommend we fix and merge the linting issues here then rebase this branch on `main` afterwards?

sbfnk (approved):
Great stuff!

Issue #411, opened 2023-07-03
Discretisation


sbfnk:
The current discretisation of delay distributions integrates between integer boundaries [0, 1], [1, 2], ...

Recent work has shown that this leads to bias. A better option would be to discretise as [0, 1], [0, 2], [1, 3], ...

seabbs:
* Relative to the underlying continuous distribution that may be used as a prior.

jamesmbaazam:
Can you link some of the recent work here?

seabbs:
https://github.com/epinowcast/epinowcast/blob/05b4f64be8ecb5f319c4cffb970698e91434c3a5/inst/stan/functions/discretised_logit_hazard.stan#L35 and see the package docs for the maths behind it.

jamesmbaazam:
Thanks, Sam. I'll take a look.

PR #504, opened 2023-11-10
distribution interface to `dist_spec`

### LINKED ISSUE #382: Reconsider distribution interface.
very much agree on renaming and agree we should/can support the natural scale input with an internal transform. I think we should doc that that isn't an ideal approach though.

_Originally posted by @seabbs in https://github.com/epiforecasts/EpiNow2/pull/363#discussion_r1112992419_

At the moment the `mean` and `sd` variables in `dist_spec` mean different things depending on whether a gamma or lognormal is used. This is not really ideal and it would be good to think about a way to make this more intuitive, where parameters aquire their "natural" meaning.

### LINKED ISSUE #498: gamma_dist_def: scale is inverse scale?
It looks like in `gamma_dist_def` that scale is actually its inverse? This matches with the stan parameterization (which was passed in some models as alpha beta) but not with the R version. 

https://github.com/epiforecasts/EpiNow2/blame/961c8c91063c6da381633ada387e2fba0337c101/R/dist.R#L340

THe renaming slipped through in the recent work on `dist_spec` but I am not clear if there was a deeper functionality issue and if that remains.

sbfnk:
The aim of this PR is to introduce a new distribution interface to replace the current `dist_spec` interface. This should make user interaction more intuitive and be more easily extensible if developers want to add more distributions (Weibull seems an obvious candidate). Eventually (following a deprecation cycle as `dist_spec` became part of a release necessitated by updates to `rstan`) `dist_spec` would become an internal function and the user would only interact with it via the distribution functions.

Examples of the new interface:

``` r
library("EpiNow2")
#> 
#> 
#>     Gamma
Gamma(mean = 3, sd = 1)
#> - gamma distribution:
#>   shape:
#>     9
#>   rate:
#>     3
Gamma(mean = 3, sd = 1, max = 14)
#> - gamma distribution (max: 14):
#>   shape:
#>     9
#>   rate:
#>     3
Gamma(mean = Normal(mean = 3, sd = 1), sd = 4, max = 14)
#> Warning in new_dist_spec(params, "gamma"): Uncertain gamma distribution
#> specified in terms of parameters that are not the "natural" parameters of the
#> distribution (shape, rate). Converting using a crude and very approximate
#> method that is likely to produce biased results. If possible, it is preferable
#> to specify the distribution directly in terms of the natural parameters.
#> - gamma distribution (max: 14):
#>   shape:
#>     - normal distribution:
#>       mean:
#>         0.56
#>       sd:
#>         0.31
#>   rate:
#>     - normal distribution:
#>       mean:
#>         0.19
#>       sd:
#>         0.18
Gamma(shape = Normal(mean = 3, sd = 1), rate = Normal(mean = 2, sd = 0.5), max = 14)
#> - gamma distribution (max: 14):
#>   shape:
#>     - normal distribution:
#>       mean:
#>         3
#>       sd:
#>         1
#>   rate:
#>     - normal distribution:
#>       mean:
#>         2
#>       sd:
#>         0.5
Gamma(shape = 2, rate = 1, max = 14)
#> - gamma distribution (max: 14):
#>   shape:
#>     2
#>   rate:
#>     1
LogNormal(meanlog = 1.5, sdlog = 1)
#> - lognormal distribution:
#>   meanlog:
#>     1.5
#>   sdlog:
#>     1
LogNormal(mean = Normal(1.5, 0.1), sd = Normal(1, 0.1))
#> Warning in new_dist_spec(params, "lognormal"): Uncertain lognormal distribution
#> specified in terms of parameters that are not the "natural" parameters of the
#> distribution (meanlog, sdlog). Converting using a crude and very approximate
#> method that is likely to produce biased results. If possible, it is preferable
#> to specify the distribution directly in terms of the natural parameters.
#> - lognormal distribution:
#>   meanlog:
#>     - normal distribution:
#>       mean:
#>         0.22
#>       sd:
#>         0.043
#>   sdlog:
#>     - normal distribution:
#>       mean:
#>         0.61
#>       sd:
#>         0.071
pmf(c(0.2, 0.4, 0.4))
#> - nonparametric distribution
#>   PMF: [0.2 0.4 0.4]
pmf(c(0.2, 0.4, 0.2, 0.05))
#> - nonparametric distribution
#>   PMF: [0.24 0.47 0.24 0.059]
```

<sup>Created on 2024-01-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Somewhat sad to be masking `base::gamma` and `stats::sd` but perhaps unavoidable - or of course could name them differently.

To do following initial review:
- [x] finalise changes in the stan models to support delay distributions with arbitrary numbers of parameters
- [x] document all new functions
- [x] add appropriate input and other checks (e.g. everything passed to the stan model needs to have a max)
- [x] add warnings when using non-natural uncertain parameterisations
- [x] update tests
- [x] update all uses of `dist_spec` in examples / vignettes
- [x] update `adjust_infection_to_report` to use this and deprecate `gamma_dist_def` and `lnorm_dist_def`
- [ ] write a brief guide for developers on how to add other distributions
- [x] add news item

Tests etc are expected to fail at initial state.

Closes #382
Closes #498


sbfnk:
As of [e152fe6](https://github.com/epiforecasts/EpiNow2/pull/504/commits/e152fe6daae5062cc35d06089d5963c41a66271f) this passes all existing tests locally (with plenty of warnings due to deprecations).

sbfnk:
The attempt at a principled way of converting parameters with uncertainty was abandoned in favour of a simple but very crude one (taking the average relative uncertainty from the untransformed parameters and applying it to the transformed ones), with a warning to the user.

One action is failing due to reasons unrelated to this PR. As far as I'm concerned it's ready to be merged.

seabbs (commented):
This all looks really really nice. From a read this all looks correct.

1. Can we get away with the overwriting of `gamma` etc.? Perhaps we do have to append or preprend something? This would just make things slightly less clean. 
2. It looks like there are some performance issues for models with gaussian processes regardless of delays. Do you think this is just from changing the parameterization of the generation time (and the resulting slightly difference in uncertainty) or do you think this indicates some kind of issue with the updated stan code?

> As part of this we could probably deprecate sample_approx_dist and adjust_infection_report, and replace dist_skel with a simpler version that just calculates the pmf in line with https://github.com/epiforecasts/EpiNow2/issues/411

Yes I agree

vignettes/epinow.Rmd.orig
```diff
@@ -36,10 +36,7 @@ This should be replaced with parameters relevant to the system that is being stu
 library("EpiNow2")
 options(mc.cores = 4)
 reported_cases <- example_confirmed[1:60]
-reporting_delay <- dist_spec(
-  mean = convert_to_logmean(2, 1), mean_sd = 0,
-  sd = convert_to_logsd(2, 1), sd_sd = 0, max = 10
-)
+reporting_delay <- lognormal(mean = 2, sd = 1, max = 10)
 delay <- example_incubation_period + reporting_delay
 rt_prior <- list(mean = 2, sd = 0.1)
```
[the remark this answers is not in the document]
sbfnk: Agree in principle though this would probably require extension of the distribution interface to not automatically discretise, or only to discretise when converting to stan code (possibly a good idea anyway?).

sbfnk:
> Can we get away with the overwriting of gamma etc.? Perhaps we do have to append or preprend something? This would just make things slightly less clean.

One option would be to do as in https://github.com/alexpghayes/distributions3 and capitalise to `Gamma`, `Normal`, `Lognormal` etc. I was wondering if we could build on that package anyway but couldn't see it easily done given our quite specific requirements (especially having parameters with uncertainty). 

A related question I had was whether they should be called `lnorm` and `norm` instead of `lognormal` and `normal`. Not really a fan but it would be more in line with R syntax.

vignettes/estimate_infections_workflow.Rmd
```diff
@@ -230,10 +237,23 @@ def <- estimate_infections(
   delays = delay_opts(delay),
   rt = rt_opts(prior = rt_prior)
 )
-#> Warning: There were 15 divergent transitions after warmup. See
+#> Warning: There were 3 divergent transitions after warmup. See
 #> https://mc-stan.org/misc/warnings.html#divergent-transitions-after-warmup
 #> to find out why this is a problem and how to eliminate them.
+#> Warning: There were 1000 transitions after warmup that exceeded the maximum treedepth. Increase max_treedepth above 15. See
```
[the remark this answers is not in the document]
sbfnk: Problem has seemed to disappear with narrower uncertainty (based on relative variance vs. sd which I think makes sense anyway). Since max treedepth is not a huge problem anyway I think we can proceed.

vignettes/epinow.Rmd.orig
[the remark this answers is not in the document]
sbfnk: https://github.com/epiforecasts/EpiNow2/issues/525

sbfnk:
> Reference in new



> > Can we get away with the overwriting of gamma etc.? Perhaps we do have to append or preprend something? This would just make things slightly less clean.
> 
> One option would be to do as in https://github.com/alexpghayes/distributions3 and capitalise to `Gamma`, `Normal`, `Lognormal` etc. I was wondering if we could build on that package anyway but couldn't see it easily done given our quite specific requirements (especially having parameters with uncertainty).
> 
> A related question I had was whether they should be called `lnorm` and `norm` instead of `lognormal` and `normal`. Not really a fan but it would be more in line with R syntax.

On reflection I think the capitalised functions are the way forward.

sbfnk:
Now updated to have capitalised distribution functions (updated in original reprex).

In doing so the interface has also been updated to seperate it more cleanly from the stan model, for which parameters are now only created at the time of putting together arguments for the model (in `create_stan_delays`). This will permit the distribution interface to be potentially used in other packages, and across this package (e.g. to specify priors of parameters that aren't delay distributions).

sbfnk:
Some actions still failing but for reasons that seem unrelated.

sbfnk:
We could think about deprecating `+` following this as it's now the same as `c` (and tolerance is applied when converting to stan parameters). I like it but it just duplicates the functionality.

We could also think about capitalising `pmf` (e.g. `PMF`) to be in line with the other functions.

sbfnk:
Not for this PR (which is already bloated) but we should update the plotting function now that `dist_spec` can represent continuous as well as discrete distribution - any distribution will have to be discretised for plotting but could be done at arbitrary resolution if the underlying object represents a continuous distribution.

R/dist.R
```diff
@@ -832,314 +867,157 @@ tune_inv_gamma <- function(lower = 2, upper = 21) {
...
+                      params_mean = numeric(0), params_sd = numeric(0),
+                      mean, sd = 0, mean_sd = 0, sd_sd = 0,
+                      max = Inf, pmf = numeric(0), fixed = FALSE) {
+
+  lifecycle::deprecate_warn(
+    "2.0.0",
+    "dist_spec()",
+    details = c(
+      paste0(
+        "Please use distribution functions such as `gamma` or `lognormal` ",
```
seabbs, line 920:
```suggestion
        "Please use distribution functions such as `Gamma()` or `Lognormal()` ",
```

R/get.R
```diff
@@ -177,46 +177,66 @@ get_regional_results <- function(regional_output,
...
-#' @seealso [dist_spec()]
 #' @export
 get_dist <- function(data, disease, source, max_value = 14, fixed = FALSE) {
   lifecycle::deprecate_warn(
-    "2.0.0", "get_dist()", "dist_spec()",
-    "The function will be removed completely in version 2.1.0."
+    "2.0.0", "get_dist()",
+    details = c(
+      paste(
+        "Please use distribution functions such as `gamma()` or `lognormal()`",
```
seabbs, line 186:
```suggestion
        "Please use distribution functions such as `Gamma()` or `Lognormal()`",
``

**Review by seabbs — COMMENTED** [review]
Running just the EpiNow2 vignette I see a depreciation warning for dist_spec but am having trouble tracking it down.

**Review by seabbs — COMMENTED** [review]
Running just the EpiNow2 vignette I see a depreciation warning for dist_spec but am having trouble tracking it down.

**Comment by sbfnk** [comment]
> Running just the EpiNow2 vignette I see a depreciation warning for dist_spec but am having trouble tracking it down.

That vignettes haven't actually been updated yet (and this one is precompiled which may explain why you can't find the warning in this branch).

**Comment by seabbs** [comment]
After f2f discussion I see that: 

the .orig version of the epinow vignette throws a depreciation warning when `epinow` is called. 

I (well actually @sbfnk) see a `dist_spec` call in the `EpiNow2` vignette: https://github.com/epiforecasts/EpiNow2/blob/7fb64a2dbfbcd314402f331230e9d94dfb97625c/vignettes/EpiNow2.Rmd.orig#L43

File: R/opts.R [hunk 5]
```diff
@@ -187,7 +197,7 @@ delay_opts <- function(dist = dist_spec(), ..., fixed = FALSE) {
 #'
 #' # truncation dist
 #' trunc_opts(dist = dist_spec(mean = 3, sd = 2, max = 10))
```
seabbs, line 199:
```suggestion
#' trunc_opts(dist = LogNormal(mean = 3, sd = 2, max = 10))
```

R/dist.R
```diff
@@ -1151,79 +1029,300 @@ dist_spec_plus <- function(e1, e2, tolerance = 0.001) {
...
+    } else {
+      return(y$max)
+    }
+  }, numeric(1))
+  return(ret)
+}
+##
+##' Discretise a <dist_spec>
+##'
+##' By default it will discretise all the distributions it can discretise
```
seabbs, line 1209: Suggest add some content that this represents single interval censoring

R/dist.R
```diff
@@ -1382,48 +1474,374 @@ plot.dist_spec <- function(x, ...) {
...
+##' @inheritParams stats::GammaDist
+##' @author Sebastian Funk
+##' @rdname Distributions
+##' @title Probability distributions
+##' @order 2
+##' @export
+##' @examples
+##' Gamma(mean = 4, sd = 1)
+##' Gamma(shape = 16, rate = 4)
+##' Gamma(shape = Normal(16, 2), rate = Normal(4, 1))
```
seabbs, line 1597:
```suggestion
##' Gamma(shape = Normal(16, 2), rate = Normal(4, 1))
##' Gamma(shape = Normal(16, 2), scale = Normal(1/4, 1))
```

R/dist.R
```diff
@@ -1382,48 +1474,374 @@ plot.dist_spec <- function(x, ...) {
...
+##'
+##' @inheritParams stats::Lognormal
+##' @param mean,sd mean and standard deviation of the distribution
+##' @param max Numeric, maximum value of the distribution. The distribution will
+##' be truncated at this value. Default: `Inf`, i.e. no maximum.
+##' @return A `dist_spec` representing a distribution of the given
+##'   specification.
+##' @author Sebastian Funk
+##' @export
+##' @rdname Distributions
```
seabbs, line 1576: Issue with roxygen here and elsewhere being ##` rather than #`

R/dist.R
```diff
@@ -1382,48 +1474,374 @@ plot.dist_spec <- function(x, ...) {
...
+##' being passed to the stan models.
+##' @param distribution Character; the distribution to use.
+##' @return A character vector, the natural parameters.
+##' @author Sebastian Funk
+##' @keywords internal
+##' @examples
+##' \dontrun{
+##' natural_params("gamma")
+##' }
+natural_params <- function(distribution) {
```
seabbs, line 1659: dist is very long and hard to pass. I think it might be better to split into dist-tools and dist

R/dist.R
```diff
@@ -1382,48 +1474,374 @@ plot.dist_spec <- function(x, ...) {
...
+  rel_unc <- mean(vapply(params, sd_dist, numeric(1))^2 / unlist(ux))
+  ## store natural parameters
+  x <- list()
+  if (distribution == "gamma") {
+    if ("mean" %in% names(ux) && "sd" %in% names(ux)) {
+      x$shape <- ux$mean**2 / ux$sd**2
+      x$rate <- x$shape / ux$mean
+    } else if (!("rate" %in% names(ux)) && ("scale" %in% names(ux))) {
+      x$shape <- ux$shape
+      x$rate <- 1 / ux$scale
```
seabbs, line 1831: This looks like it should work but the test gap means it might not. I'd suggest add a test for the gamma along the lines of the example I suggested and also adding a unit test here

seabbs:
I think the bones of this are good but its proving quite a hassle to review as its such a monolith and there isn't a clear design spec to use as an anchor. 

Would it be helpful for me to point a clean up PR at this PR to deal with minor issues and to write a little scaffold of the design approach that we can agree is how things are implemented?

R/dist.R
sbfnk: 85ec4e2a

sbfnk:
> Would it be helpful for me to point a clean up PR at this PR to deal with minor issues and to write a little scaffold of the design approach that we can agree is how things are implemented?

Yes, that would be great. Branch is now rebased to master and minor suggestions applied (not tested yet and needs re-rendering documentation for checks).

R/dist.R
sbfnk: Split now into:
- `dist_spec.R`: everything related to `<dist_spec>`. This should now not use any other code from the package so could in principle be extracted.
- `deprecated.R`: deprecated functions (could add other ones from the rest of the package here later to improve code readability)
- `estimate_delay.R`: everything related to estimating delays
- moved `sample_approx_dist` to `adjust_infection_to_report` as it's the only place where it's used. Eventually they might be deprecated together as suggested in #520

sbfnk:
> Would it be helpful for me to point a clean up PR at this PR to deal with minor issues and to write a little scaffold of the design approach that we can agree is how things are implemented?

Added a sketch at https://github.com/epiforecasts/EpiNow2/blob/dist-interface/inst/dev/design_dist.md

seabbs:
Amazing. Will pick up the review again tomorrow

sbfnk:
Trying to gather remaining open points (most of which could be dealt with in separate issues after merging):
- decision on `pmf` vs. `Pmf` vs. `PMF`: https://github.com/epiforecasts/EpiNow2/pull/504#issuecomment-1893462674
- decision on possibly removing `+`: https://github.com/epiforecasts/EpiNow2/pull/504#issuecomment-1893462674
- update plotting: https://github.com/epiforecasts/EpiNow2/pull/504#issuecomment-1929143699
- update vignettes: https://github.com/epiforecasts/EpiNow2/pull/504#issuecomment-1944457214
- review/update design sketch: https://github.com/epiforecasts/EpiNow2/pull/504#issuecomment-1964559108

seabbs:
It seems like only the last two are for this PR and the others are issues for future work?

sbfnk:
> It seems like only the last two are for this PR and the others are issues for future work?

Also (1) (`pmf`) I think. Updating vignettes will require some text as well as code updates so potentially better done in a separate PR once we're happy with things as they are.

seabbs:
>  Updating vignettes will require some text as well as code updates so potentially better done in a separate PR once we're happy with things as they are.

I guess the call here is are we happy having broken vignettes on the main branch. IMO I would prefer not. A reasonable solution is to address that in its own PR but targetted at this branch and wait on that before merging?

seabbs:
> We could also think about capitalising pmf (e.g. PMF) to be in line with the other functions.

Do you have a strong view on this? We are either choosing to be consistent with ourselves (`pmf`) or base (`PMF`) right?

sbfnk:
> Do you have a strong view on this?

I don't and was hoping anyone else might. We could also call it `NonParametric(pmf = c(...))` which is perhaps more accurate if more to type?

seabbs:
> We could also call it NonParametric(pmf = c(...)) which is perhaps more accurate if more to type?

Hmm yeah this is quite appealing to me I think. As could have different inputs (for example as a hazard). I think its worth the typing tax

jamesmbaazam:
One minor comment is to provide `discretize()` as an alternative/alias version of `discretise()`.

sbfnk:
`pkgdown.yml` still has to be updated with all the new functions.

sbfnk:
@seabbs still planning to add your review here (happy to wait if so) or easier to raise as issues following a merge if checks pass?

seabbs:
I don't think we should merge something this big in without careful review. 

I'm still in progress on this and have another deadline today but should be able to knock it out tomorrow.

sbfnk:
OK great!

sbfnk:
> I guess the call here is are we happy having broken vignettes on the main branch. IMO I would prefer not.

Vignettes have been fixed / re-rendered in [d1c3cf7](https://github.com/epiforecasts/EpiNow2/pull/504/commits/d1c3cf7236af5c9dd4b0d5f2aebf19cf8224db2c) and [4ffd533](https://github.com/epiforecasts/EpiNow2/pull/504/commits/4ffd5330d3eb678749618d21939cd796a03529aa), respectively.

R/checks.R
```diff
@@ -53,3 +53,53 @@ check_reports_valid <- function(reports, model) {
     assert_numeric(reports$confirm, lower = 0)
   }
 }
+
+#' Validate probability distribution for passing to stan
+#'
+#' @description
+#' `check_stan_delay()` checks that the supplied data is a `<dist_spec>`,
+#' that it is lognormal or gamma, and that it has a finite maximum.
```
seabbs, line 61: out of date with currently supported dists.
```suggestion
#' that it is a supported distribution, and that is has a finite maximum.
```

R/create.R
```diff
@@ -516,6 +516,24 @@ create_stan_data <- function(reported_cases, seeding_time,
   return(data)
 }
 
+##' Create initial conditions for delays
+##'
+##' @inheritParams create_initial_conditions
+##' @return A list of initial conditions for delays
+##' @keywords internal
+create_delay_inits <- function(data) {
```
seabbs, line 524: Something I have been playing around with is dropping this approach and instead sampling from the prior in the stan model. It would be nice as it would reduce duplication but it might be a bit unstable and also requires more stan code to be carried around for other tasks.

NEWS.md
```diff
@@ -9,6 +9,7 @@
 * `simulate_infections` has been renamed to `forecast_infections` in line with `simulate_secondary` and `forecast_secondary`. The terminology is: a forecast is done from a fit to existing data, a simulation from first principles. By @sbfnk in #544 and reviewed by @seabbs.
 * A new `simulate_infections` function has been added that can be used to simulate from the model from given initial conditions and parameters. By @sbfnk in #557 and reviewed by @jamesmbaazam.
 * The function `init_cumulative_fit()` has been deprecated. By @jamesmbaazam in #541 and reviewed by @sbfnk.
+* The interface to generating delay distributions has been completely overhauled. Instead of calling `dist_spec()` users now specify distributions using functions that represent the available distributions, i.e. `LogNormal()`, `Gamma()` and `Fixed()`. Uncertainty is specified using calls of the same nature, to `Normal()`. By @sbfnk in #504 and reviewed by @seabbs.
```
seabbs, line 12: Can we link out to the design doc it seems to give useful context.

seabbs (approved):
I think this is good to be merged in after both another site read and a functionality review. A few minor comments outstanding. 

I do think that any porting of this for use elsewhere needs to come with a refactor/redesign as the current implementation seems quite brittle (all the ifelse and magic string based detection of distributions). 

I also expect there to be a few edge cases this review has missed that will need fixing once this is on main but we will see (and 🤞🏻) .

Nice work on this - a lot of work but definitely nice to use!

seabbs (approved):

PR #511, opened 2023-11-14
Use more consistent representations in function documentation and hyperlink functions

### LINKED ISSUE #508: Hyperlink `*_opts()` functions for better signposting in help files
Let's take `estimate_truncation()`. A lot of the old arguments are now deprecated, with some signposting to the preferred argument `truncation`, which in turn links to `trunc_opts()`. From the user's perspective, this can be frustrating. We need to hyperlink the `*_opts()` functions so that users don't have to manually type `?trunc_opts()` to read about its usage.

### LINKED ISSUE #510: Consistency in documentation style
Would be nice to adopt a consistent style in the documentation in upcoming documentation and apply it retrospectively as well. 

I would suggest the following:

* Functions should be written as `[function()]` (especially with the closing brackets to differentiate from objects or function arguments).
* Objects and arguments as `object` or `argument`.
* Classes as `<class>`, for example, "An `<data.frame>`".

Thoughts?

jamesmbaazam:
This PR fixes #508 by hyperlinking all referenced functions in the function documentation across the code base. It also closes #510 by applying a consistent style and formatting to functions, arguments, R packages, objects, and classes.


jamesmbaazam:
@seabbs Here, https://github.com/epiforecasts/EpiNow2/blob/65e6c5a5ac8d08daa677ea8a3086913328b0bac5/R/get.R#L6, has the name `regional_rt_pipeline` changed? I can't seem to find it.

seabbs:
Yup I think it's no regional_epinow (as discussed today)

R/create.R
```diff
@@ -144,8 +144,8 @@ create_shifted_cases <- function(reported_cases, shift,
 #' Construct the Required Future Rt assumption
 #'
 #' @description `r lifecycle::badge("stable")`
-#' Converts the `future` argument from `rt_opts()` into arguments that can be
-#' passed to `stan`.
+#' Converts the `future` argument from [rt_opts()] into arguments that can be
+#' passed to `{stan}`.
```
seabbs, line 148: It's not an r package so do you think this syntax approach is the right?

seabbs (dismissed):
Thanks James this looks great. Only comment is about if Stan should be wrapped in brackets as it's not an R package?

R/create.R
jamesmbaazam: You're right. We should probably reserve this for only R packages. I guess we should not format it all, i.e., 
`{stan}` -> stan.

jamesmbaazam: This has been fixed in 5921c926751924ca11f1397baa16582d0e192c12.

jamesmbaazam:
@seabbs I believe this should be ready for merge.

sbfnk (approved):
Great work.

PR #518, opened 2023-11-17
Discretise with two-day censoring windows

### LINKED ISSUE #411: Discretisation

sbfnk:
This has been shown to reduce bias in the discretisation. More detail in https://package.epinowcast.org/dev/articles/distributions.html

Closes #411


sbfnk:
Some delay-related tests failing - will revisit after merging #504

jamesmbaazam (approved):
I've crosschecked this against [the epinowcast vignette on discredited delays](https://package.epinowcast.org/dev/articles/distributions.html#discretisation-and-adjustment-for-maximum-delay) and it looks good to me.

sbfnk:
> I've crosschecked this against [the epinowcast vignette on discretised delays](https://package.epinowcast.org/dev/articles/distributions.html#discretisation-and-adjustment-for-maximum-delay) and it looks good to me.

Thanks! Added you as a reviewer - if happy please re-approve.

jamesmbaazam (approved):

seabbs (commented):
Sorry i haven't had a chance to fully review but the stan code looks in line with what I expect. Happy to see this merged once @jamesmbaazam is happy (I think we want a big news note about this at the least though as it will change what people are getting).

I think this is true but also we should double check that the R side discretisation implementation is either the same approximation or the more accurate simulation base approach epinowcast uses. 

Note both of these should be replaced soon with and at least R side standalone package that handles this all numerically.

sbfnk:
> I think this is true but also we should double check that the R side discretisation implementation is either the same approximation or the more accurate simulation base approach epinowcast uses.

Yes it should be - we're already testing that in https://github.com/epiforecasts/EpiNow2/blob/93c072d9aad839ccda76fa75f74ebc9d403f4d26/tests/testthat/test-dist.R#L4

jamesmbaazam (approved):
I'm happy for this to be merged.

PR #528, opened 2023-12-18
Add support for missing NAs in estimate_infection() model

seabbs:
This PR adds support for missing data in the form of NAs but dropping them from the likelihood rather than imputing that they are zero and then optionally 

Note that this is a breaking change and may be undesirable for some users where missing dates do indicate zeros vs being truly missing. 

I have tested by knocking out data from the example and running the model.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -431,7 +431,7 @@ create_stan_data <- function(reported_cases, seeding_time,
 
   cases <- reported_cases[(seeding_time + 1):(.N - horizon)]$confirm
   cases <- cases$confirm
-  cases[, lookup := seq_len(.N))]]
+  cases[, lookup := seq_len(.N)]
   complete_cases <- cases[!is.na(cases$confirm)]
   complete_cases <- cases$confirm
   cases_time <- cases$lookup
--- a/R/create.R
+++ b/R/create.R
@@ -13,8 +13,10 @@
 #' threshold then the zero is replaced using `fill`.
 #'
 #' @param fill Numeric, defaults to NA. Value to use to replace NA values or
-#' zeros that are flagged by `zero_threshold`. If the default NA is used then
-#' dates with NA values will be skipped in model fitting.
+#' zeroes that are flagged because the 7-day average is above the
+#' `zero_threshold`. If the default NA is used then dates with NA values or with
+#' 7-day averages above the `zero_threshold` will be skipped in model fitting. If
+#' this is set to 0 then the only effect is to replace NA values with 0.
 #'
 #' @inheritParams estimate_infections
 #' @importFrom data.table copy merge.data.table setorder setDT frollsum
--- a/R/create.R
+++ b/R/create.R
@@ -9,14 +9,14 @@
 #'
 #' @param zero_threshold `r lifecycle::badge("experimental")` Numeric defaults
 #' to Inf. Indicates if detected zero cases are meaningful by using a threshold
-#' number of cases based on the 7 day average. If the average is above this
+#' number of cases based on the 7-day average. If the average is above this
 #' threshold then the zero is replaced using `fill`.
 #'
 #' @param fill Numeric, defaults to NA. Value to use to replace NA values or
 #' zeroes that are flagged because the 7-day average is above the
 #' `zero_threshold`. If the default NA is used then dates with NA values or with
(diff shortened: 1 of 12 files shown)


R/create.R
```diff
@@ -1,27 +1,33 @@
...
 #' start of the time series be filtered out.
 #'
 #' @param zero_threshold `r lifecycle::badge("experimental")` Numeric defaults
 #' to Inf. Indicates if detected zero cases are meaningful by using a threshold
 #' number of cases based on the 7 day average. If the average is above this
-#' threshold then the zero is replaced with the backwards looking rolling
-#' average. If set to infinity then no changes are made.
+#' threshold then the zero is replaced using `fill`.
+#'
+#' @param fill Numeric, defaults to NA. Value to use to replace NA values or
```
sbfnk, line 15: I'm a bit at loss as to what one would ever set this to if not NA. Can we add an example? Or else just make it NA and remove the argument (my preference until convinced of the need for it)?

R/create.R
```diff
@@ -35,35 +41,33 @@ create_clean_reported_cases <- function(reported_cases, horizon,
...
+      date >= min(date[confirm[!is.na(confirm)] > 0])
+    ]
   }
-
+  # Calculate the 7-day moving average.
+  reported_cases <-
+    reported_cases[
+      ,
+      `:=`(average_7_day = (
+          data.table::frollsum(confirm, n = 8, na.rm = TRUE)
```
sbfnk, line 57: Should this not be `n = 7`?

R/create.R
```diff
@@ -451,11 +467,11 @@ create_stan_data <- function(reported_cases, seeding_time,
       delay = data$seeding_time, horizon = data$horizon
     )
   )
-  # initial estimate of growth
+  # initial estimate of growth33
```
sbfnk, line 470:
```suggestion
  # initial estimate of growth
```

sbfnk (changes requested):
Very nice! A couple of minor questions/confusions on my side.

R/create.R
seabbs: so you might want to set this to zero to get the old behaviour back - in some set of cases that will be what is actually going on in peoples data (which is why we had it like this in the first place) but could easily argue that should be handled externally/differently

R/create.R
seabbs: I think that this is just moving code around that was here before and I think the logic is that this is the 7 day moving average horizon = -1 where the current day is 0. Does that logic hold up? I am not sure!

R/create.R
```diff
@@ -35,35 +41,33 @@ create_clean_reported_cases <- function(reported_cases, horizon,
...
   if (filter_leading_zeros) {
     reported_cases <- reported_cases[order(date)][
-      ,
-      cum_cases := cumsum(confirm)
-    ][cum_cases > 0][, cum_cases := NULL]
+      date >= min(date[confirm[!is.na(confirm)] > 0])
+    ]
   }
-
+  # Calculate the 7-day moving average.
```
sbfnk, line 52:
```suggestion
  # Calculate `average_7_day` which for rows with `confirm == 0`
  # (the only instance where this is being used) equates to the 7-day
  # right-aligned moving average at the previous data point.
```

R/create.R
```diff
@@ -1,27 +1,33 @@
...
 #' @param zero_threshold `r lifecycle::badge("experimental")` Numeric defaults
 #' to Inf. Indicates if detected zero cases are meaningful by using a threshold
 #' number of cases based on the 7 day average. If the average is above this
-#' threshold then the zero is replaced with the backwards looking rolling
-#' average. If set to infinity then no changes are made.
+#' threshold then the zero is replaced using `fill`.
+#'
+#' @param fill Numeric, defaults to NA. Value to use to replace NA values or
+#' zeros that are flagged by `zero_threshold`. If the default NA is used then
+#' dates with NA values will be skipped in model fitting.
```
sbfnk, line 17:
```suggestion
#' zeroes that are flagged because the 7-day average is above the
#' `zero_threshold`. If the default NA is used then dates with NA values or with
#' 7-day averages above the `zero_threshold` will be skipped in model fitting.
```

sbfnk (dismissed):

sbfnk (approved):

Issue #512, opened 2023-11-15
Use of choices here and elsewhere should be moved into the argument default


seabbs:
https://github.com/epiforecasts/EpiNow2/blob/65e6c5a5ac8d08daa677ea8a3086913328b0bac5/R/summarise.R#L802

sbfnk:
Rationale being that this exposes to the user what the options are.

PR #534, opened 2024-01-22
Add option to accumulate observations

### LINKED ISSUE #531: Accumulate observations where data are NAs
Thanks to the work in https://github.com/epiforecasts/EpiNow2/pull/528 we now support NAs in the data and use an index for observations. This opens up the possibility of alternatively accumulating observations on dates where the data is NA, providing a potentially more consistent interface for modelling data reported at a non-daily time step than what there is at the moment.

sbfnk:
This PR adds an argument `na` to `obs_opts` with an `"accumulate"` option that adds any NA data points to the following data point in the inference. The idea is that this will allow fitting data at any time interval (e.g. weekly) including where spacing is irregular (e.g., monthly, or weekly where on some occasions data is reported a day later because of a holiday etc.).

Two points for potential discussion are:
1. When setting this to `accumulate` the first data point is ignored as we don't know when we should start accumulating for that one. This means that for a someone who would like that data point to be considered it would be advantageous to add a dummy data point to the beginning of the data set. We could add an option to do that, or document it, or leave it unmentioned, or find another solution.
2. Because we either skip or accumulate NAs we can't combine them, e.g. to have weekly data with missing values. Enabling this would, I think, require some sort of that marks dates explicitly as missing vs. NA.

~I've updated `estimate_secondary` to still work with the examples/tests but it doesn't work with NA values yet. This is for another issue/PR.~ It was necessary to update `estimate_secondary` in order to pass tests/checks. This now also works with NA values.

Closes #531


seabbs (commented):
Really nice. 

> Enabling this would, I think, require some sort of that marks dates explicitly as missing vs. NA.

I think this would be my preferred option as it would be more general but I also think it can be addressed in its own review as it would be a superset of this PR. 

My thought on how that would work is to have a new variable (`accumulate`) that indicates which days should be summed.

R/opts.R
```diff
@@ -446,6 +446,15 @@ gp_opts <- function(basis_prop = 0.2,
 #' empty a mean (`mean`) and standard deviation (`sd`) needs to be supplied
 #' defining the normally distributed scaling factor.
 #'
+#' @param na Character. Options are "missing" (the default) and "accumulate".
+#'   This determines how NA values in the data are interpreted. If set to
```
seabbs, line 450: Sorry this is so boring but indents here are not in line with the rest of this doc....

R/opts.R
```diff
@@ -471,18 +480,22 @@ obs_opts <- function(family = "negbin",
                      week_effect = TRUE,
                      week_length = 7,
                      scale = list(),
+                     na = "missing",
```
seabbs, line 483:
```suggestion
                     na = c("missing", "accumulate"),
```
Any reason not to handle it like this?

inst/stan/functions/observation_model.stan
```diff
@@ -51,22 +51,46 @@ void truncation_lp(array[] real truncation_mean, array[] real truncation_sd,
   }
 }
 // update log density for reported cases
-void report_lp(array[] int cases, vector reports,
+void report_lp(array[] int cases, array[] int cases_time, vector reports,
                array[] real rep_phi, real phi_mean, real phi_sd,
-               int model_type, real weight) {
+               int model_type, real weight, int accumulate) {
+  int n = num_elements(cases_time) - accumulate; // number of observations
```
seabbs, line 57: I would have thought an approach that used a lookup for each case_obs would have been more efficient here but this looks correct. I think any changes to this can be addressed in a following PR.

inst/stan/functions/observation_model.stan
```diff
@@ -51,22 +51,46 @@ void truncation_lp(array[] real truncation_mean, array[] real truncation_sd,
...
+        obs_reports[current_obs] += reports[i];
+      }
+      if (i == cases_time[current_obs + 1]) {
+        current_obs += 1;
+      }
+      i += 1;
+    }
+    obs_cases = cases[2:(n + 1)];
+  } else {
+    obs_reports = reports[cases_time];
```
seabbs, line 76: I guess I am saying that cases_time would become optionally an array with a sum. Another option might be to make this a matrix multiplication problem

NEWS.md
```diff
@@ -25,6 +25,7 @@
 ## Model changes
 
 * Updated the parameterisation of the dispersion term `phi` to be `phi = 1 / sqrt_phi ^ 2` rather than the previous parameterisation `phi = 1 / sqrt(sqrt_phi)` based on the suggested prior [here](https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations#story-when-the-generic-prior-fails-the-case-of-the-negative-binomial) and the performance benefits seen in the `epinowcast` package (see [here](https://github.com/epinowcast/epinowcast/blob/8eff560d1fd8305f5fb26c21324b2bfca1f002b4/inst/stan/epinowcast.stan#L314)). By @seabbs in # and reviewed by @sbfnk.
+* Added an `na` argument to `obs_opts()` that allows the user to specify whether NA values in the data should be interpreted as missing or accumulated in the next non-NA data point. By @sbfnk.
```
seabbs, line 28:
```suggestion
* Added an `na` argument to `obs_opts()` that allows the user to specify whether NA values in the data should be interpreted as missing or accumulated in the next non-NA data point. By @sbfnk in #534 and reviewed by @seabbs.
```

seabbs (changes requested):
This all looks good and is a nice feature to have. Some reservations about the precise implementation but as flagged can be addressed by follow up work.

> This means that for a someone who would like that data point to be considered it would be advantageous to add a dummy data point to the beginning of the data set. We could add an option to do that, or document it, or leave it unmentioned, or find another solution.

I think my preferred option here would be to add a message when this method is used? (i.e dropping the first data point and use a dummy if you wish).

R/opts.R
sbfnk: 😲 blame Emacs. Will update.

R/opts.R
sbfnk: Oh I thought we were doing it the other way round but looking at #512 we aren't. I agree that this is clearer.

R/opts.R
sbfnk: Actually this seems to be what roxygen suggests, and it does make the code more readable.
https://mpn.metworx.com/packages/roxygen2/7.1.0/articles/rd.html#object-specifics
Will change all of the ones here to be consistent.

inst/stan/functions/observation_model.stan
sbfnk: Something like a matrix $M_{ij}$ which is 0 except if modelled observation $i$ gets added to true observation $j$? It's going to be quite sparse so do you think actually more efficient/ Or perhaps I'm misunderstanding.

And the array version would be the same, i.e. `cases_time` would be $C_i = j$ following the notation above?

seabbs: > It's going to be quite sparse so do you think actually more efficient? Or perhaps I'm misunderstanding.

Yes exactly and then handle it explicitly as a sparse matrix. I don't really know if it would be more efficient but my mind naturally likes to jump to the most complex possible solution :)

seabbs: > And the array version would be the same, i.e. cases_time would be 
 following the notation above?

Yes

seabbs (approved):
LGTM to me. I think the outstanding points are all for a new issues so if you agree we can resolve and merge?

seabbs:
> Might be annoying but will remind us to implement a better solution if so.

If things move over to `{cli}` this could be made a once per session thing to tone it down a bit.

seabbs (approved):

PR #537, opened 2024-02-02
Cmdstanr option

### LINKED ISSUE #48: Option for cmdstanr?
Just a general question. I built a fork using [cmdstanr](https://mc-stan.org/cmdstanr/) which should be extensible and avoid some of the rstan issues in the future. Downside is that cmdstanr isn't (yet) on CRAN and it adds an external system requirement for cmdstan. I don't know if there is interest, but I could update my code and include cmstanr as an option. It could be an argument to `estimate_infections`

```
 use_cmdstan = FALSE # in arguments
```

Then in body of function:

```
if (use_cmdstan){
    
    if(is.null(model)){
      model <- system.file("stan", "estimate_infections.stan", package = "EpiNow2")
    }
    
    #stan_file <- here::here("inst", "stan", "estimate_infections.stan")
    
    mod <- cmdstanr::cmdstan_model(model, compile = TRUE)
    
    if (verbose) {
      message(paste0("Running for ", samples," samples (across ", chains, 
                     " chains each with a warm up of ", warmup, " iterations each) and ",
                     data$t," time steps of which ", horizon, " are a forecast"))
    }
    
    if(is.null(threads_per_chain)){
      threads_per_chain <- 1L
    }
    
    cmdfit <- mod$sample(data = data, chains = chains,
                      parallel_chains = cores,
                      init = init_fun, iter_warmup = warmup,
                      iter_sampling =  ceiling(samples / chains), 
                      max_treedepth = max_treedepth,
                      threads_per_chain = threads_per_chain,
                      adapt_delta = adapt_delta,
                      refresh = ifelse(verbose, 50, 0),
                      save_warmup = debug)
    
    fit <- rstan::read_stan_csv(cmdfit$output_files())
```
 I don't know how much IO time could add, but I have been finding cmdstanr is more stable.

sbfnk:
Closes #48 by implementing a `backend = "cmdstanr"` option.

This contains everything in #472 except the profiling bits (which don't seem supported by `rstan`).

This necessitated some (ultimately probably positive) streamlining of how the stan arguments are constructed for the different models.


sbfnk:
All checks now passing except windows one which looks unrelated.

seabbs:
> All checks now passing except windows one which looks unrelated.

Agree, I see this in multiple places including the `cmdstanr` repo.

.github/workflows/R-CMD-as-cran-check.yaml
```diff
@@ -36,9 +36,18 @@ jobs:
...
         with:
-          extra-packages: any::rcmdcheck
-          needs: check
+          dependencies: NA
+          extra-packages: |
+            rcmdcheck
+            stan-dev/cmdstanr
+            testthat
+
+      - name: Install cmdstan
```
seabbs, line 45: I've just written a cached version of this for epinowcast (in github.com/epinowcast/actions) that we can either use directly or pull into the repo in order to slightly speed up CI

.gitignore
```diff
@@ -37,3 +37,10 @@ src
 .vscode
 
 vignettes/results
+
+inst/stan/dist_fit
```
seabbs, line 41: It might be marginally safer to `.gitignore` all of `inst/stan` and then allow just `*.stan`. Pretty minimal though

R/estimate_infections.R
```diff
@@ -315,28 +309,31 @@ estimate_infections <- function(reported_cases,
 #' @importFrom futile.logger flog.debug
 #' @importFrom utils capture.output
 #' @inheritParams fit_model_with_nuts
+#' @inheritParams stan_opts
 #' @return A stanfit object
 #' @author Sam Abbott
 init_cumulative_fit <- function(args, samples = 50, warmup = 50,
```
seabbs, line 315: we should really depreicate this at some point. I don't think it helps and its just code to carry around

R/extract.R
```diff
@@ -53,16 +53,82 @@ extract_static_parameter <- function(param, samples) {
...
+  }
+  samples_df <- data.table::data.table(stan_fit$draws(
+    variables = pars, format = "df")
+  )
+  # convert to rstan format
+  samples_df <- suppressWarnings(data.table::melt(
+    samples_df, id.vars = c(".chain", ".iteration", ".draw")
+  ))
+  samples_df <- samples_df[,
+    index := sub("^.*\\[([0-9,]+)\\]$", "\\1", variable)
```
seabbs, line 94: 😱

seabbs (commented):
This looks really great and I think its all in place. I've just done a read through and will play with the new functionality and then review again

sbfnk:
Skipping checks on Windows now so all passing.

@seabbs for info the action from epinowcast caused errors on macOS so reverted to the previous implementation.

seabbs:
>  action from epinowcast caused errors on macOS so reverted to the previous implementation

Yeah not sure what is going on there as it didn't happen in testing. Something to do with cross platform caching I assume. Will dig

R/create.R
```diff
@@ -630,8 +630,14 @@ create_initial_conditions <- function(data) {
...
 #'
-#' @param init Initial conditions passed to `{rstan}`. Defaults to "random" but
-#' can also be a function (as supplied by [create_initial_conditions()]).
+#' @param init Initial conditions passed to `{rstan}`. Defaults to "random"
+#' (initital values randomly drawn between -2 and 2) but can also be a
+#' function (as supplied by [create_initial_conditions()]).
+#'
+#' @param model Character, name of the model for which arguments are
+#' to be created.
+#' @param fixed_param Logiccal, defaults to `FALSE`. Should arguments be
```
seabbs, line 639:
```suggestion
#' @param fixed_param Logical, defaults to `FALSE`. Should arguments be
```

R/create.R
```diff
@@ -630,8 +630,14 @@ create_initial_conditions <- function(data) {
 #'
 #' @param data A list of stan data as created by [create_stan_data()]
 #'
-#' @param init Initial conditions passed to `{rstan}`. Defaults to "random" but
-#' can also be a function (as supplied by [create_initial_conditions()]).
+#' @param init Initial conditions passed to `{rstan}`. Defaults to "random"
+#' (initital values randomly drawn between -2 and 2) but can also be a
```
seabbs, line 634:
```suggestion
#' (initial values randomly drawn between -2 and 2) but can also be a
```

R/stan.R
```diff
@@ -0,0 +1,88 @@
...
+##' Return a stan model object for the appropriate backend
+##'
+##' @inheritParams stan_opts
+##' @inheritParams package_model
+##' @return A stan model object (either \code{rstan::stanmodel} or
+##'   \code{cmdstanr::CmdStanModel}, depending on the backend)
+##' @author Sebastian Funk
+##' @keywords internal
+stan_model <- function(backend = "rstan", model = "estimate_infections") {
+  if (backend == "cmdstanr") {
```
seabbs, line 64:
```suggestion
  backend <- match.arg(backend)
```

R/stan.R
```diff
@@ -0,0 +1,88 @@
...
+
+##' Return a stan model object for the appropriate backend
+##'
+##' @inheritParams stan_opts
+##' @inheritParams package_model
+##' @return A stan model object (either \code{rstan::stanmodel} or
+##'   \code{cmdstanr::CmdStanModel}, depending on the backend)
+##' @author Sebastian Funk
+##' @keywords internal
+stan_model <- function(backend = "rstan", model = "estimate_infections") {
```
seabbs, line 63:
```suggestion
stan_model <- function(backend = c("rstan", "cmdstanr"), model = "estimate_infections") {
```

I think arguably model should also be scoped to the actual list of models we have but as this is an internal function it seems low risk (and it will clearly error if misued).

R/stan.R
```diff
@@ -0,0 +1,88 @@
...
+  return(object)
+}
+
+#' Fit a model using the chosen backend.
+#'
+#' Internal function for dispatch to fitting with NUTS or VB.
+#' @inheritParams fit_model_with_nuts
+#' @keywords internal
+fit_model <- function(args, id = "stan") {
+  if (args$method == "sampling") {
```
seabbs, line 78:
```suggestion
  args$method <- arg_match(args$method, choices = c("sampling", "vb"))
  if (args$method == "sampling") {
```

R/opts.R
```diff
@@ -678,33 +765,57 @@ rstan_opts <- function(object = NULL,
...
                       backend = "rstan",
                       init_fit = NULL,
                       return_fit = TRUE,
                       ...) {
-  backend <- arg_match(backend, values = "rstan")
-  if (backend == "rstan") {
-    opts <- rstan_opts(
-      samples = samples,
-      ...
+  method <- arg_match(method, values = c("sampling", "vb"))
```
seabbs, line 792: Given these are checked here my other suggestions may not really be needed but it also seems like it can't hurt.

tests/testthat/test-epinow.R
```diff
@@ -38,6 +38,33 @@ test_that("epinow produces expected output when run with default settings", {
...
+    )
+  )))
+
+  expect_equal(names(out), expected_out)
+  df_non_zero(out$estimates$samples)
+  df_non_zero(out$estimates$summarised)
+  df_non_zero(out$estimated_reported_cases$samples)
+  df_non_zero(out$estimated_reported_cases$summarised)
+  df_non_zero(out$summary)
+  expect_equal(names(out$plots), c("infections", "reports", "R", "growth_rate", "summary"))
```
seabbs, line 65:
```suggestion
  expect_equal(
    names(out$plots), c("infections", "reports", "R", "growth_rate", "summary")
  )
```

seabbs (approved):
This is really nice. Checked out locally and everything is now working as expected (it looks like the issue I was having was on my end but no idea what it was). A few minor comments and a typo correction. 

My only more substantive comment is that this approach (i.e setting the backend and then adapting all the other inputs is similar to what `brms` does and is potentially nice for the user but will likely be a bit of a pain to extend vs passing in a module that handles fitting (perhaps with some S3 class). Not suggesting we redesign here as this approach makes a lot of sense with how `EpiNow2` is plumbed.

R/stan.R
sbfnk: Yes, may as well. Will implement this and the suggestion for `backend` with a `values` option to `arg_match` to be consistent with elsewhere.

sbfnk: The other way round actually.

seabbs (approved):
LGTM

seabbs (dismissed):
LGTM

seabbs (approved):

PR #540, opened 2024-02-05
Fix tests

### LINKED ISSUE #539: snapshots require update due to data.table printing change
As of data.table 1.15.0 the class is printed by default (see Note 14 in the corresponding [NEWS](https://cran.r-project.org/web/packages/data.table/news/news.html)). This will cause snapshot tests to fail until they are updated.

sbfnk:
This PR fixes various recent testing failures/warnings.

Closes #539


sbfnk:
Merging as minor (no NEWS item needed either).

Issue #538, opened 2024-02-05
repo size


sbfnk:
The repo has grown fairly large (~1 GB), but the files currently in the repo are only 11 MB in size. It might be nice, particularly towards those on low bandwidth connections or paying by volume, to look at reducing the size without using any relevant development history.

Using `git filter-repo --analyze` reveals a few potential easy gains:

```sh
> cat filter-repo/analysis/directories-deleted-sizes.txt 
=== Deleted directories by reverse size ===
Format: unpacked size, packed size, date deleted, directory name
     6810551    3891751 2020-07-22 docs
     6341588    3835465 2020-07-22 docs/reference
     4357982    3715367 2020-07-22 docs/reference/figures
     3724958    1913831 2022-12-19 deps/bootstrap-5.1.3
     3118580    1835087 2022-12-19 deps/bootstrap-5.1.3/fonts
    23619299     747946 2023-02-03 src
     9092536     729112 2023-01-17 inst/pkg-structure
       53863       8984 2023-02-02 .devcontainer
       35864       5688 2023-02-02 .devcontainer/library-scripts
       26185       1104 2020-07-22 docs/news
           0         90 2022-10-15 tests/testthat/test-data
```

```sh
> head -n 10 filter-repo/analysis/path-deleted-sizes.txt 
=== Deleted paths by reverse accumulated size ===
Format: unpacked size, packed size, date deleted, path name(s)
    65775542   64539999 2020-11-30 synthetic.rds
     5907877    5236611 2020-11-08 man/figures/unnamed-chunk-11-1.png
     5207701    4641809 2021-06-03 reference/figures/unnamed-chunk-11-1.png
     3976705    3530160 2020-11-08 man/figures/unnamed-chunk-12-1.png
     3776311    3362607 2020-07-22 docs/reference/figures/unnamed-chunk-11-1.png
     3552356    3153533 2021-06-03 reference/figures/unnamed-chunk-12-1.png
     3143380    3080722 2023-10-03 data/example_regional_epinow.rda
     1619219    1504604 2021-06-03 reference/epinow-5.png
```

At the very least this suggests to me that all the directories above, as well as all png files in `man/figures` (which, if I understand correctly, aren't used anywhere) could be purged. A line to exclude png files in `man/figures` could also be added to `.gitignore`. This could be followed by a deeper investigation of blob sizes for existing files.

seabbs:
[...]

PR #542, opened 2024-02-06
remove unused figures

sbfnk:
see #538


sbfnk:
Minor change not affecting functionality so merging without review or news item.

PR #545, opened 2024-02-13
Case study on reconstructing infections for COVID and Ebola

adamkucharski:
This is a draft PR with a vignette illustrating how to reconstruct infections from delayed outcomes for COVID and Ebola. This code is likely to be streamlined further with incoming functionality to `epiparameter` and `EpiNow2` – notes are given in bold in the `.Rmd`. There is also an issue with current estimation of infections for Ebola, which are estimated to exponential but should be a small outbreak – would value comments/suggestions!

*Subsequent note: I explored the issue of mis-estimation of infections in more detail. Initially wondered if it might be a period with zero initial cases in the dataset, but omitting these did not fix the problem. So instead compared the $R_t$ model with the `rt=NULL` non-mechanistic version.*

*This is the (erroneous) estimate I got with `rt=NULL`:*
rt_null

*And the plausible version with $R_t$ implementation:*
rt_fitted

*I have updated the vignette to include my understanding of why it might be preferable (estimation using low counts works better with a more constrained model process), but there may well be a better explanation. In any case, probably useful to show users a case study both with large counts (COVID) and small counts (Ebola) to demonstrate functionality – we use similar examples in the {cfr} package to show advantages/limitations of different approaches.*

### DIFF
--- a/vignettes/case_study_estimating_infection_dynamics.Rmd
+++ b/vignettes/case_study_estimating_infection_dynamics.Rmd
@@ -36,7 +36,7 @@ We want to estimate infection dynamics from incidence data on delayed outcomes s
 :::
 
 First, we load the necessary packages for this vignette.
-```{r, include =FALSE}
+```{r}
 # Load EpiNow2 1.4.9 version with dist-interface available (TO DO: replace later when in main)
 remotes::install_github("epiforecasts/EpiNow2@dist-interface")
 library(EpiNow2)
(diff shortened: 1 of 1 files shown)


vignettes/case_study_estimating_infection_dynamics.Rmd
```diff
@@ -0,0 +1,319 @@
...
+serial_ebola_params <- get_parameters(serial_int_ebola_discrete) # Get parameters
+
+serial_interval_ebola <- Gamma(
+  shape = serial_ebola_params[["shape"]],
+  rate = 1/serial_ebola_params[["scale"]],
+  max = serial_int_ebola_discrete_max
+)
+```
+
+**Notes: Gamma() function did not seem to take a shape/scale parameterisation ("Incompatible combination of parameters of a gamma distribution specified.")**
```
avallecam, line 278: I found this too and filled an issue https://github.com/epiforecasts/EpiNow2/issues/580

sbfnk:
Thanks for this PR, @adamkucharski! Really nice to see this application, and great to surface this use case of the package model.

In terms of where this sits it might be best to host it outside the package and add it to the [repository of case studies](https://epiforecasts.io/EpiNow2/dev/articles/case-studies.html) so as not to keep the package documentation maintainable and not to add unneeded dependencies here: in this case it would we'd have to add `dplyr` , `httr`, `cfr`, and `epiparameter` only for this vignette in order to pass checks.

With this in mind I'll close the PR and invite you to either put it in its own repository (which would enable full reproducibility with e.g. `renv`) or a gist and then open a new PR to add it to the case studies. If anyone disagrees with this approach please feel free to re-open.

vignettes/case_study_estimating_infection_dynamics.Rmd
sbfnk: This should be fixed now.

adamkucharski:
That sounds sensible (hadn't appreciated the dependency headaches that vignette-only packages would cause). Will look at moving to separate repo, and updating with the new interoperability functionality coming down the line for epiparameter/EpiNow2.

In meantime, would be interested to hear thoughts on the `rt=NULL` issue for small datasets above.

sbfnk:
> In meantime, would be interested to hear thoughts on the `rt=NULL` issue for small datasets above.

Opened an issue in https://github.com/epiforecasts/EpiNow2/issues/611 and will propose a fix.

For what it's worth in this case the problem can also be fixed by using a zero-centered prior (`backcalc = backcalc_opts(prior = "none")`) instead of one centered around shifted reported cases (which is more relevant when trying to use for real-time estimates).

adamkucharski:
For reference, have moved draft to this branch: https://github.com/epiverse-trace/howto/tree/additional-inference

PR #551, opened 2024-02-14
install cmdstanr in touchdown

sbfnk:
seeing action failures which this is trying to fix


sbfnk:
This seems to fix the immediate problem so merging and raising the local fixes as PRs in touchstone.

PR #586, opened 2024-02-28
don't install pandoc in synthetic validation

### LINKED ISSUE #585: Synthetic validation action is failing
This seems to be due to a [recent change](https://github.com/r-lib/actions/commit/890cb242244424bbe759953ada2ffaf55333fce2) in the `setup-r-dependencies` action which now fails if suggested packages can't be found (`cmdstanr` in our case). A simple fix is probably to set `install-pandoc` to `false` where it isn't needed (e.g. synthetic validation).

sbfnk:
## Description

This PR closes #585 (if it works) by explicitly not installing pandoc.

### DIFF
--- a/.github/workflows/synthetic-validation.yaml
+++ b/.github/workflows/synthetic-validation.yaml
@@ -25,6 +25,7 @@ jobs:
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           dependencies: NA
+          install-pandoc: false
           extra-packages: |
             here
             dplyr


seabbs:
@jamesmbaazam tagged you for review as it seems like you could do more review (:)) happy to be pinged instead if not feeling like it (or you know busy)

jamesmbaazam:
I'm not sure I understand this fix. Could you please explain how turning off installation of `pandoc` prevent the failure due to `cmdstanr` not being found?

jamesmbaazam (commented):
I'm not sure I understand this fix. Could you please explain how turning off installation of pandoc prevent the failure due to cmdstanr not being found?

sbfnk:
Good question! Sorry for not explaining better.

The `setup-r-dependencies` now calls `pak::pkg_deps` to check if `rmarkdown` is a package dependence. 
https://github.com/r-lib/actions/blob/4307cec77e4003d61a3b23a07781dcf847c147d7/setup-r-dependencies/action.yaml#L152

This fails if a package isn't available (in our case, `cmdstan` because it is not on CRAN). 

I'm not sure if there's a good solution. We can't add a `Remotes` field and be CRAN compliant. That said to me it seems over the top to install `pandoc` in this action just because `rmarkdown` is a Suggest even if we're not planning to use it, so I'd be happy to manually switch of `pandoc` here to avoid the issue.

jamesmbaazam (approved):
LGTM after the explanation. We just need to update the branch before merge.

PR #588, opened 2024-02-29
fix estimate_secondary without delays

### LINKED ISSUE #587: `estimate_secondary` fails without delays
### Summary:
`estimate_secondary` fails when delays are empty (`delay = delay_opts()`)

### Description:
This might be by design but there is actually no reason why we can't use the model to estimate scaling between two time series when there is no design. Either way we should either throw an informative error or (prefer) handle situations where there is no delay.

### Reproducible Steps:
``` r
library("EpiNow2")
library("data.table")

cases <- as.data.table(example_confirmed)[, primary := confirm]
cases[, scaling := 0.4]
cases[, meanlog := 1.8][, sdlog := 0.5]

# Simulate secondary cases
sec <- simulate_secondary(cases, type = "incidence")
sec[
  ,
  c("confirm", "scaling", "meanlog", "sdlog", "index", "scaled", "conv") :=
    NULL
]
inc <- estimate_secondary(sec, delays = delay_opts())
#> Warning: The meaning of the 'max' argument has changed compared to previous versions. It now indicates the maximum of a distribution rather than the length of the probability mass function (including 0) that it represented previously. To replicate previous behaviour reduce max by 1.
#> This warning is displayed once every 8 hours.
#> Error in eval(expr, envir, enclos) : 
#>   Exception: no more scalars to read (in 'estimate_secondary', line 468, column 2 to column 43)
#> character(0)
#> error occurred during calling the sampler; sampling not done
#> Error in `fit_model_with_nuts()`:
#> ! model fitting was timed out or failed
```

<sup>Created on 2024-02-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### EpiNow2 version

1.4.9000

sbfnk:
## Description

This PR closes #587.

In order to do so some parts of `calculate_secondary` had to be moved into the main model file.

### DIFF
--- a/inst/stan/estimate_secondary.stan
+++ b/inst/stan/estimate_secondary.stan
@@ -40,6 +40,15 @@ transformed parameters {
 
   {
     vector[delay_type_max[delay_id] + 1] delay_rev_pmf;
+    vector[t] scaled;
+    vector[t] convolved = rep_vector(1e-5, t);
+    // scaling of primary reports by fraction observed
+    if (obs_scale) {
+      scaled = scale_obs(primary, obs_scale_sd > 0 ? frac_obs[1] : obs_scale_mean);
+    } else {
+      scaled = primary;
+    }
+
     if (delay_id) {
       delay_rev_pmf = get_delay_rev_pmf(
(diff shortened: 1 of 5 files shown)


seabbs (approved):
LGTM

PR #616, opened 2024-03-15
cmdstanr benchmarking

### LINKED ISSUE #391: Add in model profiling via touchstone
We could include the in model profiling as a `touchstone` benchmark. Would have to do it running key functions from R which is a pain vs the integrated cmdstanr profiling but still might be useful

_Originally posted by @seabbs in https://github.com/epiforecasts/EpiNow2/issues/363#issuecomment-1543834388_

sbfnk:
## Description

If it works (and I don't know if it will) then this PR closes #391.

It doesn't use touchstone (as I couldn't see how this might be integrated) but a custom action and the scripts from #415.

### DIFF
--- a/.github/workflows/stan-model-benchmark.yaml
+++ b/.github/workflows/stan-model-benchmark.yaml
@@ -61,7 +61,7 @@ jobs:
       - name: Post comment
         uses: actions/github-script@v7
         env:
-          BENCHMARK: steps.output.outputs.BENCHMARK
+          BENCHMARK: ${{ steps.output.outputs.BENCHMARK }}
         with:
           script: |
             github.rest.issues.createComment({
--- a/.github/workflows/stan-model-benchmark.yaml
+++ b/.github/workflows/stan-model-benchmark.yaml
@@ -43,7 +43,7 @@ jobs:
         run: |
           mv inst/stan inst/stan_branch
           git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin main
-          git checkout main inst/stan
+          git checkout origin/main inst/stan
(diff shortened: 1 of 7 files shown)


sbfnk:
There might be some tweaking to do but in principle this works now. We might just replace touchstone with this as it's proven a bit unwieldy and we really care about performance of the stan model, not so much of the R code.

@seabbs might be of interest for epinowcast, too.

jamesmbaazam (approved):
This is super! I've gone through it and not seen any issues.

sbfnk:
> This is super! I've gone through it and not seen any issues.
> 
> Could we provide a textual interpretation of the CIs as a column (for example, "slower"/"faster"), similar to touchstone?

I've added this, see latest benchmarking comment.

jamesmbaazam (dismissed):

seabbs:
This is really lovely and agree good to merge. IMO it complements vs replaces touchstone as it doesn't test whole workflows and it doesn't test a fair few model permutations (I believe). 

_This would be a good candidate for making into a generic action others can use I think with a bit of tooling around it to do the benchmarking. I'd like it for `epinowcast` for example._

seabbs:
Ah one actual comment. Maybe we want to only print a comment if a benchmark has changed in order to reduce PR noise or trigger on a / command?

sbfnk:
> Ah one actual comment. Maybe we want to only print a comment if a benchmark has changed in order to reduce PR noise or trigger on a / command?

done (and seems to work)

sbfnk:
> _This would be a good candidate for making into a generic action others can use I think with a bit of tooling around it to do the benchmarking. I'd like it for `epinowcast` for example._

I agree

PR #624, opened 2024-03-22
Add laplace and pathfinder approximate inference

### LINKED ISSUE #403: Optimisation
Stan supports optimisation for point estimates. This might be a nice feature to add (with appropriate caveats returned to the user).

sbfnk:
## Description

This PR closes #403.

It adds laplace and pathfinder approximations when the cmdstanr backend is used. Will need a bit more testing but they both seem more robust than the standard variational inference version. If this is confirmed we could add a note to e.g. one of the vignettes to say that these could be used to get an approximate idea before running exact inference for production.

`fit_model_with_vb` has been re-used and re-named for this but as it was really an internal function I didn't start a deprecation cycle.


seabbs (approved):
LGTM. I think marking them as experimental (as they are) is good enough

PR #632, opened 2024-04-08
Update DESCRIPTION - remove styler

seabbs:
As noted in #631 we have  `{styler}` in the suggests for no good reason. This PR removes it.


sbfnk (approved):
Thanks!

PR #648, opened 2024-05-02
rename argument silent -> strict

### LINKED ISSUE #647: rename `silent` argument
The `silent` argument of `discretise()` is not a good name. It should also default to `FALSE` (i.e., throw an error if we can't discretise).

sbfnk:
## Description

This PR closes #647 .

- [ ] I have not added a news item linked to this PR as it provides only a small update to a previous one.


R/dist_spec.R
```diff
@@ -406,9 +406,10 @@ max.dist_spec <- function(x, ...) {
...
 #' @title Discretise a <dist_spec>
 #' @param x A `<dist_spec>`
-#' @param silent Logical; if `TRUE` then any distribution that can't be
-#'   discretised will be returned as is. If `FALSE` then an error will be
-#'   thrown.
+#' @param strict Logical; If `TRUE` (default) an error will be thrown if a
+#' distribution cannot be discretised (e.g., because no finite maximum has been
+#' specified or parameters are uncertain). If if `FALSE` then any distribution
+#' that cannot be discretised will be returned as is.
 #' @return A `<dist_spec>` where all distributions with constant parameters are
```
jamesmbaazam, line 413: Do we actually need this argument? `strict` only controls whether an error should be thrown or not. It seems more idiomatic to just error if something doesn't work.

jamesmbaazam (dismissed):
I'm approving with some reservations as I'm not convinced we need the argument in the first place 😬 . Approving it because keeping it isn't harmful.

R/dist_spec.R
```diff
@@ -406,9 +406,10 @@ max.dist_spec <- function(x, ...) {
 #' (i.e. those with finite support and constant parameters).
 #' @title Discretise a <dist_spec>
 #' @param x A `<dist_spec>`
-#' @param silent Logical; if `TRUE` then any distribution that can't be
-#'   discretised will be returned as is. If `FALSE` then an error will be
-#'   thrown.
+#' @param strict Logical; If `TRUE` (default) an error will be thrown if a
+#' distribution cannot be discretised (e.g., because no finite maximum has been
+#' specified or parameters are uncertain). If if `FALSE` then any distribution
```
jamesmbaazam, line 411:
```suggestion
#' specified or parameters are uncertain). If `FALSE` then any distribution
```

R/dist_spec.R
sbfnk: My original intention was to remove but then that would basically require reproducing all the code when doing https://github.com/epiforecasts/EpiNow2/blob/4d975edf6098c64f8dbba98bcf4ffb52e29de399/R/create.R#L731

The complication is that we can have composite delay distributions (e.g. the sum of the incubation period and reporting delay), and we might be able to discretise one but not the other. Perhaps the whole thing is overengineered anyway but for now I think we want to give the user the chance to say: discretise whatever you can, and feel free to leave the rest alone.

jamesmbaazam (approved):

PR #667, opened 2024-05-10
advance deprecation cycle

sbfnk:
## Description

Now that 1.5.0 is released this advances all deprecation cycles by one step.

- [X] I have reviewed Checks for this PR and addressed any issues as far as I am able.


R/deprecated.R
```diff
@@ -223,87 +76,18 @@ dist_spec <- function(distribution = c(
                         "lognormal", "normal", "gamma", "fixed", "empty"
                       ),
                       params_mean = numeric(0), params_sd = numeric(0),
-                      mean, sd = 0, mean_sd = 0, sd_sd = 0,
                       max = Inf, pmf = numeric(0), fixed = FALSE) {
 
-  lifecycle::deprecate_warn(
+  lifecycle::deprecate_stop(
```
jamesmbaazam, line 81: Won't this break the distribution interface since it's called internally? Seems like we want to use `deprecate_soft()` except it won't warn users. Seems convoluted to me.

sbfnk: It shouldn't be called any more - `new_dist_spec()` is used internally to construct the objects.

jamesmbaazam (dismissed):

R/deprecated.R
jamesmbaazam: Got it!

jamesmbaazam (dismissed):

jamesmbaazam (approved):
LGTM.

Issue #618, opened 2024-03-19
Return forecasts for easy processing with `scoringutils`


sbfnk:
Is your feature request related to a problem? Please describe:
Forecast evaluation is an important part of broader workflows around forecasting. [scoringutils](https://epiforecasts.io/scoringutils) is an R package for evaluating forecasts. It would be nice to integrate tightly with it by returning forecasts and data in a format that allows for easy processing there.

Describe the solution you'd like:
Functions like `estimate_infections`, `simulate_infections`, `forecast_infections` to return forecast in data that allows for calling of `scoringutils::as_forecast()` without any further processing - either directly or (perhaps more likely) through a function `get_forecast()`

seabbs:
or a custom as.forecast method?

jamesmbaazam:
> or a custom as.forecast method?

I'm in favour of this since it'll live here instead of `scoringutils` (even though we have control over this since they're both developed by us).

jamesmbaazam:
Conditional on #451.

sbfnk:
> being unable to add this integration whilst scoringutils remained a suggest dependency.

Is that because you need to import the generic method?

PR #695, opened 2024-06-13
Add vignette on benchmarking model options

### LINKED ISSUE #629: Vignette on approximations and speedups
Type of issue:
Proposal for a new vignette

Detail:
People often state that estimation in the package is slow (e.g. https://journals.plos.org/digitalhealth/article/figure?id=10.1371/journal.pdig.0000052.t002). This is to some degree a function of the default choice of model and MCMC algorithm. Depending on the use case choosing a different model (e.g. the nonmechanistic model, or a random walk on Rt) or an approximation (VB/laplace/pathfinder) can address this issue. It would be great to write a vignette that outlines these options and discusses implications on estimates.

### LINKED ISSUE #730: Change Getting started documentation to point users to faster models
This issue can be solved after merging the benchmarking vignette in #695.
 
I was curious about how the paper ["Real-time estimation of the epidemic reproduction number: Scoping review of the applications and challenges"](https://journals.plos.org/digitalhealth/article?id=10.1371/journal.pdig.0000052) measured run times of the various R packages and came across this section in the supplementary material where more details are given (page 3):

> Computational speed has been assessed by each author of this study with different computer specifications (see Table B). The main function in each package/tool was taken from provided examples (if available) and wrapped in the system.time() R function to measure the execution time. The main function estimated the reproduction number for each R package/tool except for epicontacts, which estimates the serial interval. We chose the following classifications: <10 seconds = very good, 10 seconds – 5 minutes = good, >5 minutes = poor. The classification allocated to each package was based on the agreement of at least 2 out of the 3 computers. We note that such direct comparisons of the runtimes of the different models may not be fair, as the examples provided by each package which we have used to assess speed vary in terms of the dataset used, model complexity, and dimensionality of the reproduction number to estimate. Nevertheless, we assume that examples will always be relatively simple and therefore their computational speed may be a good overall indicator of speed of reproduction number estimation in general using a given package. 

This has got me thinking about whether we should change our docs to use the quicker models that sacrifice accuracy as that is what users will interact with first (copy & paste to try out) but with a caveat. We can then signpost to the slower but more accurate models for real-world use cases.

I'm also making a note here to raise an issue in EpiEstim to re-assess the speed score in [this table](https://mrc-ide.github.io/EpiEstim/articles/alternative_software.html) using other faster and relatively accurate models with evidence in #695.

jamesmbaazam:
## Description

This PR closes #629.

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive benchmarking vignette comparing model speed and forecast accuracy across different epidemic phases and model configurations.
  * Introduced a new "Model benchmarks" menu item in the documentation navigation for easier access to benchmarking results.

* **Chores**
  * Updated package metadata to suggest the use of the `scoringutils` package.
  * Improved project configuration settings for workspace, history saving, and code formatting.


jamesmbaazam:
The `{rstan}` models are working fine but the `{cmdstanr}` models are giving various errors that I will report on.

sbfnk:
I can't reproduce this - do you need to update `cmdstanr` or `cmdstan` (using `cmdstanr::install_cmdstan()`) possibly?

My versions are:
``` r
devtools::package_info("cmdstanr") |>
  dplyr::filter(package == "cmdstanr")
#>  package  * version    date (UTC) lib source
#>  cmdstanr   0.8.1.9000 2024-06-23 [1] Github (stan-dev/cmdstanr@9878dda)
#> 
#>  [1] /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/library
cmdstanr::cmdstan_version()
#> [1] "2.35.0"
```

<sup>Created on 2024-07-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

vignettes/speedup_options.Rmd.orig
```diff
@@ -0,0 +1,369 @@
...
+  ggplot2::theme_bw() +
+  ggtitle("Reconstructing infections")
+plot(infections_plot)
+```
+
+## Some considerations
+
+### Mechanistic vs non-mechanistic models
+
+- Estimation in _EpiNow2_ using the mechanistic approaches (prior on Rt) is often much slower than the non- mechanistic approach. The mechanistic model is slower because it models intricate details about the processes and mechanisms that drive $Rt$ estimates. The nonmechanistic model, on the other hand, is faster because it does not model these intricate details. If performing a retrospective analysis, the _non-mechanistic model_ (where `rt = NULL`) is better suited and a faster alternative to the mechanistic $Rt$ model.
```
jamesmbaazam, line 361:
```suggestion
- Estimation in _EpiNow2_ using the mechanistic approaches (prior on Rt) is often much slower than the non-mechanistic approach. The mechanistic model is slower because it models intricate details about the processes and mechanisms that drive $Rt$ estimates. On the other hand, the non-mechanistic model (where `rt = NULL`) is faster because it does not model these intricate details. If performing a retrospective analysis, the _non-mechanistic model_ is better suited and a faster alternative to modelling $Rt$.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+``` r
+library(EpiNow2)
+library(scoringutils)
+library(data.table)
+library(rstan)
+library(cmdstanr)
+library(ggplot2)
+```
+
+This vignette outlines available options for speeding up model runtimes in _EpiNow2_. We will compare the models by run time, quantitative, and qualitative performance of fit against the "true" trajectory of the data.
```
jamesmbaazam, line 26:
```suggestion
This vignette explores available forecasting model options in _EpiNow2_ and how they vary by speed and forecast performance. We will compare the models by run time and forecast performance (quantitatively and qualitatively) against the "true" trajectory of the data.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+library(rstan)
+library(cmdstanr)
+library(ggplot2)
+```
+
+This vignette explores available forecasting model options in _EpiNow2_ and how they vary by speed and forecast performance. We will compare the models by run time and forecast performance (quantitatively and qualitatively) against the "true" trajectory of the data.
+
+## The benchmarking data
+
+We will simulate the "true" infections and Rt data using for benchmarking using `forecast_infections()`.
```
jamesmbaazam, line 30:
```suggestion
To compare the model options, we will need a dataset for which we know the "true" values and trajectories. So, we will start by generating the "true" infections and Rt data using _EpiNow2_'s `forecast_infections()` function. 
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+
+This vignette explores available forecasting model options in _EpiNow2_ and how they vary by speed and forecast performance. We will compare the models by run time and forecast performance (quantitatively and qualitatively) against the "true" trajectory of the data.
+
+## The benchmarking data
+
+We will simulate the "true" infections and Rt data using for benchmarking using `forecast_infections()`. 
+
+`forecast_infections()` requires a fitted estimates object from `estimate_infections()` or `epinow()`, the trajectory of the reproduction number, `R`, and the number of samples to simulate. So, we will set these up first.
+
+To obtain the `estimates` object, we have to fit the `estimate_infections()` model to observed data to recover realistic parameter values for the simulation. To do this, we will use the first 100 observations of the `example_confirmed` dataset vendored with this package. This dataset provides a time series of cases by reporting date. We will also use the `example_generation_time`, `example_incubation_period`, and `example_reporting_delay` values that come with the package.
```
jamesmbaazam, line 34:
```suggestion
To obtain the `estimates` object, we will run the `estimate_infections()` model on real-world observed data to recover realistic parameter values. We will use the first 100 observations of the `example_confirmed` data set, the `example_generation_time`, `example_incubation_period`, and `example_reporting_delay` values that come with _EpiNow2_.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+
+## The benchmarking data
+
+We will simulate the "true" infections and Rt data using for benchmarking using `forecast_infections()`. 
+
+`forecast_infections()` requires a fitted estimates object from `estimate_infections()` or `epinow()`, the trajectory of the reproduction number, `R`, and the number of samples to simulate. So, we will set these up first.
+
+To obtain the `estimates` object, we have to fit the `estimate_infections()` model to observed data to recover realistic parameter values for the simulation. To do this, we will use the first 100 observations of the `example_confirmed` dataset vendored with this package. This dataset provides a time series of cases by reporting date. We will also use the `example_generation_time`, `example_incubation_period`, and `example_reporting_delay` values that come with the package. 
+
+Several of the parameters we will use through out this vignette will have the same value, so we will set these up first.
```
jamesmbaazam, line 36:
```suggestion
Several of the parameters we will use through out this vignette will have the same value, so we will set these up in one place.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+# Forecast infections and the trajectory of Rt
+forecast <- forecast_infections(estimates, R = R_noisy, samples = 10)
+```
+
+```
+#> DEBUG [2024-06-13 20:11:46] simulate_infections: Running in exact mode for 1 samples (across 1 chains each with a warm up of 1 iterations each) and 154 time steps of which 40 are a forecast
+```
+
+We will now extract the benchamrking data:
+- `R_true`: the median of the simulated Rts, and
```
jamesmbaazam, line 118:
```suggestion
- `R_true`: the median of the simulated $Rt$, and
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+``` r
+R_true <- forecast$summarised[variable == "R"]$median
+
+# Get the posterior samples from which to extract the simulated infections and reported cases
+posterior_sample <- forecast$samples[sample == 1]
+
+# Extract the simulated infections
+infections_true <- posterior_sample[variable == "infections"]$value
+
+# Extract the simulated reported cases
```
jamesmbaazam, line 131:
```suggestion
# Extract the simulated reported cases and rename the "value" column to "confirm" (to match EpiNow2 requirements)
```

vignettes/speedup_options.Rmd
jamesmbaazam, line 30:
```suggestion
We will simulate the "true" infections and $R_t$ data using for benchmarking using `forecast_infections()`. 
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+reported_cases_true <- posterior_sample[
+  variable == "reported_cases", .(date, confirm = value)
+]
+```
+
+Now, to the main part of this vignette: we will compare the run times of different models in _EpiNow2_.
```
jamesmbaazam, line 137:
```suggestion
Now, to the main part of this vignette: we will compare the run times of different models options.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+
+## Model options
+
+We will use the following models:
+- _gp_: A Gaussian process prior on the reproduction number $R_t$ and no random walk, using _MCMC sampling_.
+- _vb_: A Gaussian process prior on the reproduction number $R_t$ and no random walk, using the _variational inference_ algorithm.
+- _non_mechanistic_: A model with that uses _back calculation_.
```
jamesmbaazam, line 144:
```suggestion
- _non_mechanistic_: A model with that uses _back calculation and sets no prior on $R_t$.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+- _rw1_no_gp_: A model with a 1-day random walk and Gaussian process turned off.
+- _rw7_no_gp_: A model with a 7-day random walk and Gaussian process turned off.
+- _rw7_gp_: A model with a 7-day random walk and Gaussian process turned on.
+- _pf_: A model that uses the "pathfinder" algorithm (from the `{cmdstanr}` package) for sampling.
```
jamesmbaazam, line 148:
```suggestion
- _pf_: A model that uses the "pathfinder" algorithm (from the [`{cmdstanr}`](https://github.com/stan-dev/cmdstanr) package) for sampling.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+- _lp_: A model that uses the "Laplace" algorithm (from the `{cmdstanr}` package) for sampling.
```
jamesmbaazam, line 149:
```suggestion
- _lp_: A model that uses the "Laplace" algorithm (from the [`{cmdstanr}`](https://github.com/stan-dev/cmdstanr) package) for sampling.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+|rw1_no_gp       |  5.972183|
+|rw7_no_gp       |  2.413113|
+|rw7_gp          |  6.252660|
+|vb              |  2.673222|
+
+### Forecast evaluation
+
+Now, we will compare the estimated and true values using the continuous ranked probability score (CRPS). The CRPS is a proper scoring rule that measures the accuracy of probabilistic forecasts. We will use the `crps()` function from the `{scoringutils}` package.
+
+To calculate the CRPS for the estimated Rt and infections, we will first set up a function that makes sure the data and estimates are of the same length and calls the `crps_sample()` function from the `{scoringutils}` package.
```
jamesmbaazam, line 376:
```suggestion
To calculate the CRPS for the estimated $R_t$ and infections, we will first set up a function that makes sure the data and estimates are of the same length and calls the `crps_sample()` function from the `{scoringutils}` package.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+# A function to calculate the CRPS
+calc_crps <- function(x, truth) {
+  shortest_obs_length <- min(ncol(x), length(truth))
+  reduced_truth <- tail(truth, shortest_obs_length)
+  reduced_x <- tail(t(x), shortest_obs_length)
+  return(crps_sample(reduced_truth, reduced_x))
+}
+```
+
+Now, we can calculate CRPS for the Rt and infection estimates of the various models and using `calc_crps()`.
```
jamesmbaazam, line 388:
```suggestion
Now, we can calculate CRPS for the $R_t$ and infection estimates of the various models using `calc_crps()`.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,487 @@
...
+  xlab("Time") +
+  ylab("CRPS") +
+  ggplot2::theme_bw() +
+  ggtitle("Reconstructing infections")
+plot(infections_plot)
+```
+
+![plot of chunk infections_plot](figure/infections_plot-1.png)
+
+## Some considerations
```
jamesmbaazam, line 475:
```suggestion
## Some considerations when using the model options
```

jamesmbaazam:
I have now pushed the vignette with the run of all the models using MCMC ([a472439](https://github.com/epiforecasts/EpiNow2/pull/695/commits/a4724392d323cfc4f40440e1c5215d0649c36456)). 

To do:

- [ ] Run the approximate methods ("vb", "pathfinder", and "laplace")
- [ ] Add a model using "vb" from cmdstanr

jamesmbaazam:
I noticed something interesting about `pathfinder` struggling to fit large case reports and have created a separate issue for it here  #728.

@sbfnk The only thing in the way of this vignette getting merged is the struggle to get `pathfinder` and `laplace` working. If we don't want to delay this any further, we could finalise this version, mention that a future enhancement will include those two methods, and get this merged, then add them as an enhancement when I've figured them out. All the models including `vb` from `{rstan}` and `{cmdstanr}` are working currently. I will precompile them all and push.

jamesmbaazam:
From meeting with Seb today:
- [ ] Finalise current vignette by running all models, leaving out Pathfinder and Laplace for future enhancement.
- [ ] Add more metrics alongside run time:
         - [  ] total CRPS over time
         - [  ] CRPS at the last time point (real-time performance)
         - [  ] Rank over time

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,458 @@
...
+    "laplace"
+  ),
+  description = c(
+    "The model with no priors on Rt",
+    "Gaussian process turned on and with a non-stationary prior on $R_t$ (default)",
+    "Gaussian process turned on and with a stationary Gaussian process prior on $R_t$",
+    "1-day random walk, Gaussian process turned off, and a non-stationary prior on $R_t$ (default)",
+    "1-day random walk, Gaussian process turned on, and a non-stationary prior on $R_t$ (default)",
+    "7-day random walk, Gaussian process turned off, and a non-stationary prior on $R_t$ (default)",
+    "7-day random walk, Gaussian process turned on, and a non-stationary prior on $R_t$ (default)",
```
seabbs, line 165: I think it might be useful to describe the models using categories vs free text (so process model changes, Rt model changes, fitting method changes, fitting package changes) if possible?

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,458 @@
...
+  labs(x = "Time", y = "CRPS", title = "Estimating infections with various model options") +
+  ggplot2::theme_bw()
+plot(infections_plot)
+```
+
+## Some considerations when using the model options
+
+### Mechanistic vs non-mechanistic models
+
+- Estimation in _EpiNow2_ using the mechanistic approaches (prior on $R_t$) is often much slower than the non-mechanistic approach. The mechanistic model is slower because it models intricate details about the processes and mechanisms that drive $R_t$ estimates using the renewal equation. The non-mechanistic model, on the other hand, runs much faster but does not use the renewal equation to generate infections. Because of this none of the options defining the behaviour of the reproduction number are available in this case, limiting flexibility. It also means that the model is questionable for forecasting. If performing a retrospective analysis, the _non-mechanistic model_ (where `rt = NULL`) is better suited and a faster alternative to the mechanistic $R_t$ model.
```
seabbs, line 450: I think there are some quite strong statements in here. It might be helpful to discuss the evidence on which they rest. I am particularly uncomfortable with the non-mechanistic recommendation for retrospective use (where I would probably personally suggest the gp on Rt vs the residual approach right now without evidence).

seabbs, line 450:
```suggestion
- Estimation in _EpiNow2_ using the mechanistic approaches (prior on $R_t$) is often much slower than the non-mechanistic approach. The mechanistic model is slower because it models details about the processes and mechanisms that drive $R_t$ estimates using the renewal equation. The non-mechanistic model, on the other hand, runs much faster but does not use the renewal equation to generate infections. Because of this none of the options defining the behaviour of the reproduction number are available in this case, limiting flexibility. It also means that the model is questionable for forecasting. If performing a retrospective analysis, the _non-mechanistic model_ (where `rt = NULL`) is better suited and a faster alternative to the mechanistic $R_t$ model.
```

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,458 @@
...
+infections_crps <- lapply(
+  infections_estimated,
+  calc_crps,
+  truth = infections_true
+)
+```
+
+We will now post-process the CRPS results to make them easier to visualise by adding a  "time" column and reshaping the data to long format.
+
+```{r fit_postprocessing}
```
seabbs, line 404: I'd like to see scores summarised by period (with evidence, partial evidence, and forecast) in order more easily evidence the recommendations below.

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,458 @@
...
+  obs = obs,
+  horizon = horizon,
+  output = "fit"
+)
+```
+
+For the `R` data, we will set up an arbitrary trajectory and add some Gaussian noise.
+```{r R-data}
+# Arbitrary reproduction number trajectory
+R <- c(
```
seabbs, line 102: as this test R is non-stationary and drawn from model different data generating processes (i.e stepwise changes, linear changes etc) I think it somewhat biases you towards less mechanistic model options (more because there is less room for mechanistic choices to do well as none of the latent process models are well defined to fit to this data). Probably should just note this complexity.

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,458 @@
...
+  infections_df,
+  id.vars = c("time"),
+  variable.name = "model"
+)
+```
+
+Let's visualise the CRPS results for the $R_t$ and infection estimates.
+
+```{r rt_plot}
+rt_plot <- ggplot(rt_df, aes(x = time, y = value, colour = model)) +
```
seabbs, line 427: Maybe all of these should be relative to the default mode (as this is what people are using unless they do something else) and stratified by type (process model change or sampler change)

vignettes/speedup_options.Rmd
jamesmbaazam: Not sure I understand. Can you give an example by rewording one of the examples above?

vignettes/speedup_options.Rmd
jamesmbaazam: What sort of evidence would be useful here aside the CRPS?

jamesmbaazam: Thanks

vignettes/speedup_options.Rmd
jamesmbaazam: Thanks. I've added a caveat here bc56702e03fe2bf850ed30966e731bde94bb416e.

vignettes/speedup_options.Rmd
jamesmbaazam: Please see the latest version in 495881519f2368faed9080eda7532640e4d20d14 and let me know what you think.

vignettes/speedup_options.Rmd
jamesmbaazam: The vignette has now been extensively rewritten. Please see the latest version in 495881519f2368faed9080eda7532640e4d20d14 and let me know what you think

vignettes/speedup_options.Rmd
jamesmbaazam: These have now been included. Please see the latest version in 495881519f2368faed9080eda7532640e4d20d14 and let me know what you think.

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,710 @@
...
+  rt = rt_opts(prior = rt_prior_default, rw = 14),
+  gp = NULL,
+  obs = obs,
+  horizon = 0, # no forecasting
+  output = "fit"
+)
+#> DEBUG [2024-08-07 14:41:13] epinow: Running in exact mode for 2000 samples (across 4 chains each with a warm up of 250 iterations each) and 114 time steps of which 0 are a forecast
+```
+
+For the `R` data, we will set up an arbitrary trajectory and add some Gaussian noise.
```
seabbs, line 101: Suggest expanding on why this Rt trajectory is the way it is. I.e constant periods, periods of linear change and what that might look like in the real world

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,710 @@
...
+#> DEBUG [2024-08-07 14:41:54] simulate_infections: Running in exact mode for 1 samples (across 1 chains each with a warm up of 1 iterations each) and 154 time steps of which 40 are a forecast
+```
+
+We will now extract the required data:
+- `R_true`: the median of the simulated $R_t$ values,
+- `infections_true`: the infections by date of infection, and
+- `reported_cases_true`: the reported cases by date of report.
+
+``` r
+R_true <- forecast$summarised[variable == "R"]$median
```
seabbs, line 131: why is R_true not just R?

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,710 @@
...
+## Model options
+
+Below we describe each model.
+
+
+Table: Model options
+
+|model               |description                                                                                                            |
+|:-------------------|:----------------------------------------------------------------------------------------------------------------------|
+|default_mcmc_rstan  |Default model; Gaussian process on new infection and with a non-stationary prior on $R_t$                              |
```
seabbs, line 156: The default only has a GP on the residuals of Rt (i.e this is what makes its Rt estimates non-stationary). Suggest a reword here as it sounds like its on new infections.

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,710 @@
...
+
+|model               |gaussian_process |rt_model_rw |rt_model_gp    |sampling              |package  |
+|:-------------------|:----------------|:-----------|:--------------|:---------------------|:--------|
+|default_mcmc_rstan  |on               |None        |non_stationary |MCMC                  |rstan    |
+|non_mechanistic     |on               |None        |None           |MCMC                  |rstan    |
+|no_rw_gp            |on               |None        |stationary     |MCMC                  |rstan    |
+|rw1_no_gp           |off              |1-day       |non_stationary |MCMC                  |rstan    |
+|rw1_gp              |on               |1-day       |non_stationary |MCMC                  |rstan    |
+|rw7_no_gp           |off              |7-days      |non_stationary |MCMC                  |rstan    |
+|rw7_gp              |on               |7-days      |non_stationary |MCMC                  |rstan    |
```
seabbs, line 180: is this suggesting having both a random walk on Rt and a GP? I think that might be very hard to fit?

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,710 @@
...
+knitr::kable(runtimes_dt[, c("model", "description", "runtime")], caption = "Run times for various _EpiNow2_ models")
+```
+
+
+
+Table: Run times for various _EpiNow2_ models
+
+|model               |description                                                                                                            |                  runtime|
+|:-------------------|:----------------------------------------------------------------------------------------------------------------------|------------------------:|
+|default_vb_rstan    |Default model; Variational algorithm for sampling instead of MCMC and using [`{rstan}`]                                |                     3.7s|
```
seabbs, line 343: As I said in a previous review round I think its useful to distinguish in the output between model differences and fitting approaches differences. What do you think about that?

vignettes/speedup_options.Rmd
```diff
@@ -0,0 +1,710 @@
...
+    digits = 2
+  )
+)
+```
+
+
+
+Table: Run times and $R_t$ estimation/forecast performance - measured by total CRPS - of various _EpiNow2_ models (ordered by run times)
+
+|model               |                  runtime| estimate| estimate based on partial data| forecast|description                                                                                                            |
```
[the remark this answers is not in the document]
seabbs: Personally I would try and pick an estimation time that makes this clear and discuss this.

seabbs (commented):
I have some fairly similar comments to last time and a few more. Shall we think about organising a meeting to chat through some of this? 

I am wondering if we can really get away with a single example like this and instead maybe we need to use rolling fits (i.e a lot of computation) and just a single fit as a example). If we did that we could say things with more certainty but it would be a much bigger piece of work. 

(I know lots of critical comments here but this is a nice piece of work and is definitely going in the right direction).

seabbs (commented):
I have some fairly similar comments to last time and a few more. Shall we think about organising a meeting to chat through some of this? 

I am wondering if we can really get away with a single example like this and instead maybe we need to use rolling fits (i.e a lot of computation) and just a single fit as a example). If we did that we could say things with more certainty but it would be a much bigger piece of work. 

(I know lots of critical comments here but this is a nice piece of work and is definitely going in the right direction).

vignettes/speedup_options.Rmd
jamesmbaazam: As in the naming?

jamesmbaazam:
### Update 

I met with @seabbs yesterday and decided to pivot this vignette to explore 3 models (Default model, Non-residual Rt model (gp_on = “R0”), 7-day RW Rt model) in terms of fitting methods, i.e., approximate vs exact (MCMC, pathfinder, laplace, and variational bayes). The justification is that users will often follow a decision tree to decide which models to use: (i) use case: retrospective vs real-time (i) compute limitations.

### Outline

- Define benchmarking data and models
- Fit at different time points
- Evaluate run times and forecast quality (bias and coverage) for complete and partial data and forecasts (do the models show consistent behavior?) stratified by fitting method
- Provide a table of recommendations for which models to use for what

@seabbs Is this accurate?

jamesmbaazam:
I'm picking this up again and will proceed with outline in https://github.com/epiforecasts/EpiNow2/pull/695#issuecomment-2286498896.

jamesmbaazam:
I'll be picking up this work this week after a long break.

sbfnk:
The plan as stated above sounds really good - just a couple of comments:

> explore 3 models (Default model, Non-residual Rt model (gp_on = “R0”), 7-day RW Rt model) 

I'd suggest to add the nonmechanistic model.

> Evaluate run times and forecast quality (bias and coverage) for complete and partial data and forecasts (do the models show consistent behavior?) stratified by fitting method

With this can we really get at the stated intention of highlighting differences between real-time vs. retrospective use? It would make sense to me to do this on Rt and across the whole "estimate" part of the time series (for retrospective) vs. partial data (for real-time) but ignore forecasts as that introduces additional complexity of which model to use (latest vs. predicted Rt).

> Provide a table of recommendations for which models to use for what

This would be a fantastic outcome!

jamesmbaazam:
Haven't pushed to this PR in a while because I was away. I have some local changes that I'll be pushing soon. Current main challenge is with some stochastic failures when using the approximate methods. If they continue, we may have to revise the scope again.

jamesmbaazam:
I am away until Jan 3 and will reboot this vignette on my return.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,988 @@
...
+
+# Wrapper for extracting the results and making them into a data.table
+get_model_results <- function(results_by_snapshot, variable) {
+  # Get model results list
+  lapply(
+    results_by_snapshot,
+    function(model_results) {
+      lapply(model_results, extract_results, variable)
+    }
+  )
```
jamesmbaazam, line 458: Simplify these nested `lapply()` calls with [purrr::map_depth()](https://purrr.tidyverse.org/reference/map_depth.html).

kaitejohnson:
@jamesmbaazam Would it be possible to share a rendered html, since I don't think we've set anything up yet to post a preview of the articles as a comment (though we could)

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  delays = delay_opts(delay),
+  obs = obs,
+  forecast = forecast_opts(horizon = horizon),
+  verbose = FALSE
+)
+```
+
+Additionally, from the benchmarking data, we choose the following dates to represent the periods of growth, peak, and decline.
+```{r snapshot_dates}
+snapshot_dates <- as.Date(c("2020-05-15", "2020-06-21", "2020-06-28"))
```
sbfnk, line 360: Might be good to show a plot of R with these highlighted with vertical lines - looking at that plot I think I would have located these dates elsewhere (e.g. the "decline" one is only just starting to decline).

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  if (!inherits(estimates, c("matrix"))) return(rep(NA_real_, length(truth)))
+  # Assumes that the estimates object is structured with the samples as rows
+  shortest_obs_length <- min(ncol(estimates), length(truth))
+  reduced_truth <- tail(truth, shortest_obs_length)
+  estimates_transposed <- t(estimates) # transpose to have samples as columns
+  reduced_estimates <- tail(estimates_transposed, shortest_obs_length)
+  return(
+    crps_sample(
+      log10(reduced_truth),
+      log10(reduced_estimates)
```
sbfnk, line 506: only works if not never zero (so perhaps fine if it works)

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+calc_crps <- function(estimates, truth) {
+  # if the object is not a matrix, then it's an NA (failed run)
+  if (!inherits(estimates, c("matrix"))) return(rep(NA_real_, length(truth)))
+  # Assumes that the estimates object is structured with the samples as rows
+  shortest_obs_length <- min(ncol(estimates), length(truth))
+  reduced_truth <- tail(truth, shortest_obs_length)
+  estimates_transposed <- t(estimates) # transpose to have samples as columns
+  reduced_estimates <- tail(estimates_transposed, shortest_obs_length)
+  return(
+    crps_sample(
```
sbfnk, line 504: if you're only using `crps_sample()` from `scoringutils` would it make sense to use `scoringRules` instead? As the `scoringutils` version is just a thin wrapper.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+
+### Evaluating model performance
+
+Now, we will evaluate the performance of the models using the [continuous ranked probability score (CRPS)](https://en.wikipedia.org/wiki/Scoring_rule#Continuous_ranked_probability_score). The CRPS is a [proper scoring rule](https://en.wikipedia.org/wiki/Scoring_rule#Propriety_and_consistency) that measures the accuracy of probabilistic forecasts. When comparing models, the smaller the CRPS, the better.
+
+We will focus on evaluating the results of fitting with MCMC. We will also show the results for the approximate sampling methods but will not delve deeply into them as we do not recommended for full inference due to their experimental nature currently, which makes them unstable. In a latter section, we will provide ideas on how the approximate methods can be used in `{EpiNow2}` workflows.
+
+```{r process-crps}
+# Process CRPS for Rt
+rt_crps <- process_crps(results, "R", R)
```
sbfnk, line 714: I wonder if it would be good to look at the estimate on the day of the inference for a real-time measure and then e.g. the 7-day forecast rather than the whole "estimate from partial data" / "forecast" time series. Debatable perhaps but if you use the whole series I think it would make more sense to use the multivariate CRPS / energy score (which is not yet implemented in `scoringutils` but is available in `scoringRules::es_sample()`.

sbfnk (commented):
Wow, this is a lot of great work!

General comment is that it has elements of a vignette and elements of a paper so it's not quite clear where it stands on the distinction between the two. I would say for a vignette it's a bit too complex defining functions calling other functions etc. - for a paper it definitely has some great new content but the style is more like a walkthrough/vignette.

With that in mind I have focused on code correctness / implementation in my review so far. We can discuss the text later perhaps.

Anyway, very cool stuff and really interesting!

sbfnk:
I think we can push this to the next release to give us a bit more time to review content and scope - no need to rush it into 1.7.0 (which is otherwise pretty much ready to go).

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+    rt = NULL,
+    stan = stan_opts(method = "pathfinder", backend = "cmdstanr")
+  ),
+  # The non-mechanistic model with laplace fitting
+  non_mechanistic_laplace = list(
+    rt = NULL,
+    stan = stan_opts(method = "laplace", backend = "cmdstanr")
+  ),
+  # The 7-day RW Rt model with MCMC fitting
+  rw7_mcmc = list(
```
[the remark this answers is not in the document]
seabbs: I thought this was a model you were investigating at on point?

kaitejohnson (commented):
@jamesmbaazam Mostly a lot of questions, feel free to ignore any of these as I am sure you have already discussed a lot of it.

Overall this is a really nice analysis. From an outside perspective, I am really interested in how the approximate methods compare in terms of forecast performance to MCMC. Also, it wasn't clear that these were evaluated only on the out-of-sample data, and if not, separately evaluating the out-of-sample forecast performance seems important (but I may have missed this).

vignettes/benchmarks.Rmd.orig
seabbs: dude

seabbs: (how are we going to goose downloads with that approach ;) )

seabbs: More seriously I think this indicates we should just use more of the features of scoringutils

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+```{r share_inputs}
+# Observation model options
+obs <- obs_opts(
+  scale = Normal(0.1, 0.025),
+  return_likelihood = TRUE
+)
+# Rt prior
+rt_prior_default <- Normal(2, 0.1)
+# Number of cores
+options(mc.cores = 6)
```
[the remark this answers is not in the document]
seabbs: or do some kind of core detection probably.

vignettes/benchmarks.Rmd.orig
sbfnk: Yep, ultimately related also to https://github.com/epiforecasts/EpiNow2/issues/618

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+
+Let's proceed to simulate the true infections and $R_t$ data by sampling from $10$ posterior samples.
+```{r true-data}
+# Forecast infections and the trajectory of Rt
+forecast <- forecast_infections(
+  estimates$estimates,
+  R = R_noisy,
+  samples = 10
+)
+```
```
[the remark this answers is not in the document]
seabbs: Aha we are basically being extremely clever and pragmatic but I agree we should tell people this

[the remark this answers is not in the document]
seabbs: The idea is that by using a the posterior for everything but the Rt trajectory you are getting a simulation that very accurately reflects a real world setting whilst still being able to know and change the true Rt trajectory

vignettes/benchmarks.Rmd
```diff
@@ -0,0 +1,967 @@
...
+
+## Considerations for choosing an appropriate model
+
+### Model types (Semi-mechanistic vs non-mechanistic)
+
+Estimation in `{EpiNow2}` using the semi-mechanistic approaches (putting a prior on $R_t$) is often much slower than the non-mechanistic approach. The mechanistic model is slower because it models aspects of the processes and mechanisms that drive $R_t$ estimates using the renewal equation. The non-mechanistic model, on the other hand, runs much faster but does not use the renewal equation to generate infections. Because of this none of the options defining the behaviour of the reproduction number are available in this case, limiting its flexibility.
+
+It's worth noting that the non-mechanistic model in `{EpiNow2}` is equivalent to that used in the [`{EpiEstim}`](https://mrc-ide.github.io/EpiEstim/index.html) R package as they both estimate $R_t$ from case time series and the generation interval distribution.
+
```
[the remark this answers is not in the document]
seabbs: Could you clarify a bit what you mean bit by this there? The renewal?  

> Also I am guessing the difference is that EpiEstim would require specifying just the mean delay, where the non-mechanistic EpiNow2 is able to incorporate the full delay distribution, which is likely to be important for fitting to more delayed data

It doesn't take any delays at all and the approach to smoothing is different (here it is being fit). We should probably talk about this somewhere.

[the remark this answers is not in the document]
kaitejohnson: My understanding of EpiEstim is that it does a direct computation of the posterior estimate of R(t) based on the cases and the GI, assuming cases are just latent incident infections shifted by some mean delay. 

So when its written that the non-mechanistic one is like EpiEstim, I assume it does something similar, but since the input has the user specify a full delay distribution, I (perhaps incorrectly) assumed it is using that full distribution (rather than just the mean).

What do you mean it doesn't take any delays?

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
sbfnk: Me? Possibly but I don't think I would recommend it - given all the existing identifiability issues I don't think we want to layer on more flexibility.

vignettes/benchmarks.Rmd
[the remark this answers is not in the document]
seabbs: EpiEstim doesnt do anything with delays and uses the serial interval not the GT. I think what the text means here is that underlying infection generating process Epinow2 and EpiEstim uses are the same but not the rest of the model and that is infact only partially true because of differences in the latent model and differences in the uncertainty. Personally I would just nuke this paragraphf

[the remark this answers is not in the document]
kaitejohnson: Ok yeah I don't think I understand how the non-mechanistic model works then!

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+library(purrr)
+library(lubridate)
+library(scales)
+library(posterior)
+```
+
+In using `{EpiNow2}`, users will often be faced with two decision points that will guide their choice of an appropriate model: (i) use case: retrospective vs real-time analysis, and (ii) limited computing resources. `{EpiNow2}` provides a range of customisations of the default model to suit these decision points.
+
+The aim of this vignette is to show how these model customisations affect the speed and accuracy of the model. We will benchmark four (4) `{EpiNow2}` model options chosen to cover a range of use cases. We will benchmark the default model, the model with the 7-day random walk prior on $R_t$, the non-mechanistic model, which has no explicit prior on $R_t$, and the non-residual $R_t$ model, which assumes a stationary prior on $R_t$.
+
```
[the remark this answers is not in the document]
jamesmbaazam: Good catch. See f31db297d6e264959ca745918563fcf33ea7ddc2.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+We will analyse the estimation and forecasting performance of these models when solved with MCMC sampling. `{EpiNow2}` also provides approximate sampling methods to solve these models ([variational inference](https://mc-stan.org/docs/cmdstan-guide/variational_config.html), [pathfinder method](https://mc-stan.org/docs/cmdstan-guide/pathfinder_config.html), and [laplace sampling](https://mc-stan.org/docs/cmdstan-guide/laplace_sample_config.html)). These algorithms are, however, not recommended for use in pipelines, so we will not emphasise their use. We will, however, highlight how the models perform when solved with these approximate sampling methods and provide an overview of when they may be appropriate.
+
```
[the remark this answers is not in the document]
jamesmbaazam: > Perhaps use language "production pipelines" , and explain that they can be useful for developing and testing, 

I do this later in the vignette in the considerations sections.

> and where things might go wrong e.g. the variance

I think it might be too much detail for this vignette but happy to be convinced :).

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: This is addressed in the considerations section but happy to repeat it here.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: > Nitpicky and feel free to ignore but MCMC is also an approximation of the posterior, so might want to rephrase slightly.

See c44ae64e41d84d81c463fd69813111c778656363.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  samples = 10
+)
+```
+
+Let's now extract the true data for benchmarking:
+- `R_true`: the median of the simulated $R_t$ values,
+- `infections_true`: the infections by date of infection, and
+- `reported_cases_true`: the reported cases by date of report.
+```{r extract-true-data}
+R_true <- forecast$summarised[variable == "R"]$median
```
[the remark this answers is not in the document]
jamesmbaazam: I will update but just to note that this is [from the synthetic evaluation scripts](https://github.com/epiforecasts/EpiNow2/blob/main/inst/dev/recover-synthetic/rt.R#L30-L32) so might have to update there too?

Also, just to understand, are some of the methods/models more sensitive to the number of samples?

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+
+We will now proceed to define and run the different model options.
+
+## Models
+
+### Descriptions
+
+Below we describe each model.
+```{r model-descriptions,echo = FALSE}
+model_descriptions <- dplyr::tribble(
```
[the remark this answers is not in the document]
jamesmbaazam: Awesome. Thanks.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: Now off in 36c9885ce1dcb1d3aec8f6391209c46a1aa48300.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  ),
+  # The non-mechanistic model with laplace fitting
+  non_mechanistic_laplace = list(
+    rt = NULL,
+    stan = stan_opts(method = "laplace", backend = "cmdstanr")
+  ),
+  # The 7-day RW Rt model with MCMC fitting
+  rw7_mcmc = list(
+    rt = rt_opts(
+      prior = rt_prior_default,
```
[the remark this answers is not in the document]
jamesmbaazam: Fixed in 3bcb7e28fc7e8f8a8be44b93fc7e022d86f93e27.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  meanlog = 0.5,
+  sdlog = 0.5,
+  max = 10
+)
+
+# Combine the incubation period and reporting delay into one delay
+delay <- incubation_period + reporting_delay
+
+# 7-day forecast window
+horizon <- 7
```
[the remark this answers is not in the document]
jamesmbaazam: Would you suggest longer? I'll change to 14 but feel free to chip in.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+    ),
+    stan = stan_opts(method = "pathfinder", backend = "cmdstanr")
+  )
+)
+```
+
+## Running the models
+
+All the models will share the configuration for the generation time, incubation period, reporting delay, and the forecast horizon, so we will define them once and pass them to the models.
+
```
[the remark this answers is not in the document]
jamesmbaazam: You're right. They were redefined here but no need to. See f5e3c490ff53be65c3546a4de180e4643132504e.

vignettes/benchmarks.Rmd
```diff
@@ -0,0 +1,967 @@
...
+rt_crps_approx <- rt_crps_dt_final[fitting != "mcmc"]
+
+# Process plot data and plot
+rt_tv_crps_plot_approx <- plot_crps_over_time(rt_crps_approx, "Time-varying model performance in estimating Rt (approximate methods)")
+rt_tv_crps_plot_approx  +
+    facet_wrap(fitting~epidemic_phase)
+```
+
+![plot of chunk plot-rt-tv-crps-approx](benchmarks-plot-rt-tv-crps-approx-1.png)
+
```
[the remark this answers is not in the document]
seabbs: it would but I think this might be out of scope here (i.e its a new research/doc issue IMO)

vignettes/benchmarks.Rmd
```diff
@@ -0,0 +1,967 @@
...
+# Process data and plot
+rt_total_crps_mcmc_plot <- plot_total_crps(rt_total_crps_mcmc, "Aggregated model performance in estimating Rt (mcmc)")
+rt_total_crps_mcmc_plot  +
+  scale_shape_manual(values = c(0, 1, 17, 8)) +
+  facet_wrap(~type)
+```
+
+![plot of chunk crps-plotting-rt-total](benchmarks-crps-plotting-rt-total-1.png)
+
```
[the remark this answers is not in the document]
seabbs: or a point plot I am irationally anti bar

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
seabbs: I think if we can avoid hard coding this it would be good.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
kaitejohnson: Yeah my thinking is just a bit longer to be able to elucidate some of the differences in forecast performance more clearly, as 7 days is a relatively "easy" forecasting problem

vignettes/benchmarks.Rmd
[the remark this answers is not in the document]
jamesmbaazam: To avoid confusion, I've remove the statement (2e9e7ef71854750aa66ffa46f980c61be287b1dc).

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  "rw7_laplace",  "rw7", "7-day random walk prior on $R_t$; fitting with laplace approximation",
+  "non_residual_mcmc",  "non_residual",  "Stationary prior on $R_t$; fitting with mcmc",
+  "non_residual_vb",  "non_residual", "Stationary prior on $R_t$; fitting with variational bayes",
+  "non_residual_pathfinder",  "non_residual",  "Stationary prior on $R_t$; fitting with pathfinder algorithm",
+  "non_residual_laplace",   "non_residual", "Stationary prior on $R_t$; fitting with laplace algorithm"
+)
+
+knitr::kable(model_descriptions, caption = "Model options")
+```
+
```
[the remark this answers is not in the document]
jamesmbaazam: It was more of a space issue. The table becomes too big to show. So, the split here is between required computational changes and model definitions/details. Happy to hear other ways to combine this.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: We (me & Seb) agreed to use 7 days (2b8db1c295689d9eb904feddf8b76b79f18bd2fd).

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+# Create a version of epinow() that works like base::try() and works even if some models fail.
+safe_epinow <- purrr::safely(epinow)
+# Run the models over the different dates
+results <- lapply(
+  data_snaps, function(data) {
+    lapply(
+      model_configs,
+      function(model) {
+        # Use subset of the data
+        data_to_fit <- data[1:(.N - model_inputs$forecast$horizon)] # exclude the forecast horizon
```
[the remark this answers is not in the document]
jamesmbaazam: > Why do you subtract the forecast horizon here?

The idea was to train the model on the data excluding the horizon and then evaluate on the full time series including the horizon (i.e., out-of-sample evaluation).

vignettes/benchmarks.Rmd.orig
jamesmbaazam: Addressed in 09d46e076aed7dccd05d5a596a39cc1a4537b8f2.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: I now fold the model components part in a summary div. Let's see if that works 09d46e076aed7dccd05d5a596a39cc1a4537b8f2.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: This has been added in 09d46e076aed7dccd05d5a596a39cc1a4537b8f2.

seabbs:
Before I look at this @jamesmbaazam could you give me something on what the summary of the last review was and what you have done on it? I don't want to reopen a bunch of closed topics etc.

seabbs (commented):
I will give this a more through read but my high level concern of treating the fitting method as something that can be tuned for forecast performance remains. 

What I mean by this is that all of the non-MCMC approaches are trying to emulate the MCMC approach. Success for them is doing so even if the the MCMC approach doesn't recover the parameters in this setting I think. 

This is because otherwise you end up in situations where a non-MCMC approach scores better because its bias vs MCMC happens to lead to a better forecast in a certain setting i.e because the model is overly uncertain for example. The message for users here is to tune the model not to change fitting approach.

I think a much safer thing to do is compare the models as you do at the moment but then compare the other fitting methods to the MCMC approaches separately (i.e how close are they to being the same across settings etc). 

You can kind of do this from yor figures but only if you are clear that is the aim and I think it is easy to look at it from other perspectives that as I say above I think are flawed.

All of this being said the text on the fitting approaches is very clear. However, there are no references back to the figures so its a little unclear what people are meant to do with them? 

I feel like this has been discussed so apologies if I have forgotten what the push back on the above was.

vignettes/speedup_options.Rmd
jamesmbaazam: This has been changed to a new trajectory.

vignettes/benchmarks.Rmd.orig
jamesmbaazam: This has been implemented.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: This has been implemented. I take 1 sample.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  "non_residual_pathfinder",  "non_residual",  "Stationary prior on $R_t$; fitting with pathfinder algorithm",
+  "non_residual_laplace",   "non_residual", "Stationary prior on $R_t$; fitting with laplace algorithm"
+)
+
+knitr::kable(model_descriptions, caption = "Model options")
+```
+
+These are the components of each model.
+```{r model-components,echo = FALSE}
+model_components <- dplyr::tribble(
```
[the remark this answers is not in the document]
jamesmbaazam: This has been improved in the latest version.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: The current version now makes them all the same length.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  data[]
+}
+
+# Function to calculate the CRPS of log-transformed values
+calc_crps <- function(estimates, truth) {
+  # if the object is not a matrix, then it's an NA (failed run)
+  if (!inherits(estimates, c("matrix"))) return(rep(NA_real_, length(truth)))
+  # Assumes that the estimates object is structured with the samples as rows
+  shortest_obs_length <- min(ncol(estimates), length(truth))
+  reduced_truth <- tail(truth, shortest_obs_length)
```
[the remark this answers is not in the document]
jamesmbaazam: The current version now uses the `head`.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+  estimates_transposed <- t(estimates) # transpose to have samples as columns
+  reduced_estimates <- tail(estimates_transposed, shortest_obs_length)
```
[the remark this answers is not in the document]
jamesmbaazam: Current version now uses `head`.

vignettes/benchmarks.Rmd.orig
jamesmbaazam: this has been removed.

vignettes/benchmarks.Rmd.orig
jamesmbaazam: Stuck with using scoringutils.

vignettes/benchmarks.Rmd.orig
jamesmbaazam: Current version how computes the total CRPS over the forecast horizon, the Rt estimate on the day of the inference for a real-time measure, and then the 7-day forecast of infections.

vignettes/benchmarks.Rmd.orig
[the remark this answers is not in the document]
jamesmbaazam: The current version does core detection with the `parallel` package.

vignettes/benchmarks.Rmd.orig
```diff
@@ -0,0 +1,914 @@
...
+```
+
+```{r plot-infections-crps-mcmc}
+# Plot CRPS over time for infections
+infections_crps_mcmc_dt <- infections_crps_dt_final[fitting == "mcmc"]
+infections_crps_mcmc_plot <- plot_crps_over_time(infections_crps_mcmc_dt, "Time-varying model performance in estimating infections (mcmc)")
+infections_crps_mcmc_plot +
+    facet_wrap(~epidemic_phase, ncol = 1)
+```
+
```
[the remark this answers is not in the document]
jamesmbaazam: The current version now only evaluates out-of-sample performance.

vignettes/benchmarks.Rmd
```diff
@@ -0,0 +1,821 @@
...
+      }
+    } else {
+        return(subset_draws(obj_mat, variable))
+      }
+    }
+}
+
+# Apply `extract_results()` to a nested list of model runs per snapshot date.
+get_model_results <- function(results_by_snapshot, variable) {
+  # Get model results list
```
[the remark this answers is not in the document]
jamesmbaazam: Fixed in [5edd05d](https://github.com/epiforecasts/EpiNow2/pull/695/commits/5edd05dfe5396b4c7e4235d982ff8a5a0565e9bd).

vignettes/benchmarks.Rmd
```diff
@@ -0,0 +1,821 @@
...
+``` r
+# Function to extract the "timing", "Rt", "infections", and "reports" variables from an
+# epinow() run. It expects a model run, x, which contains a "results" or "error" component.
+# If the model run successfully, "error" should be NULL.
+extract_results <- function(x, variable) {
+  stopifnot(
+    "variable must be one of c(\"timing\", \"R\", \"infections\", \"reports\")" =
+      variable %in% c("timing", "R", "infections", "reports")
+  )
+  # Return NA if there's an error
```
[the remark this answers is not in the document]
jamesmbaazam: Fixed in [5edd05d](https://github.com/epiforecasts/EpiNow2/pull/695/commits/5edd05dfe5396b4c7e4235d982ff8a5a0565e9bd).

vignettes/benchmarks.Rmd
```diff
@@ -0,0 +1,691 @@
...
+
+The aim of this vignette is to show the trade-offs between select model customisations in terms of model speed/run times and nowcasting and real-time forecasting. We will explore four (4) `{EpiNow2}` model options, including the default model. The models, chosen to cover typical use cases, are customisations of the default prior on how $R_t$ is generated over time.
+
+We will evaluate how well the models perform when fitted with the [MCMC sampling algorithm](https://mc-stan.org/docs/reference-manual/mcmc.html) in stan because MCMC is the state-of-the-art algorithm for fitting these kinds of models.
+
+## Data
+
+To compare the models, we will simulate an epidemic with waves capturing the growth, peak, and decline phase. We will then extract subsets of the data capturing the three phases for use as scenarios. All the models will be fit to the three phases and evaluated.
+
+Throughout this vignette, several argument values, including the observation model options and the $R_t$ model prior will be reused, so we will define them here. Note that we use 7 cores out of 8 cores for parallelisation.
```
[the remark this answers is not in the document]
jamesmbaazam: This .Rmd file is rendered from the Rmd.orig so not technically hardcoded.

jamesmbaazam:
I've given this a thought and I think a rolling window evaluation will be more robust than the current data scenarios (growth, peak, decline) approach.  Basically, each model will be fit to rolling windows of 14 days, using 10 weeks for training and 2 weeks for testing/evaluation. 

The results can be evaluated against a constant baseline forecast and assessed using the relative skill, i.e. 

$$1- \dfrac{WIS_{model}}{WIS_{default model}}$$.

sbfnk:
> I've given this a thought and I think a rolling window evaluation will be more robust than the current data scenarios (growth, peak, decline) approach. Basically, each model will be fit to rolling windows of 14 days, using 10 weeks for training and 2 weeks for testing/evaluation.

I agree in principle - but I think this would move this more into paper territory as opposed to a vignette. If you're up for extending it in this way I'd suggest to write it up as a paper (with all the code in a dedicated repo) and write up a guidance document here that's based on the lessons learned from the study.

jamesmbaazam:
> > I've given this a thought and I think a rolling window evaluation will be more robust than the current data scenarios (growth, peak, decline) approach. Basically, each model will be fit to rolling windows of 14 days, using 10 weeks for training and 2 weeks for testing/evaluation.
> 
> I agree in principle - but I think this would move this more into paper territory as opposed to a vignette. If you're up for extending it in this way I'd suggest to write it up as a paper (with all the code in a dedicated repo) and write up a guidance document here that's based on the lessons learned from the study.

I agree and will do that. Are we moving to close this PR? 

We could consider a bullet style list of alternatives for users to try when aiming to speed things up. I always get the question of how to speed things up but agree this vignette got out of control and had many scope creeps including evaluating the model's performance aside the speed up.

sbfnk:
> I agree and will do that. Are we moving to close this PR?

Happy to - we can later address the linked issue with 

> a bullet style list of alternatives for users to try when aiming to speed things up

jamesmbaazam:
Please do. I'll move the content here to a paper repo and add an analysis plan before I proceed.

PR #716, opened 2024-07-11
pass check if only tolerance is set

### LINKED ISSUE #665: setting tolerance instead of max in delay options doesn't work
Summary:
The `delay_opts()` family of functions sets a default tolerance and thus should work without setting a maximum. Currently, however, doing so results in an error.

Reproducible Steps:
``` r
library("EpiNow2")
#> 
#> 
#>     Gamma
generation_time_opts(Gamma(mean = 3.6, sd = 3.1))
#> Error in check_stan_delay(dist): All distribution passed to the model need to have a finite maximum
```

<sup>Created on 2024-05-09 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Related: #533

sbfnk:
## Description

This PR closes #665.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 ## Bug fixes
 
-- a bug was fixed that caused delay option functions to report an error if only the tolerance was specified. By @sbfnk.
+- a bug was fixed that caused delay option functions to report an error if only the tolerance was specified. By @sbfnk in #716 and reviewed by @jamesmbaazam.
 - a bug was fixed where `forecast_secondary()` did not work with fixed delays. By @sbfnk in #717 and reviewed by @seabbs.
 
 ## Documentation
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,10 @@
 
 - `epinow()` now returns the "timing" output in a "time difference"" format that is easier to understand and work with. By @jamesmbaazam in #688 and reviewed by @sbfnk.
 
+## Bug fixes
+
+- a bug was fixed that caused delay option functions to report an error if only the tolerance was specified. By @sbfnk.
+
 ## Documentation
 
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -10,7 +10,8 @@
 
 ## Bug fixes
 
-- a bug was fixed where `forecast_secondary()` did not work with fixed delays. By @sbfnk in #717 and reviewed by @sbfnk.
+- a bug was fixed that caused delay option functions to report an error if only the tolerance was specified. By @sbfnk.
```
jamesmbaazam, line 13:
```suggestion
- a bug was fixed that caused delay option functions to report an error if only the tolerance was specified. By @sbfnk in #716 and reviewed by @jamesmbaazam.
```

R/checks.R
```diff
@@ -96,8 +96,9 @@ check_stan_delay <- function(dist) {
       "and infinite maximum."
     )
   }
+  assert_numeric(attr(dist, "tolerance"), lower = 0, upper = 1)
   # Check that `dist` has a finite maximum
-  if (any(is.infinite(max(dist)))) {
+  if (any(is.infinite(max(dist))) && !(attr(dist, "tolerance") > 0)) {
     stop("All distribution passed to the model need to have a finite maximum")
```
jamesmbaazam, line 102: I think this error message can be improved. For example, "If tolerance is set to 0, all distributions must have a finite `max`."

```r
> delay_opts(LogNormal(mean = 4, sd = 1), tolerance = 0)
Error in check_stan_delay(dist) : 
  All distribution passed to the model need to have a finite maximum
```

jamesmbaazam (changes requested):
Thanks, Seb. This is a better fix than I suggested in the issue.

R/checks.R
sbfnk: Have reworded to "All distribution passed to the model need to have a finite maximum, which can be achieved either by setting `max` or non-zero `tolerance`."

jamesmbaazam (approved):

Issue #525, opened 2023-11-30
Change all priors to new dist interface, not just delays


sbfnk:
For a future PR this should probably be updated to use `normal()`.

_Originally posted by @seabbs in https://github.com/epiforecasts/EpiNow2/pull/504#discussion_r1408162989_
            
This will require moving the conversion to discretised distribution to the processing for stan (which is possibly a good idea anyway).

sbfnk:
Will be picked up once #504 is merged.

jamesmbaazam:
The linked comment does not lead anywhere so I'm struggling to get the context. I came [here](https://epiforecasts.io/EpiNow2/articles/estimate_infections_workflow.html#initial-reproduction-number) after seeing the code chunk here needs to be updated.

PR #724, opened 2024-07-17
Allow specifying tolerance at to `dist_spec` upon definition

### LINKED ISSUE #533: specify tolerance instead of max for distributions
Instead of specifying the maximum one could support specifying a tolerance, i.e. minimum `pmf` / `1-cmf` as is already implemented in the convolution (or the conversion to stan code if going with the approach in #504). This might require looping over incremental values of `max` to find the "correct" one.

sbfnk:
## Description

This PR closes #533.

Doing so and paving the way for #525 required a few changes to the infrastructure of how distributions are handled. This PR:

- introduces a `multi_dist_spec` class representing distributions that are a convolution of multiple distributions
- adds accessor functions `get_parameters()` and `get_distribution()` for extracting these from single or combined distributions, and `ndist()` for getting the number of distributions that make up a combined distribution
- makes `max` and `tolerance` attributes, with functionality to specify the tolerance either at the level of the convolved or constituent distribution(s)
- renames `apply_tolerance()` to `bound_dist()` as it now also applies the `max`
- updates the plotting function to work with continuous or discrete distributions
- updates the design document to reflect these changes 


R/dist_spec.R
```diff
@@ -131,6 +131,9 @@ discrete_pmf <- function(distribution =
 #' This combines the parameters so that they can be fed as multiple delay
 #' distributions to [epinow()] or [estimate_infections()].
 #'
+#' Note that distributions that already are combinations of other distributions
```
jamesmbaazam, line 134: Related to this, can we add a more informative error for the case where one wants to nest an already nested parameter. Here, user is trying to make the mean of the mean of the Normal uncertain.

```r
> dist2 <- Gamma(
+     mean = Normal(Normal(3, 3), 0.5), sd = Normal(2, 0.5), max = 20
+ )
Error in if (rel_unc > 0) { : missing value where TRUE/FALSE needed
In addition: Warning message:
In new_dist_spec(params, "gamma", ...) :
  Uncertain gamma distribution specified in terms of parameters that are not the "natural" parameters of the distribution (shape, rate). Converting using a crude and very approximate method that is likely to produce biased results. If possible, it is preferable to specify the distribution directly in terms of the natural parameters.
```

R/dist_spec.R
```diff
@@ -150,14 +153,30 @@ discrete_pmf <- function(distribution =
 c.dist_spec <- function(...) {
   ## process delay distributions
   dist_specs <- list(...)
-  if (!(all(vapply(dist_specs, is, FALSE, "dist_spec")))) {
+  if (length(dist_specs) == 1) return(dist_specs[[1]])
+  if (!(all(vapply(dist_specs, is, "dist_spec", FUN.VALUE = logical(1))))) {
     stop(
       "Distribution can only be concatenated with other delay ",
       "distributions."
```
jamesmbaazam, line 160:
```suggestion
      "All distributions must be of type `<dist_spec>`."
```

R/dist_spec.R
```diff
@@ -190,43 +209,48 @@ c.dist_spec <- function(...) {
 #' # The mean of the sum of two distributions
 #' mean(dist1 + dist2)
 mean.dist_spec <- function(x, ..., ignore_uncertainty = FALSE) {
```
jamesmbaazam, line 211: L187 says "ignoring any uncertainty" but I see there's an option to not do that here so maybe the doc should be updated?

R/dist_spec.R
```diff
@@ -235,53 +259,57 @@ mean.dist_spec <- function(x, ..., ignore_uncertainty = FALSE) {
...
-      return(0)
+sd.dist_spec <- function(x, ...) {
+  if (x$distribution == "nonparametric") {
+    ## nonparametric
+    mean_pmf <- sum((seq_along(x$pmf) - 1) * x$pmf)
+    return(sum((seq_along(x$pmf) - 1)**2 * x$pmf) - mean_pmf^2)
+  } else {
+    ## parametric
+    if (!all(vapply(x$parameters, is.numeric, logical(1)))) {
+      return(NA_real_)
```
jamesmbaazam, line 284: Should this warn at least?

R/dist_spec.R
```diff
@@ -235,53 +259,57 @@ mean.dist_spec <- function(x, ..., ignore_uncertainty = FALSE) {
...
-      } else if (y$distribution == "fixed") {
-        0
-      } else {
-        stop(
-          "Don't know how to calculate standard deviation of ",
-          y$distribution, " distribution."
-        )
-      }
+      stop(
+        "Don't know how to calculate standard deviation of ",
```
jamesmbaazam, line 297:
```suggestion
        "Cannot calculate standard deviation of ",
```

jamesmbaazam:
I expect `bound_dist()` to always set attributes but it only does if specified. Is this expected behaviour?

```r
> attributes(LogNormal(1,2))
$names
[1] "parameters"   "distribution"

$class
[1] "dist_spec" "list"     

> attributes(LogNormal(1,2, max= 1, tolerance = 0.1))
$names
[1] "parameters"   "distribution"

$class
[1] "dist_spec" "list"     

$max
[1] 1

$tolerance
[1] 0.1
```

jamesmbaazam:
> I expect `dist_spec()`, which calls `bound_dist()` internally, to always set the `max` and `tolerance` attributes but it only does if specified. Is this expected behaviour?
> 
> ```r
> > attributes(LogNormal(1,2))
> $names
> [1] "parameters"   "distribution"
> 
> $class
> [1] "dist_spec" "list"     
> 
> > attributes(LogNormal(1,2, max= 1, tolerance = 0.1))
> $names
> [1] "parameters"   "distribution"
> 
> $class
> [1] "dist_spec" "list"     
> 
> $max
> [1] 1
> 
> $tolerance
> [1] 0.1
> ```

Upon further digging, I see it rather only sets them when they're finite. I think for consistency, it would probably be better to always set them and rather not use them downstream if `max = Inf` or `tolerance = 0`.

R/dist_spec.R
```diff
@@ -235,53 +259,57 @@ mean.dist_spec <- function(x, ..., ignore_uncertainty = FALSE) {
...
+        x$distribution, " distribution."
```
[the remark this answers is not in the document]
jamesmbaazam: Also, I believe this error will never be called because there's a default method that calls `stats::sd(x)`? It's probably safer to move it to the default method.

R/dist_spec.R
```diff
@@ -343,42 +384,49 @@ max.dist_spec <- function(x, ...) {
 #'
 #' # The maxf the sum of two distributions
 #' discretise(dist1 + dist2, strict = FALSE)
-discretise <- function(x, strict = TRUE) {
+discretise.dist_spec <- function(x, strict = TRUE, ...) {
   if (!is(x, "dist_spec")) {
```
jamesmbaazam, line 388: This should now be converted to a `check_dist_spec()` function + another for `multi_dist_spec()` as they are repeated a lot here.

R/dist_spec.R
```diff
@@ -549,75 +565,105 @@ print.dist_spec <- function(x, ...) {
 #' Note that currently uncertainty in distributions is not plot.
```
jamesmbaazam, line 565: I guess this no longer applies.

R/dist_spec.R
```diff
@@ -549,75 +565,105 @@ print.dist_spec <- function(x, ...) {
...
+        return(is.na(sd_dist) || sd_dist > 0)
+      }, logical(1))
+      if (!any(uncertain)) {
+        samples <- 1 ## only need 1 sample if fixed
+      }
+      dists <- lapply(seq_len(samples), function(y) {
+        fix_dist(extract_single_dist(x, i), strategy = "sample")
+      })
+      tolerance <- attr(x, "tolerance")
+      if (is.null(tolerance)) {
```
jamesmbaazam, line 620: See my comment about always setting `tolerance` in https://github.com/epiforecasts/EpiNow2/pull/724#issuecomment-2252676751. It'll cut out these logics.

R/deprecated.R
```diff
@@ -633,3 +633,37 @@ dist_skel <- function(n, dist = FALSE, cum = TRUE, model,
...
+#'
+#' @description `r lifecycle::badge("deprecated")`
+#' This function is deprecated. Use `bound_dist()` instead.
+#' @param x A `<dist_spec>`
+#' @param tolerance Numeric; the desired tolerance level. Any part of the
+#' cumulative distribution function beyond 1 minus this tolerance level is
+#' removed.
+#' @return A `<dist_spec>` where probability masses below the threshold level
+#' have been removed
+#' @keywords internal
```
jamesmbaazam, line 647: Do we need to deprecate internal code?

jamesmbaazam (changes requested):
Just a few comments here and there that need addressing.

R/dist_spec.R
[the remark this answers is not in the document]
sbfnk: This bit of code should really never be called but this is supposed to cater for the hypothetical situation where someone implements a distribution but doesn't provide an analytical formula here for estimating the standard deviation. It's just a safety net.

R/deprecated.R
sbfnk: This was an external function, it's the deprecation that has caused this to be converted to internal.

sbfnk:
> Upon further digging, I see it rather only sets them when they're finite. I think for consistency, it would probably be better to always set them and rather not use them downstream if max = Inf or tolerance = 0.

I'm in two minds about this. I see your point but setting them is a way to signal that the user has not made a choice here which then allows e.g. the delay option functions to make an informed default choice when distributions are used for a specific purpose.

With this in mind, does the current set up make more sense, or do you think we should try to work around a set up where these attributes are always set.

R/dist_spec.R
sbfnk: I would gently push back on this as the only repeated thing is the class check so I think having a separate function adds is overkill here.

R/dist_spec.R
sbfnk: Fixed in 1825fe42c1d3e2021fe12aba2e2fdb9ddef9d545

R/dist_spec.R
sbfnk: Fixed in 6b35983fe3604b1980b2260aa8a8acce918f58bf

R/dist_spec.R
sbfnk: I think we use this a fair bit internally (the post-processing the NAs) so this would produce a lot of warnings. I've added a note to the documentation instead in 07eb4fb17d29524271ec65cc08ada262d11c6089

R/dist_spec.R
sbfnk: Fixed in 2aadbbd6525b146089d3aeb9602e665713d1b245

jamesmbaazam:
> > Upon further digging, I see it rather only sets them when they're finite. I think for consistency, it would probably be better to always set them and rather not use them downstream if max = Inf or tolerance = 0.
> 
> I'm in two minds about this. I see your point but setting them is a way to signal that the user has not made a choice here which then allows e.g. the delay option functions to make an informed default choice when distributions are used for a specific purpose.

Mmmh I think we may be thinking of the word "setting" differently here. I thought the attributes are rather set when the user makes a choice? i.e., In the case where a user sets a finite `max`, they have made a choice?

jamesmbaazam (approved):

sbfnk:
> Mmmh I think we may be thinking of the word "setting" differently here. I thought the attributes are rather set when the user makes a choice? i.e., In the case where a user sets a finite max, they have made a choice?

I agree. My point was that someone who specifies `Gamma(2, 2)` has not set a tolerance (and thus may be happy for the package to make an informed choice where we need distributions to be constrained) whereas someone who does `Gamma(2, 2, tolerance = 0)` has.

jamesmbaazam:
> > Mmmh I think we may be thinking of the word "setting" differently here. I thought the attributes are rather set when the user makes a choice? i.e., In the case where a user sets a finite max, they have made a choice?
> 
> I agree. My point was that someone who specifies `Gamma(2, 2)` has not set a tolerance (and thus may be happy for the package to make an informed choice where we need distributions to be constrained) whereas someone who does `Gamma(2, 2, tolerance = 0)` has.

Oh, I see. That makes sense.

sbfnk:
I've updated this to be clearer, i.e. now you get

```r
> gt_opts(Gamma(shape = 2, rate = 2))                                                          
Unconstrained distributon passed as a delay. Constraining with default tolerance 0.001
- gamma distribution:
  shape:
    2
  rate:
    2
```

R/dist_spec.R
sbfnk: see main discussion - keeping this as is for now but to be reviewed if/when we move `<dist_spec>` into a separate package.

R/dist_spec.R
```diff
@@ -307,22 +335,33 @@ sd_dist <- function(x) {
...
-      return(length(y$pmf) - 1)
-    } else if (y$distribution == "fixed") {
-      return(y$parameters$value)
+  if (get_distribution(x) == "nonparametric") {
+    ## nonparametric
+    return(length(get_pmf(x)) - 1)
+  } else if (get_distribution(x) == "fixed") {
+    return(get_parameters(x)$value)
+  } else {
+    if (is.null(attr(x, "max"))) {
```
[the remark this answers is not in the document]
sbfnk: see main discussion - keeping this as is for now but to be reviewed if/when we move `<dist_spec>` into a separate package.

jamesmbaazam (approved):
LGTM.

PR #745, opened 2024-08-14
Optimise convolutions

seabbs:
## Description

This PR is unlinked to an issue but applies some optimises to speed up convolutions (a key pain point of the stan code base). It also replaces the custom `reverse_mf` function with the new `reverse` built in.


seabbs:
*Note waiting on benchmarks for a more formal speed comparison.*

seabbs:
Slightly surprisingly this doesn't seem to make a great deal of difference. Regardless I think it should be more stable and is definitely easier to test.

jamesmbaazam:
Thanks for this, Sam. I'll review it today.

jamesmbaazam:
Sorry, I missed this notification. I'll finish reviewing all your PRs by close of Thursday. I'm prioritising the GP vectorisations PR due to the speed gains.

inst/stan/functions/convolve.stan
```diff
@@ -1,36 +1,100 @@
...
+ * Convolve a vector with a reversed probability mass function.
+ *
+ * This function performs a discrete convolution of two vectors, where the second vector
+ * is assumed to be an already reversed probability mass function.
+ *
+ * @param x The input vector to be convolved.
+ * @param y The already reversed probability mass function vector.
+ * @param len The desired length of the output vector.
+ * @return A vector of length `len` containing the convolution result.
+ * @throws If `len` is longer than the sum of lengths of `x` and `y`.
```
jamesmbaazam, line 45:
```suggestion
 * @throws If `len` is not of equal length to the sum of the lengths of `x` and `y`.
```

inst/stan/functions/convolve.stan
```diff
@@ -1,36 +1,100 @@
...
+ * @param xlen Length of the x vector
+ * @param ylen Length of the y vector
+ * @return An array of integers: {start_x, end_x, start_y, end_y}
+ */
+array[] int calc_conv_indices_xlen(int s, int xlen, int ylen) {
+  int s_minus_ylen = s - ylen;
+  int start_x = max(1, s_minus_ylen + 1);
+  int end_x = s;
+  int start_y = max(1, 1 - s_minus_ylen);
+  int end_y = ylen;
```
jamesmbaazam, line 14: Can we combine the two functions using an `ifelse` control?

```stan
 if (s <= xlen) {
    end_x = s;
    end_y = ylen;
  } else {
    end_x = xlen;
    end_y = ylen + xlen - s;
  }
```

inst/stan/functions/convolve.stan
```diff
@@ -1,36 +1,100 @@
...
+  
+  for (s in 1:xlen) {
+    array[4] int indices = calc_conv_indices_xlen(s, xlen, ylen);
+    z[s] = dot_product(x[indices[1]:indices[2]], y[indices[3]:indices[4]]);
+  }
+  
+  if (len > xlen) {
+    for (s in (xlen + 1):len) {
+      array[4] int indices = calc_conv_indices_len(s, xlen, ylen);
+      z[s] = dot_product(x[indices[1]:indices[2]], y[indices[3]:indices[4]]);
```
jamesmbaazam, line 68: I don't know if combining the two functions above will simplify the logic here though. We will still need to check if `xlen < OR > len` to determine which `for` loop index to use.

jamesmbaazam (approved):
This all looks good to me. The code is more readable now.

inst/stan/functions/convolve.stan
seabbs: we could but I'm not sure it really simplifies things (your point below)

inst/stan/functions/convolve.stan
seabbs: I think this is more efficient than that as its always do the first and then only do the second bit if its longer but yes agree.

inst/stan/functions/convolve.stan
jamesmbaazam: Yep. Was a first thought before seeing the logic below.

seabbs:
Nice and I agree. Thanks for the review. This is just waiting on our decisions in the conversations. My vote is leave as is for this PR and if we feel strongly move to a new issue to refactor? It could be a good issue for someone new to stan to tackle if we decide to go that way

jamesmbaazam:
I don't feel strongly about it.

PR #746, opened 2024-08-14
Use broadcasting vs loop in day of week

seabbs:
## Description

This PR is unrelated to an issue. It switches to using broadcasting vs a loop in the day of the week module. 

## Initial submission checklist

- [ ] My PR is based on a package issue and I have explicitly linked it.
- [x] I have tested my changes locally (using `devtools::test()` and `devtools::check()`).
- [x] I have added or updated unit tests where necessary.
- [x] I have updated the documentation if required and rebuilt docs if yes (using `devtools::document()`).
- [x] I have followed the established coding standards (and checked using `lintr::lint_package()`).
- [x] I have added a news item linked to this PR.

## After the initial Pull Request 

- [x] I have reviewed Checks for this PR and addressed any issues as far as I am able.

PR #747, opened 2024-08-15
Switching to broadcasting random walk

seabbs:
## Description

Following on from #746 this PR switches to broadcasting for the random walk parameterizations. It also adds docs and tests.


NEWS.md
```diff
@@ -27,6 +28,7 @@
 
 - Updated the documentation of the dots argument of the `stan_sampling_opts()` to add that the dots are passed to `cmdstanr::sample()`. By @jamesmbaazam in #699 and reviewed by @sbfnk.
 - `generation_time_opts()` has been shortened to `gt_opts()` to make it easier to specify. Calls to both functions are equivalent. By @jamesmbaazam in #698 and reviewed by @seabbs and @sbfnk .
+- Added stan documentation for `update_rt`. By @seabbs in #747 and reviewed by @jamesmbaazam.
```
jamesmbaazam, line 31:
```suggestion
- Added stan documentation for `update_rt()`. By @seabbs in #747 and reviewed by @jamesmbaazam.
```

R/create.R
```diff
@@ -261,15 +261,20 @@ create_future_rt <- function(future = c("latest", "project", "estimate"),
...
 #' create_rt_data(rt_opts(use_breakpoints = TRUE), breakpoints = rep(1, 10))
+#'
+#' # using random walk
+#' create_rt_data(rt_opts(rw = 7), breakpoints = rep(1, 10))
 #' }
 create_rt_data <- function(rt = rt_opts(), breakpoints = NULL,
                            delay = 0, horizon = 0) {
+
   # Define if GP is on or off
   if (is.null(rt)) {
```
jamesmbaazam, line 272: I've been thinking we should actually surface what settings are applied when `rt = NULL`. We only say that they can set it to NULL but we could add what settings are used when that is done. What do you think?

jamesmbaazam (approved):
Thanks @seabbs. To confirm that I understood everything here. This was more of a refactoring exercise to use more elegant stan tricks right?  I noticed most of the right hand side code is a better rewrite of the deleted code. i.e., what we would call vectoization in R parlance.

R/create.R
seabbs: yes I agree. A new issue though?

seabbs:
yes ideally and to ideally reduce compilation and run time (which I think it does a tiny bit)

R/create.R
jamesmbaazam: I've created a new issue.

jamesmbaazam (approved):

PR #748, opened 2024-08-19
Use multiplier to make rw non-centred

seabbs:
This is a small PR that tests making the random walk non-centred.


seabbs:
This doesn't look like it works when bp_sd is length 0 so won't be feasible for us

sbfnk:
Could we not use something like 

```stan
  array[bp_n] real<multiplier = (length(bp_sd > 0) ? bp_sd[1] : 0)> bp_effects;                   // Rt breakpoint effects
```

seabbs:
Maybe though I have no idea of what a multiplier of 0 does?

PR #751, opened 2024-08-21
Fix number of samples used in `bootstrapped_dist_fit()`

### LINKED ISSUE #725: `bootstrapped_dist_fit` example produces warnings about too many samples
Summary:
It uses 10 bootstrap iterations with 1000 samples resulting in 100 each which yields a warning (instead this should have 1000 each).

Reproducible Steps:
``` r
library("EpiNow2")
#> 
#> 
#>     Gamma

delays <- rlnorm(500, log(5), 1)
out <- bootstrapped_dist_fit(delays,
  samples = 1000, bootstraps = 10,
  dist = "lognormal"
)
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
#> Warning in EpiNow2::dist_fit(values, samples = samples, dist = dist): `samples`
#> must be at least 1000. Now setting it to 1000 internally.
```

<sup>Created on 2024-07-17 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

jamesmbaazam:
## Description

This PR closes #725.

This PR fixes an issue where the number of samples per bootstrapped was wrongly divided by the number of bootstraps. I also took the opportunity to improve the documentation and added a test for the expected warning and object.


seabbs:
I'm not sure this is correct. My read of `samples` is that it should be the totally number of samples returned by `boostrapped_dist_fit()`/

NEWS.md
```diff
@@ -11,6 +11,7 @@
 - The interface for defining delay distributions has been generalised to also cater for continuous distributions
 - When defining probability distributions these can now be truncated using the `tolerance` argument
 - Ornstein-Uhlenbeck and 5 / 2 Matérn kernels have been added. By @sbfnk in # and reviewed by @.
+- Fixed the number of samples drawn per bootstrap in `bootstrapped_dist_fit()` in #751 by @jamesmbaazam and reviewed by @<REVIEWER>.
```
jamesmbaazam, line 14:
```suggestion
- Fixed the number of samples drawn per bootstrap in `bootstrapped_dist_fit()` in #751 by @jamesmbaazam and reviewed by @seabbs.
```

jamesmbaazam:
> I'm not sure this is correct. My read of `samples` is that it should be the totally number of samples returned by `boostrapped_dist_fit()`/

This sounds a little confusing to me. `dist_fit()` will always use at least `1000` samples, so if `bootstraps` = $N (N > 1)$, we will end up with $N \times `samples`$. If the size of the samples to be returned by `bootstrapped_dist_fit()` is to always remain as `samples`, then I guess we need to combine all the samples and resample `samples` before summarising?

seabbs:
So I don't think we will with the unedited code (as it only samples `samples / boostraps` from each bootstrap. You could also take more samples from each bootstrap and then resample but that would be computationally wasteful

R/estimate_delay.R
```diff
@@ -178,24 +179,28 @@ bootstrapped_dist_fit <- function(values, dist = "lognormal",
...
     set_dt_single_thread()
 
     fit <- EpiNow2::dist_fit(values, samples = samples, dist = dist)
 
     out <- list()
     if (dist == "lognormal") {
-      out$meanlog <- sample(extract(fit)$mu, samples)
-      out$sdlog <- sample(extract(fit)$sigma, samples)
+      out$meanlog <- sample(extract(fit)$mu, single_sample_size)
+      out$sdlog <- sample(extract(fit)$sigma, single_sample_size)
```
seabbs, line 191: I mentioned this as a solution but I don't think its a good idea (i.e its computationally wasteful). Can we walk back to the actual bug we are trying to fix here?

jamesmbaazam: So, this change is to differentiate the two "samples" used in the scope of this function but now it differentiates how many samples are used in each fit and how many are returned per bootstrap.

I think it does the same as the original so the "computational wastefulness" is still intact? 😅 .

jamesmbaazam: > Can we walk back to the actual bug we are trying to fix here?

We want the total samples returned to be irrespective of the number of bootstraps right?

seabbs: but is that not already the case in the non-editted code?

jamesmbaazam: Somewhat. The original code tries to pass the default `sample = 1` in `get_single_dist()` to `dist_fit()` so we get the warning when it's overwritten to 1000. This new version ensures the right `sample` is passed to `dist_fit()`.

I'm unsure if the original issue is to address the warning being thrown or the number of samples being used in the summarised distribution.

seabbs: I think we want a fix for the single dist scenario vs a global refactor then?

sbfnk: Two possible solutions are:
1. change the example to `samples = 10000`
2. remove the hard overwriting and warning of `N` if `N<1000`
Perhaps (2) makes more sense - are we sure 1000 is a magic number here?

jamesmbaazam: How about refactor `get_single_dist()` like so

```diff
- get_single_dist <- function(values, samples = 1) {
+ get_single_dist <- function(values, samples = samples) { 
    set_dt_single_thread()

    fit <- EpiNow2::dist_fit(values, samples = samples, dist = dist)

    out <- list()
    if (dist == "lognormal") {
-      outmeanlog<−sample(extract(fit)mu, samples)
+      outmeanlog<−sample(extract(fit)mu, 1)
-      outsdlog<−sample(extract(fit)sigma, samples)
+      outsdlog<−sample(extract(fit)sigma, 1)
  } else if (dist == "gamma") {
-      outshape<−sample(extract(fit)alpha, samples)
+     outshape<−sample(extract(fit)alpha,  1)
-      outrate<−sample(extract(fit)beta, samples)
+      outrate<−sample(extract(fit)beta, 1)
    }
    return(out)
  }
```

sbfnk: I don't think I'm following what's happening here - what is `samples` (which doesn't seem to be used)?

jamesmbaazam: It's passed to `dist_fit()` so that the warnings are suppressed since the right number of samples will be used. 

From what I understand from the original code, the whole idea of `get_single_dist()` is to get only 1 sample out of the `samples` returned by `dist_fit()`.

sbfnk: I think to get a total of `samples` samples we'll need to get `samples / bootstraps` from each bootstrap as @seabbs stated above. Getting one each time will only yield `bootstraps` samples.

I still think this makes most sense - why not allow fewer than 1000 samples?

> remove the hard overwriting of `N` and warning if `N<1000

Alternative would be this (what you're suggesting but with than `samples / bootstraps` samples instead of 1.

> take more samples from each bootstrap and then resample but that would be computationally wasteful

jamesmbaazam: Option 2 will still yield the warnings if the result of `samples / bootstraps` < 1000, so, I am happy to implement option 1. Not sure why it originally had a lower bound of 1000.

sbfnk: Sounds good to me.

NEWS.md
```diff
@@ -23,6 +23,7 @@
 - Optimised convolution code to take into account the relative length of the vectors being convolved. See #745 by @seabbs and reviewed by @jamesmbaazam.
 - Switch to broadcasting the day of the week effect. By @seabbs in #746 and reviewed by @jamesmbaazam.
 - A warning is now thrown if nonparametric PMFs passed to delay options have consecutive tail values that are below a certain low threshold as these lead to loss in speed with little gain in accuracy. By @jamesmbaazam in #752 and reviewed by @seabbs.
+- Fixed the number of samples drawn per bootstrap in `bootstrapped_dist_fit()` in #751 by @jamesmbaazam and reviewed by @seabbs.
```
jamesmbaazam, line 26:
```suggestion
- `dist_fit()` can now accept any number of `samples` without throwing a warning when `samples` < 1000 in #751 by @jamesmbaazam and reviewed by @seabbs and @sbfnk.
```

Issue #523, opened 2023-11-28
Change "back-calculation" to "nonparametric" infection model throughout


jamesmbaazam:
At some point we should change back-calculation to nonparametric infection model throughout as that's what we're calling it in the model definition vignette.

_Originally posted by @sbfnk in https://github.com/epiforecasts/EpiNow2/pull/521#discussion_r1407455280_

sbfnk:
To be revisited after #345

jamesmbaazam:
To confirm, this is just a simple re-phrasing exercise right?

seabbs:
Are we sure non-mechanistic is the right name? 

https://github.com/epiforecasts/EpiNow2/blob/dbb6536262e68b32a60dab29e3af731b11b1875e/inst/stan/functions/infections.stan#L69

The model is really a few things right: 

1. A pretty standard deconvolution approach but in the Bayesian context
2. An exponential growth model
3. A log link model (this is closest to being a "non-mechanistic model" in work with @samuelbrand we call this a direct infections model)


One option could be to break those things up either with or without an overarching naming wrapper

seabbs:
Looking into this in the using the `estimate_infections` vignette we call this non-parametric and in the model definition vignette we call it non-mechanistic

sbfnk:
The way I see it the key distinction from the default model is that it doesn't use the renewal equation so if we call that the mechanism (for generating new infections) it makes it non-mechanistic. We could call it "non-renewal" instead to be more explicit?

jamesmbaazam:
> Non-renewal

Nice!

seabbs:
One option would be a larger breaking change to make the interface: 

For renewal

```r
estimate_infections(...
  model = rt_opts()
...
)
```

For deconvolution

```r
estimate_infections(
	model = deconvolution_opt()
```

For growth rate


```r
estimate_infections(
	model = growth_opts()
)
```

sbfnk:
I really like this idea - for names perhaps we could use

```r
estimate_infections(...
  model = renewal_opts()
...
)
```

instead of `rt_opts()`?

> For growth rate
> 
> ```r
> estimate_infections(
> 	model = growth_opts()
> )
> ```

We never implemented this as it's the same as renewal with `gt = Fixed(1)`.

sbfnk:
Noting that the original issue was about wording in the docs so whilst obviously related the change of options might be better suited to a separate issue/PR combo.

sbfnk:
[...]

PR #779, opened 2024-09-19
Make `rt=NULL` settings explicit

### LINKED ISSUE #761: Document/surface what settings are applied when `rt = NULL`.
I've been thinking we should actually surface what settings are applied when `rt = NULL`. We only say that they can set it to NULL but we could add what settings are used when that is done. What do you think?

_Originally posted by @jamesmbaazam in https://github.com/epiforecasts/EpiNow2/pull/747#discussion_r1734718474_

jamesmbaazam:
## Description

This PR closes #761.

This PR makes explicit the settings that are applied when `rt=NULL` in functions that use that argument.


NEWS.md
```diff
@@ -2,6 +2,7 @@
 
 ## Documentation
 
+- The documentation of the `rt` argument has been expanded in the case where `rt = NULL` to make explicit the settings that are applied in that case. By @jamesmbaazam in #779 and reviewed by <REVIEWED>.
```
sbfnk, line 5:
```suggestion
- The documentation of the `rt` argument has been expanded in the case where `rt = NULL` to make explicit the settings that are applied in that case. By @jamesmbaazam in #779 and reviewed by @sbfnk.
```

R/create.R
```diff
@@ -240,7 +240,9 @@ create_future_rt <- function(future = c("latest", "project", "estimate"),
 #' stan.
 #' @param rt A list of options as generated by [rt_opts()] defining Rt
 #' estimation. Defaults to [rt_opts()]. Set to `NULL` to switch to using back
-#' calculation rather than generating infections using Rt.
+#' calculation rather than generating infections using Rt. This is the
```
sbfnk, line 243: I think we decided to move away from back calculation so I'd just say something like

"Set to `NULL` to switch to using the nonmechanistic model instead of the renewal equation model to generate new infections." or something like that.

sbfnk (changes requested):
Good idea

R/create.R
jamesmbaazam: I wanted to change the back calculation wording but there's another issue for that here https://github.com/epiforecasts/EpiNow2/issues/523, which is also dependent on the in-progress PR.

jamesmbaazam: > I think we decided to move away from back calculation so I'd just say something like
> 
> "Set to `NULL` to switch to using the nonmechanistic model instead of the renewal equation model to generate new infections." or something like that.

Are you okay with the new text that states the internal settings explicitly?

sbfnk: yep

jamesmbaazam: See [ac33cf4](https://github.com/epiforecasts/EpiNow2/pull/779/commits/ac33cf4a121135d533e565d3b7c2d8125c519785).

sbfnk (approved):

PR #798, opened 2024-09-27
Move to suggest

### LINKED ISSUE #796: Dependencies
With the recent switch to `{cli}` we now have 21 dependencies which is above CRAN's limit of 20 (not sure how strictly this is enforced). It would be good to consider which dependencies could be removed.

Candidates are:
- `{future}`, `{future.apply}`, `{futile logger}` and remove corresponding functionality, potentially not ideal as might be breaking for people using this in production
- `{R.utils}` - likewise only used for production for timing out long runs
- `{patchwork}` only used for two plots - could they be faceted instead?
- `runner::mean_run()` only used once, in `create_shifted_cases()`, also `truncnorm::rtruncnorm` only used in two places -- both of these could be re-implemented fairly trivially, but I'm not sure what the point of that would be.

Let's see how we do with the 1.6.0 submission and then potentially reconsider.

sbfnk:
## Description

This PR closes #796.

It makes the `{future}` packages and `{progressr}` suggests instead of depends, and only uses their functionality if available.

Probably the same could be done for `{futile.logger}` but that would be a lot more complicated.


NEWS.md
```diff
@@ -30,6 +30,7 @@
 
 - `fix_dist()` has been renamed to `fix_parameters()` because it removes the uncertainty in a distribution's parameters. By @sbfnk in #733 and reviewed by @jamesmbaazam.
 - `plot.dist_spec` now uses color instead of line types to display pmfs vs cmfs. By @jamesmbaazam in #788 and reviewed by @sbfnk.
+- The use of the `{progressr}` package for displaying progress bars is now optional, as is the use of `{future}` and `{future.apply}` for parallelisation. By @sbfnk in By @sbfnk in
```
seabbs, line 33:
```suggestion
- The use of the `{progressr}` package for displaying progress bars is now optional, as is the use of `{future}` and `{future.apply}` for parallelisation. By @sbfnk in #798 and reviewed by @seabbs.
```

seabbs (approved):
Nice. LGTM

seabbs (approved):

seabbs (approved):

PR #799, opened 2024-09-29
Issue 793: Improve documentation for those specifying a non-parametric delay

### LINKED ISSUE #793: Improve documentation for those passing in a GI
Type of issue:
Commenting on existing documentation/ language in vignettes

Detail:
While the model definition makes it clear that the implementation of the renewal approach expects time indexing starting at day 1, those not reading the model definition carefully may miss this. I think it would be nice to flag to the user that if they pass in a vector to the GI, its assumed to be 1 indexed not 0 indexed (so the first value reflects the probability of transmission on day 1 of infection not day 0, the day they were infected). 

Could add to:
- the main EpiNow2 vignette (would need to add an example where GI is being set by the user)
- the documentation for `generation_time` or `generation_time_opts`

---------

kaitejohnson:
## Description

This PR closes #793.

It expands on the vignette details describing how to specify a generation interval and reporting delay. Specifically, it gives an example of how to provide a fixed non-parametric vector delay distribution. I added language explaining that the generation interval should be indexed starting at day 1 of an infection whereas the reporting delays should be indexed starting at day 0 (as is consistent with the model definition, but might not be clear to a user). 

I looked through the `generation_time_opts()` and the distributions handling but wasn't sure if this distinction was appropriate in either of those, happy to add some language about this in the documentation if theres a specific place you all think is most appropriate.

The context for this was that its easy to accidentally pass in a shifted by one generation interval distribution if assuming the GI is 0 indexed.

I think a separate issue could be to use `primarycensoreddist` to generate a GI pmf in `data-raw` and pass that in as an example either instead of or in addition to the current `example_generation_time`


sbfnk:
Thanks @kaitejohnson, this is great and definitely a very good idea to clarify.

I might be misremembering but I think generation intervals are in fact zero-indexed but the 0-component is set to zero as of version 1.4.0, here (`left_truncate` being set to 1 for generation intervals):
https://github.com/epiforecasts/EpiNow2/blob/95c36430dca62d40e57a3bf3968b488983891d93/inst/stan/functions/delays.stan#L64

> The context for this was that its easy to accidentally pass in a shifted by one generation interval distribution if assuming the GI is 0 indexed.

The point that this is not clear and easy to get wrong by accident obviously stands regardless. I think it would be great to clarify also in `gt_opts()` as you suggest.

> I think a separate issue could be to use primarycensoreddist to generate a GI pmf in data-raw and pass that in as an example either instead of or in addition to the current example_generation_time

I agree.

kaitejohnson:
I see I missed that change, can adjust the language accordingly. 

To make sure my interpretation is now correct, if a user now passes in a GI pmf, it should be 0-indexed (just like the other delay pmfs), because under the hood the left truncation and renormalization occurs (in the function you sent).

sbfnk:
> To make sure my interpretation is now correct, if a user now passes in a GI pmf, it should be 0-indexed (just like the other delay pmfs), because under the hood the left truncation and renormalization occurs (in the function you sent).

Yes, exactly.

seabbs:
Looking at the code this does appear to be the case and I must say I had no idea. 

I think we should throw an information message in `gt_opts` when a fixed distribution is passed as for thos passing vectors I can see this being very confusing.

dylanhmorris:
If `NonParametric` generation interval PMFs are going to be zero-indexed, I suggest erroring whenever there is non-zero mass in the 0 bin. I think this will cause the fewest (and the least costly) surprises for users.

If a user wishes to specify a PMF bin-by-bin, it's reasonable to ask them to decide explicitly how to handle any mass in the 0th bin. So failing if a user falsely assumes 1-indexing is imo worth the cost of forcing drop-and-renormalize users to perform that operation manually.

sbfnk:
> If `NonParametric` generation interval PMFs are going to be zero-indexed, I suggest erroring whenever there is non-zero mass in the 0 bin. I think this will cause the fewest (and the least costly) surprises for users.

That, I think, is a really good suggestion and would do away with need for an explicit warning in the case that the first bin is zero.

kaitejohnson:
The vignette is updated to explain that the GI should be 0 indexed with a mass of 0 on the first element. I didn't adjust any documentation in `generation_time_opts()` as that seems to be addressed in #808. 

If this feels like too much detail for a vignette, feel free to ignore. I think the warning will flag for most using a fixed non-parametric GI.

sbfnk (changes requested):
Content-wise this looks great now.

Just one request: could you move these changes over to `EpiNow2.Rmd.orig`? The `.Rmd` files are generated from these with pre-computed results (so they don't have to be re-rendered every time we build the package which takes quite a lot of time/computation). We can then build the Rmd via the corresponding action in https://github.com/epiforecasts/EpiNow2/blob/main/.github/workflows/render-EpiNow2.yaml

seabbs:
And add yourself to the DESCRIPTION as a contributor

vignettes/EpiNow2.Rmd
```diff
@@ -87,6 +87,22 @@ example_incubation_period
...
+
+```r
+example_non_parametric_gi <-  NonParametric(pmf = c(0, 0.3, 0.5, 0.2))
+
+example_non_parametric_delay <- NonParametric(pmf = c(0.01, 0.1, 0.5, 0.3, 0.09))
+```
+These distributions are passed to downstream functions in the same way that the 
+parametric distributions are. The difference in the `NonParametric` case is that 
+these will be treated as fixed rather than estimated with the specified parametric priors. 
+
```
sbfnk, line 104:
```suggestion
parametric distributions are. 
```

This is not strictly true unless (normal) priors are specified for the parameters of the parametric distributions. If passing a parametric distributions with fixed parameters (e.g. `Lognormal(meanlog = 1, sdlog = 1)`) these will internally be converted to a nonparametric distribution and treated as fixed.

kaitejohnson:
I am guessing its ok that I didn't remove the changes from `EpiNow2.Rmd` since these will get overwritten by the GH action?

sbfnk:
> I am guessing its ok that I didn't remove the changes from `EpiNow2.Rmd` since these will get overwritten by the GH action?

You’re guessing right I think.

sbfnk (approved):

vignettes/EpiNow2.Rmd
```diff
@@ -87,6 +87,20 @@ example_incubation_period
...
 ```
 
+Users can also pass a non-parametric delay distribution vector using the `NonParametric` option 
+for both the generation interval and reporting delays. It is important to note that if doing so,
+both delay distributions are 0-indexed, meaning the first element corresponds to the probability mass
+at day 0 of an individual's infection. Because the discretised renewal equation doesn't support mass on day 0, the generation interval should be passed in as a 0-indexed vector with a mass of zero on day 0. 
+If this is not the case, a warning will indicate that the vector is being left-truncated and renormalized.
+
+```r
+example_non_parametric_gi <-  NonParametric(pmf = c(0, 0.3, 0.5, 0.2))
```
jamesmbaazam, line 97: I know this has been merged but given the preceding text, I would have expected to see the warning, i.e., the example would use `gt_opts()` like so

```r
> gt_opts(NonParametric(pmf = c(0.1, 0.3, 0.5, 0.2)))
- nonparametric distribution
  PMF: [0.091 0.27 0.45 0.18]
Warning message:
Specifying nonparametric generation times with nonzero first element was deprecated in
EpiNow2 1.6.0.
ℹ Since zero generation times are not supported by the model, the generation time will be
  left-truncated at one.
ℹ In future versions this will cause an error. Please ensure that the first element of
  the nonparametric generation interval is zero.
ℹ The deprecated feature was likely used in the EpiNow2 package.
  Please report the issue at <https://github.com/epiforecasts/EpiNow2/issues>.
This warning is displayed once every 8 hours.
Call `lifecycle::last_lifecycle_warnings()` to see where this warning was generated. 
```

kaitejohnson: @jamesmbaazam Do you think it makes sense to include a non-parametric pmf that produces this warning as you have shown?

The one currently included has a value of 0 on day 0, so it won't produce the warning! Perhaps the preceding text isn't clear enough.

jamesmbaazam: I think the text is fine but personally would have expected the code sample to showcase the warning thrown when the delay is not 0-indexed. Additionally, I would suggest to explicitly print the results of  `example_non_parametric_gi` and `example_non_parametric_delay` in the vignette.

kaitejohnson: I can open an issue and subsequent PR to add this.

Do you think there should be an example for both the correct specification (so with the 0 on day 0) and the incorrect specification that results in a warning (as you demonstrated)?  My only concern is about bloating the vignette, otherwise I think it could make sense to show both.

sbfnk: I'm not convinced we want to show a way of specifying that we're discouraging in the warning (and that will cause an error in the future). Perhaps we should just remove the highlighted sentence if it's confusing?

jamesmbaazam: @sbfnk You're right about not encouraging the wrong specification. 

Maybe, we can reword this part "If this is not the case, a warning will indicate that the vector is being left-truncated and normalized." -> "If this is not the case, the vector will be left-truncated and normalized."

I do see that the doc of `gt_opts()` has the following wording: "Because the discretised renewal equation used in the package does not support zero generation times, any distribution specified here will be left-truncated at one, i.e. the first element of the nonparametric or discretised probability distribution used for the generation time is set to zero and the resulting distribution renormalised." Shouldn't we just reuse that here?

sbfnk: I still think that in the vignette we should only tell users how to specify this correctly and not what happens if they fail to do so (which they'll find out with the warning anyway). So I'd vote for removing the sentence altogether.

jamesmbaazam: @kaitejohnson Would you like to take this one 😃 ?

PR #809, opened 2024-10-02
dependencies: NA for render actions

sbfnk:
all the render Rmd actions are failing because they can't install `cmdstanr`. I think this will fix it.


sbfnk:
Testing manually confirms it does so will merge and then trigger all the re-render actions.

seabbs:
Note yesterday I made a new version of the cmdstan action that can install cmdstanr internally if we need it to (I couldn't get it to install R without always reinstalling R). This version seems to work better across OS's.

PR #821, opened 2024-10-03
Update total downloads badge

seabbs:
Motivated in the same way as #820. Copies in the https badge. Doesn't match the colour but can do so.


seabbs:
I think the alternative is to change what we have currently from HTTP to https directly?

seabbs:
I'm not really sure this change is needed? The current badge loads for me on the pkgdown site but not on Github. I have a bunch of custom stuff overlaying my github though so maybe its just me

sbfnk:
it doesn't load for me so this change is probably a good thing.

sbfnk (approved):

PR #831, opened 2024-10-16
Documentation of alpha

### LINKED ISSUE #829: definition of alpha in GP spectral densities
The vectorised formulation of the Matern 3/2 kernel spectral density in https://github.com/epiforecasts/EpiNow2/pull/742 has come with a slight change in definition of alpha - what is now called alpha is actually sqrt(alpha), i.e. the square root of the magnitude.

This has a few consequences:
- the description is not correct
- the prior specification is tighter than before, leading to some unexpected changes in the fits, see e.g. https://github.com/epiverse-trace/howto/issues/81

To fix this we could
1. update the descriptions and the prior for alpha
2. change alpha to sqrt(alpha) in the spectral density of the Matern 3/2 kernel

Option (1) would be more computationally efficient at least the Matern 3/2 kernels. The downside is that we've changed a parameter in the model so will have to update the model description (esp. https://github.com/epiforecasts/EpiNow2/blob/main/vignettes/gaussian_process_implementation_details.Rmd) and perhaps issue a note to the user (although I suspect this is not necessarily a parameter many users have been tweaking).

sbfnk:
## Description

This PR closes #829.

It clarifies the documentation of alpha.


seabbs (commented):
Noting the prior change was intentional based on prior predictives but happy for it to be widened. It might be good to split this out from the doc change.

seabbs (approved):

seabbs:
I also cleared it an reran. I think a patch fix here would be to set the cache to auto-delete after say a week . Might strike a better balance

Issue #772, opened 2024-09-15
Add vignette on modelling temporally aggregated data (i.e., `obs_opts(na = "accumulate")` )


jamesmbaazam:
Thanks a lot all - I've tested it out and it's working well now! I've included the updated figure below on the same data for reference, in case helpful! 

@jamesmbaazam regarding extra documentation, I think that at some point it could be useful to add a short, dedicated section on applying to weekly data. The current documentation mentions the rw estimator for weekly, but I think having a section with an example as you do for daily could be beneficial to others as we work a lot with weekly aggregated data. The other thing would probably be then to be able to visualize on weekly scale too, but as it's possible for end users to extract the estimates and create visuals I don't think it's really necessary. 

Thanks again for your help and for a great package! 

<img width="505" alt="weekly_est_epinow" src="https://github.com/user-attachments/assets/77fe349e-caf2-4106-b4a3-1c3e677cb2b3">

_Originally posted by @martinamcm in https://github.com/epiforecasts/EpiNow2/issues/767#issuecomment-2351691083_

jamesmbaazam:
By "surface", I mean we need a vignette walkthrough. My argument for this need is here https://github.com/epiforecasts/EpiNow2/pull/774#discussion_r1760886569.

seabbs:
Agree. Perhaps edit the title and add a plan/outline?

jamesmbaazam:
# Outline of vignette

- Discuss how EpiNow2 treats/prepares input data (into complete daily data)
- Motivate for data being reported at different types of intervals;
	- daily, 
	- irregular intervals (every few days), and
	- regular but at an aggregated scale (weekly, monthly)
- Walk-throughs of how EpiNow2 supports observations reported at non-daily intervals:
   - Treating missing dates as explicitly missing (`obs_opts(na = "missing")`)
   - Treating missing dates as accumulated (`obs_opts(na = "accumulate")`)

seabbs:
We might want to wait on this until we have some sort of clarity on the approach that is going to be used to deal with https://github.com/epiforecasts/EpiNow2/issues/547?

seabbs:
Sure but if any change is imminent that might be more work? On the other hand if we still need to discuss what #547 looks like it could help inform that discussion.

jamesmbaazam:
Agrees. Let's hold this off for now. The question about treatment of missingness has been coming up a lot lately.

seabbs:
[...]

Issue #548, opened 2024-02-14
Efficiency of accumulating NAs


sbfnk:
I guess I am saying that cases_time would become optionally an array with a sum. Another option might be to make this a matrix multiplication problem

_Originally posted by @seabbs in https://github.com/epiforecasts/EpiNow2/pull/534#discussion_r1489351211_

seabbs:
I think this is still active as there hasn't been any material changes to this part of the code?

sbfnk:
Yes, agree.

PR #851, opened 2024-11-11
updated interface for accumulation

### LINKED ISSUE #547: Distinguish NA (missing) from NA (accumulated)

sbfnk:
## Description

This PR closes #547 following the design agreed in #839.


R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
+##' Fills missing data in a data set to prepare it for use within the package
+##'
+##' @description `r lifecycle::badge("experimental")`
+##'
+##' @param data Data frame with a `date` column. The other columns depend on the
+##'   model that the data are to be used, e.g. [estimate_infections()] or
+##'   [estimate_secondary()]. See the documentation there for the expected
+##'   format.
+##' @param missing Character. Options are "missing" (the default), "accumulate"
+##'   and "zero". This determines how missing dates the data are interpreted.
```
seabbs, line 10:
```suggestion
##'   and "zero". This determines how missing dates in the data are interpreted.
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
+##' Fills missing data in a data set to prepare it for use within the package
+##'
+##' @description `r lifecycle::badge("experimental")`
+##'
+##' @param data Data frame with a `date` column. The other columns depend on the
+##'   model that the data are to be used, e.g. [estimate_infections()] or
+##'   [estimate_secondary()]. See the documentation there for the expected
+##'   format.
+##' @param missing Character. Options are "missing" (the default), "accumulate"
```
seabbs, line 9: When I first read this I didn't totally get what it was for but after a few passes I did. I wondered if a change name or a little sentence more clearly stating what missing was here would help but I couldn't come up with anything so perhaps it is just a difficult thing that people need to take the time to pass

seabbs: I think the thing that confused me most was `missing = "missing"`. Perhaps it should be `missing = "na"?

R/create.R
```diff
@@ -525,14 +528,16 @@ create_obs_model <- function(obs = obs_opts(), dates) {
...
-
   cases <- data[(seeding_time + 1):(.N - horizon)]
   complete_cases <- create_complete_cases(cases)
   cases <- cases$confirm
+  accumulate <- data[-(1:seeding_time)]$accumulate
 
   stan_data <- list(
     cases = complete_cases$confirm,
     cases_time = complete_cases$lookup,
+    any_accumulate = as.integer(any(accumulate > 0)),
```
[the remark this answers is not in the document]
jamesmbaazam: Nope

```r
> any(isTRUE(c(TRUE, FALSE, TRUE)))
[1] FALSE
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##' @param na Character. How to process dates that have NA values. The options
+##'   available are the same ones as for the `missing` argument.
+##' @param initial_accumulate Integer. The number of initial dates to accumulate
+##'   if `missing` or a column name is set to `"accumulate"`. This number of
+##'   dates is added to the beginning of the data set to be accumulated onto the
+##'   first data point. This is useful, for example, for modelling weekly
+##'   incidence data, in which case this should be set to 7. If accumulating and
+##'   the first data point is one to accumulate and this is not set, then that
+##'   data point will be removed with a warning.
+##' @param obs_column Character. If given, only the column specified here will
```
seabbs, line 29: This auto-magic feature seems like it could be dangerous as people might not expect random columns epinow2 doesn't use to do something like this. Is there a reason to not make the default the core set (i.e. `confirm`)?

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##'   be used for checking missingness. This is useful if using a data set that
+##'   has NA values in multiple columns, but only one of them corresponds to
+##'   observations that are to be processed here. By default, all columns will
+##'   be processed and, e.g. `accumulate` will be set if any column is NA in any
+##'   given row.
+##' @return a data.table with an `accumulate` column that indicates whether
+##'   values are accumulated (see the documnetation of the `data` arugment in
```
seabbs, line 36:
```suggestion
##'   values are accumulated (see the documentation of the `data` argument in
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+
+  date_seq <- seq.Date(
+    min(data$date) - initial_add + 1,
+    max(data$date), by = "day"
+  )
+
+  ## mark dates that are present in the data
+  data[, ..present := TRUE]
+  complete_dates <- data.table(date = date_seq)
+  data <- merge(data, complete_dates, by = "date", all.y = TRUE)
```
seabbs, line 81: if people are preparing data from multi-regions will this break? does it matter if it does

R/create.R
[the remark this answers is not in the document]
jamesmbaazam: Alternatively, we could do `as.integer(TRUE %in% accumulate)`.

R/preprocessing.R
seabbs: is the expectation that users will nearly always set this themselves? If yes maybe we throw a message if they don't set it?

inst/stan/estimate_infections.stan
```diff
@@ -139,6 +139,13 @@ transformed parameters {
     }
   }
 
+  // accumulate reports
+  if (any_accumulate) {
+     profile("accumulate") {
+       reports = accumulate_reports(reports, accumulate);
```
seabbs, line 145: What is the interaction with truncation here? Before accumulation happened after truncation and now it happens before. I think it could be either  but perhaps after is slightly more common?

inst/stan/functions/observation_model.stan
```diff
@@ -102,52 +102,54 @@ void truncation_lp(array[] real truncation_mean, array[] real truncation_sd,
...
+ * Accumulate reports according to a binary flag at each time point
+ *
+ * This function accumulates reports according to a binary flag at each time point.
+ *
+ * @param reports Vector of expected reports.
+ * @param accumulate Array of integers indicating, for each time point, whether to accumulate or not.
+ *
+ * @return A vector of accumulated reports.
+ */
+vector accumulate_reports(vector reports, array[] int accumulate) {
```
seabbs, line 144: This is incredibly pedantic but I think this function should go above the truncation one match in model usage

seabbs (changes requested):
This is really nice. I have a few small comments. At a high level I think we are missing examples of this in the wider docs that would be nice to add but could be done in another PR. However, as we are missing these I think we need to be careful here to make sure this does actually work as we are the only checks (i.e no correctness unit tests or figure to look at that I am aware of)

NEWS.md
```diff
@@ -1,5 +1,20 @@
 # EpiNow2 (development version)
 
+## Model changes
+
+- The models now supports more complex accumulation patterns, as well as mixtures of accumulation and missingness via the new `fill_missing()` function and a logical `accumulate` column that can be supplied with the data. By @sbfnk in # and reviewed by @.
```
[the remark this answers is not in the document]
seabbs: I agree but we do need to be careful as we don't support some aggregation schemes that are not accumulation (i.e non-consecutive aggregation)

R/create.R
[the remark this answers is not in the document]
seabbs: wait can't we just do any directly?

R/estimate_infections.R
```diff
@@ -17,8 +17,13 @@
 #' for an example of using `estimate_infections` within the `epinow` wrapper to
 #' estimate Rt for Covid-19 in a country from the ECDC data source.
 #'
-#' @param data A `<data.frame>` of confirmed cases (confirm) by date
-#' (date). `confirm` must be numeric and `date` must be in date format.
+#' @param data A `<data.frame>` of confirmed cases (confirm) by date (date).
+#' `confirm` must be numeric and `date` must be in date format. Optionally
+#' this can also have a logical `accumulate` column which indicates whether
+#' data should be added to the next data point. This is useful when modelling
```
jamesmbaazam, line 23: Can we reword this to something simpler for example, 
```suggestion
#' modelled values should correspond to the same reporting pattern as the observed data. This is useful when modelling
```

jamesmbaazam, line 23:
```suggestion
#' `confirm` must be numeric and `date` must be in date format. Optionally
#' this can also have a logical `accumulate` column which indicates which time points
#'  should be aggregated/accumulated. This is useful when modelling
```

R/estimate_infections.R
```diff
@@ -176,10 +181,11 @@ estimate_infections <- function(data,
     data = dirty_reported_cases,
     cols_to_check = c("date", "confirm")
   )
+  # Fill misisng dates
```
jamesmbaazam, line 184:
```suggestion
  # Fill missing dates
```

NEWS.md
```diff
@@ -1,5 +1,20 @@
...
+
+  ```r
+  # Deprecated
+  data |>
+    estimate_infections(obs_opts(na = "accumulate"))
+
+  # Recommended workflow e.g. for weekly incidence data
```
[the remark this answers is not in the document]
sbfnk: Not in this case (as `na == "missing"` by default) - will try to clarify in the docs as you suggest below.

R/create.R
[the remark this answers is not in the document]
sbfnk:
```suggestion
    any_accumulate = as.integer(any(accumulate)),
```

R/estimate_infections.R
sbfnk: will rephrase

NEWS.md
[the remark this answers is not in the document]
sbfnk: Sounds like a good idea to mention here if we think it's something people search for but in other places we should decide on one term (and make sure we explain it well / better) as otherwise people might just be confused about whether they mean the same or not.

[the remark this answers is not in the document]
sbfnk:
```suggestion
- The models now supports more complex patterns of aggregating reported cases by accumulating them over multiple time points, as well as mixtures of accumulation and missingness via the new `fill_missing()` function and a logical `accumulate` column that can be supplied with the data. By @sbfnk in #851 and reviewed by @seabbs and @jamesmbaazam..
```

R/preprocessing.R
sbfnk: ok, will try to rephrase / clarify in the function doc.

> I think the thing that confused me most was `missing = "missing"`. Perhaps it should be `missing = "na"`?

Hmm yes that does look weird - the problem is that you could have a pattern like `missing = "missing", na = "accumulate"` which I think would be even more confusing if then we had `missing = "na"`.

I'd suggest to:
- replace the choices with `c("ignore", "accumulate", "zero")` (with ignoring meaning ignore in the likelihood)
- rename the arguments from `missing` and `na` to `missing_dates` and `na_obs`, respectively

R/preprocessing.R
sbfnk: I hadn't considered this. We could add a `by` argument, scoringutils-style?

inst/stan/functions/observation_model.stan
```diff
@@ -102,52 +102,54 @@ void truncation_lp(array[] real truncation_mean, array[] real truncation_sd,
...
+  int ot_h = num_elements(reports); // number of reporting time points modelled
+  vector[ot_h] accumulated_reports = reports;
+  for (i in 1:(ot_h - 1)) {
+    if (accumulate[i]) { // first observation gets ignored when accumulating
+      accumulated_reports[i + 1] += accumulated_reports[i];
```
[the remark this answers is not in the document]
sbfnk: I think this works as expected (because `accumulated_reports` is set to `reports` earlier:

``` r
nreports <- 9

reports <- sample(1:100, nreports)
accumulate <- rep(c(TRUE, TRUE, FALSE), times = nreports / 3)

accumulated_reports <- reports
for (i in 1:(length(reports) - 1)) {
  if (accumulate[i]) {
    accumulated_reports[i + 1] <- accumulated_reports[i + 1] +
      accumulated_reports[i]
  }
}

data.frame(
  accumulate = accumulate,
  reports = reports,
  accumulated_reports = accumulated_reports
)
#>   accumulate reports accumulated_reports
#> 1       TRUE      24                  24
#> 2       TRUE       2                  26
#> 3      FALSE      51                  77
#> 4       TRUE      82                  82
#> 5       TRUE      35                 117
#> 6      FALSE      21                 138
#> 7       TRUE      19                  19
#> 8       TRUE      39                  58
#> 9      FALSE       1                  59
```

<sup>Created on 2024-11-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

sbfnk:
> At a high level I think we are missing examples of this in the wider docs that would be nice to add but could be done in another PR. 

I agree, e.g. #772.

> However, as we are missing these I think we need to be careful here to make sure this does actually work as we are the only checks (i.e no correctness unit tests or figure to look at that I am aware of)

Yes, good point. I can add a snapshot test (which will obviously only help in the future) and I think @jamesmbaazam is looking at testing this in action.

inst/stan/functions/observation_model.stan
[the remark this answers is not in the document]
sbfnk: > Is there a reason to code it like this? 

None apart from my inability to conceive of anything more elegant (noting that #548 remains to be addressed in a future PR). What would you suggest?

> there is nothing stopping you having multiple days in the likelihood as they accumulate

You mean fitting to the cumulative data? Of course it's a bad idea but I wasn't considering actively stopping people from doing so. Perhaps we should. I was thinking of accumulation and use in the likelihood as orthogonal but we could also only evaluate the likelihood if `accumulate == FALSE` AND the data is non-NA, which would be the safer option.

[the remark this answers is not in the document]
seabbs: > You mean fitting to the cumulative data? 

Yes. 

> we could also only evaluate the likelihood if accumulate == FALSE AND the data is non-NA, which would be the safer option.

I thought this was the choice we made in the design doc but I don't feel that strongly because as you may maybe people might have a reason to fit to cumulative data

NEWS.md
```diff
@@ -1,5 +1,20 @@
 # EpiNow2 (development version)
 
+## Model changes
+
+- The models now supports more complex patterns of aggregating reported cases by accumulating them over multiple time points, as well as mixtures of accumulation and missingness via the new `fill_missing()` function and a logical `accumulate` column that can be supplied with the data. By @sbfnk in #851 and reviewed by @seabbs and @jamesmbaazam..
```
jamesmbaazam, line 5:
```suggestion
- The models now support more complex patterns of aggregating reported cases by accumulating them over multiple time points, as well as mixtures of accumulation and missingness via the new `fill_missing()` function and a logical `accumulate` column that can be supplied with the data. By @sbfnk in #851 and reviewed by @seabbs and @jamesmbaazam..
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
+##' Fills missing data in a data set to prepare it for use within the package
```
jamesmbaazam, line 1:
```suggestion
##' Fill missing data in a data set to prepare it for use within the package
```

R/preprocessing.R
jamesmbaazam: > rename the arguments from missing and na to missing_dates and na_obs, respectively

I was coming to suggest this.

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##'
+##' @param data Data frame with a `date` column. The other columns depend on the
+##'   model that the data are to be used, e.g. [estimate_infections()] or
+##'   [estimate_secondary()]. See the documentation there for the expected
+##'   format.
+##' @param missing Character. Options are "missing" (the default), "accumulate"
+##'   and "zero". This determines how missing dates in the data are interpreted.
+##'   If set to "missing", any missing dates in the observation data will be
+##'   interpreted as missing and skipped in the likelihood. If set to
+##'   "accumulate", modelled observations will be accumulated and added to the
```
jamesmbaazam, line 13:
```suggestion
##'   "accumulate", the date will be modelled but accumulated and added to the
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##'   next non-missing data point. This can be used to model incidence data that
+##'   is reported at less than daily intervals. If set to "accumulate", the
```
jamesmbaazam, line 15:
```suggestion
##'   next non-missing data point. This can be used to model incidence data that
##'   is reported at irregular/inconsistent intervals. If set to "accumulate", the
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
+##' Fills missing data in a data set to prepare it for use within the package
+##'
+##' @description `r lifecycle::badge("experimental")`
+##'
+##' @param data Data frame with a `date` column. The other columns depend on the
+##'   model that the data are to be used, e.g. [estimate_infections()] or
+##'   [estimate_secondary()]. See the documentation there for the expected
+##'   format.
+##' @param missing Character. Options are "missing" (the default), "accumulate"
+##'   and "zero". This determines how missing dates in the data are interpreted.
```
jamesmbaazam, line 10:
```suggestion
##'   and "zero". This determines how implicit missingness is treated, i.e., how missing dates in the data are interpreted.
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##'   If set to "missing", any missing dates in the observation data will be
+##'   interpreted as missing and skipped in the likelihood. If set to
+##'   "accumulate", modelled observations will be accumulated and added to the
+##'   next non-missing data point. This can be used to model incidence data that
+##'   is reported at less than daily intervals. If set to "accumulate", the
+##'   first data point is not included in the likelihood (unless `initial` is
+##'   set to a non-zero value) but used only to reset modelled observations to
+##'   zero. If "zero" then all observations on missing dates will be assumed to
+##'   be of value 0.
+##' @param na Character. How to process dates that have NA values. The options
```
jamesmbaazam, line 20:
```suggestion
##' @param na Character. This determines how explicit missing values are treated, i.e., dates that have `NA` observations. The options
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##'   available are the same ones as for the `missing` argument.
+##' @param initial_accumulate Integer. The number of initial dates to accumulate
+##'   if `missing` or a column name is set to `"accumulate"`. This number of
+##'   dates is added to the beginning of the data set to be accumulated onto the
+##'   first data point. This is useful, for example, for modelling weekly
```
jamesmbaazam, line 25:
```suggestion
##' @param initial_accumulate Integer. Only used if `missing = "accumulate"` or the input data has a column named `"accumulate"`. This is the number of
##'   dates to model and accumulate to the first date in the input data. This is useful, for example, for modelling weekly
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##'   incidence data, in which case this should be set to 7. If accumulating and
+##'   the first data point is one to accumulate and this is not set, then that
```
jamesmbaazam, line 27:
```suggestion
##'   the first data point is one to accumulate and `initial_accumulate` is not set, then that
```

NEWS.md
seabbs: as implemented it is literally accumulation and you can have multiple accumulate elements in the likelihood which would not be the same as aggregation so if we stick with that I think we need to stick with the accumulate naming

R/preprocessing.R
seabbs: I think the action I would like here is to give it the confirm default as suggested

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+                         obs_column) {
+  assert_data_frame(data)
+  assert_names(names(data), must.include = "date", disjunct.from = "accumulate")
+  if (!missing(obs_column)) {
+    assert_names(names(data), must.include = obs_column)
+  } else {
+    obs_column <- setdiff(colnames(data), "date")
+  }
+  assert_date(data$date, any.missing = FALSE)
+
```
jamesmbaazam, line 64: assert_character(missing)
assert_character(na)
assert_character(obs_column)
assert_integer(initial_accumulate, lower = 0)

jamesmbaazam, line 64: Need to make sure `initial_accumulate` is never below 0.
```suggestion
assert_character(missing)
assert_character(na)
assert_character(obs_column)
assert_integer(initial_accumulate, lower = 0)
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+  data[, accumulate := FALSE]
+  for (col in obs_column) {
+    if (missing == "accumulate") {
+      data[is.na(..present), accumulate := TRUE]
+    } else if (missing == "zero") {
+      data[is.na(..present), paste(col) := 0]
+    }
+  }
+  if (missing(initial_accumulate) &&
+      !data$accumulate[1] &&
```
jamesmbaazam, line 92:
```suggestion
      isFALSE(data$accumulate[1]) &&
```

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##' ## limit to dates once a week
+##' cases <- cases[seq(7, nrow(cases), 7)]
+##' ## set the second observation to missing
+##' cases[2, confirm := NA]
+##' ## fill missing data
+##' fill_missing(cases, missing = "accumulate", initial_accumulate = 7)
+fill_missing <- function(data,
+                         missing = c("missing", "accumulate", "zero"),
+                         na = c("missing", "accumulate", "zero"),
+                         initial_accumulate,
```
jamesmbaazam, line 54: I'm a little confused. I set `initial_accumulate = 1` and expected 2020-02-28  to be set to TRUE and an observation added at 2020-02-27. Is this expected?

```r
library(data.table)
cases <- data.table::copy(example_confirmed)
## calculate weekly sum
cases[, confirm := data.table::frollsum(confirm, 7)]
## limit to dates once a week
cases <- cases[seq(7, nrow(cases), 7)]
## set the second observation to missing
cases[2, confirm := NA]
## fill missing data
fill_missing(cases, missing = "accumulate", initial_accumulate = 1) |>
  head(14)

Key: <date>
          date confirm accumulate
        <Date>   <num>     <lgcl>
 1: 2020-02-28     647      FALSE
 2: 2020-02-29      NA       TRUE
 3: 2020-03-01      NA       TRUE
  
```

jamesmbaazam: Also, what should be the expected behavior if `intial_accumulate` is set to 0? Should it be the same as when it is not set?

sbfnk: The idea is that the default is to accumulate one observation (`initial_accumulate = 1`). 

> 2020-02-28 to be set to TRUE and an observation added at 2020-02-27. 

This to me sounds like two observations are being accumulated (`initial_accumulate = 2`)

> Also, what should be the expected behavior if initial_accumulate is set to 0? Should it be the same as when it is not set?

We should probably make it have a minimum of 1.

Happy to hear of alternative/clearer suggestions for wording this.

jamesmbaazam: > We should probably make it have a minimum of 1.

Agreed. That can be added in the assertion on `initial_accumulate` I added below.

R/preprocessing.R
```diff
@@ -0,0 +1,152 @@
...
+##' cases <- data.table::copy(example_confirmed)
+##' ## calculate weekly sum
+##' cases[, confirm := data.table::frollsum(confirm, 7)]
+##' ## limit to dates once a week
+##' cases <- cases[seq(7, nrow(cases), 7)]
+##' ## set the second observation to missing
+##' cases[2, confirm := NA]
+##' ## fill missing data
+##' fill_missing(cases, missing = "accumulate", initial_accumulate = 7)
+fill_missing <- function(data,
```
jamesmbaazam, line 51: This function errors if the data already has an `accumulate` column (from the assertion for column names). 

Should it rather throw a message and do nothing if it finds an `accumulate` column?

```r
library(data.table)
cases <- data.table::copy(example_confirmed)
## calculate weekly sum
cases[, confirm := data.table::frollsum(confirm, 7)]
## limit to dates once a week
cases <- cases[seq(7, nrow(cases), 7)]
## set the second observation to missing
cases[2, confirm := NA][, accumulate:=TRUE]
## fill missing data
fill_missing(cases, missing = "accumulate", initial_accumulate = 1, obs_column = "confirm")

Error in fill_missing(cases, missing = "accumulate", initial_accumulate = 1,  : 
  Assertion on 'names(data)' failed: Names must be disjunct from {'accumulate'}, but has elements {'accumulate'}.
```

NEWS.md
sbfnk:
```suggestion
- The models now support more complex patterns of aggregating reported cases by accumulating them over multiple time points, as well as mixtures of accumulation and missingness via the new `fill_missing()` function and a logical `accumulate` column that can be supplied with the data. By @sbfnk in #851 and reviewed by @seabbs and @jamesmbaazam..
```

R/preprocessing.R
sbfnk: What would be the rationale for this? I don't really see when we'd want to fail silently if someone calls `fill_missing()` and we're not doing anything because an `accumulate` column already exists.

R/preprocessing.R
sbfnk: I've updated the documentation in the attempt to make it clearer

R/preprocessing.R
sbfnk: I've tried to rephrase slightly differently in an attempt to make it clearer.

R/preprocessing.R
sbfnk: Have tried to rephrase.

R/preprocessing.R
sbfnk: I think if doing this we'd have to define implicit vs. explicit missingness somewhere - do you think it would make it clearer? I have tried to rephrase without introducing new definitions but please comment if you disagree with the approach.

R/preprocessing.R
sbfnk: Have tried to rephrase

sbfnk:
Thanks for the comments/suggestions - I've updated the argument names as suggested and did a pass through the docs in an attempt to make them clearer.

R/preprocessing.R
sbfnk: Alternatively perhaps we could have a more informative error message?

sbfnk: Alternatively perhaps we could have a more informative error message?

seabbs: I am pretty happy with it failing as `fill_missing` is optional. We could make it more informative but tbh I am happy

inst/stan/estimate_infections.stan
```diff
@@ -156,6 +156,13 @@ transformed parameters {
   } else {
     obs_reports = reports[1:ot];
   }
+
+  // accumulate reports
+  if (any_accumulate) {
```
seabbs, line 161: Just checking the difference between estimate_infections and truncation is intentional (i.e one accumulates after truncation and the other before)

seabbs (dismissed):
This looks good to me and happy for it to be merged. I have a open comment with a last check about the interaction between accumulation and truncation (which I think I already raised and so is likely easy to dismiss)

seabbs:
CI failures are unrelated to this PR (trying and failing to download R) so I am happy to ignore them

R/preprocessing.R
jamesmbaazam: > Alternatively perhaps we could have a more informative error message?

Yes, something like "data already has an "accumulate" column" is more friendly.

seabbs (approved):

R/preprocessing.R
sbfnk: I've decided against this as really this applies to all the `{checkmate}` checks and we don't want to change them all. I added some more clarifying documentation instead on the topic.

sbfnk:
Thanks all for the comments/reviews - merging now as failed checks are unrelated.

PR #855, opened 2024-11-15
not for merging: Explore divergent transitions

sbfnk:
I've tried to explore the source of divergent transitions in the light of the discussion in https://github.com/epiforecasts/EpiNow2/pull/847
In order to do so I fixed parameters one-by-one to see the impact on divergent transition. This also gave an opportunity to experiment with modifying stan models from R.

A report is at 
https://github.com/epiforecasts/EpiNow2/blob/explore-divergences/inst/dev/divergences.md

As found before, constraining the magnitude of the GP (alpha) seems to have the greatest impact on being able to fit the model. Even fixing the lengthscale the divergent transitions remain suggesting that alpha may be poorly identified in itself.

I don't think we need this to be added to the repo but if we think the general approach is of interest we could perhaps make it a blog post e.g. on the epiforecasts web site.

### DIFF
--- a/inst/dev/divergences.Rmd
+++ b/inst/dev/divergences.Rmd
@@ -1,0 +1,142 @@
+---
+output: github_document
+---
+
+# Examining divergent transitions in EpiNow2
+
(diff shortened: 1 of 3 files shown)


seabbs:
Are there plans for using this or can we close?

sbfnk:
I think it would be good to re-do this after #890 has been merged, as it will be possible to do it then without editing the stan model.

PR #840, opened 2024-10-29
restore alpha prior SD to 0.05

sbfnk:
## Description

This PR closes #834.

Still happy to review/reconsider if other values might work better.

I think I've got the squares / square roots right (and fixed them in the old news item) but may have got confused.

## Initial submission checklist

- [X] My PR is based on a package issue and I have explicitly linked it.
- [ ] I have tested my changes locally (using `devtools::test()` and `devtools::check()`).
- [ ] I have added or updated unit tests where necessary.
- [X] I have updated the documentation if required and rebuilt docs if yes (using `devtools::document()`).
- [X] I have followed the established coding standards (and checked using `lintr::lint_package()`).
- [X] I have added a news item linked to this PR.

## After the initial Pull Request 

- [ ] I have reviewed Checks for this PR and addressed any issues as far as I am able.

PR #856, opened 2024-11-15
reduce alpha prior sd again

sbfnk:
Suggesting to revert the recent change in the prior sd of alpha (https://github.com/epiforecasts/EpiNow2/pull/840) in the light of the issues seen in https://github.com/epiforecasts/EpiNow2/pull/847 and the exploration in https://github.com/epiforecasts/EpiNow2/pull/855.


sbfnk:
This will re-open the issue in https://github.com/epiverse-trace/howto/issues/81 which will have to (and can) be addressed in some other way.

jamesmbaazam:
> This will re-open the issue in [epiverse-trace/howto#81](https://github.com/epiverse-trace/howto/issues/81) which will have to (and can) be addressed in some other way.

Isn't the solution to use the prior in the previous version? In the how-to, the concern is not about divergences but how to set up a model.

jamesmbaazam (approved):
LGTM.

sbfnk:
> > This will re-open the issue in [epiverse-trace/howto#81](https://github.com/epiverse-trace/howto/issues/81) which will have to (and can) be addressed in some other way.
> 
> Isn't the solution to use the prior in the previous version? In the how-to, the concern is not about divergences but how to set up a model.

The solution is indeed to either use `prior = "infections"` or manually set `alpha_sd` in the howto.

PR #871, opened 2024-11-26
Change all priors to use `<dist_spec>`

### LINKED ISSUE #525: Change all priors to new dist interface, not just delays

sbfnk:
## Description

This PR closes #525 by changing all parameters to use the new interface.

In doing so and in order to keep this manageable the internals of handling parameters had to be updated a fair bit. This is now done similarly to delays in parameter arrays rather than named parameters. In principle this should make it easier to extend the model later with more parameters / probability distributions / time-varying parameters etc.

Things that could be added later fairly easy if deemed useful:
1. switch delay distribution parameters to use the new stan setup for parameters (this might simplify code so probably a good idea)
2. allow uncertain parameters of prior distributions other than delays

From tests it seems functionality has been preserved but it would be good for people to check if this is actually an improvement.

One thing to note is that now all parameters are initialised from (truncated) standard normal distributions so will have to monitor if this leads to any issues.


seabbs (commented):
Wow - this is a lot. On the face of it looks good but going to need some time to digest

sbfnk:
One thing to note is that the lengthscale prior is currently not included in this, because of the extra logic in
https://github.com/epiforecasts/EpiNow2/blob/2f27a3f99fa6d1823818dd953bbb33cd44008a9e/inst/stan/functions/gaussian_process.stan#L201
It might be worth deciding on a single shape of the prior and then this could also be specified as a `<dist_spec>` (though this currently doesn't support specifying a minimum, which would either have to be dropped or functionality added).

sbfnk:
Related to that merging this would enable a more comprehensive exploration of parameter choices and identifiability than what's done in #855 (not requiring editing the stan model) as we can fix parameters and/or try out different distributions with different parameters.

seabbs:
>  which would either have to be dropped or functionality added

I think we would need to add this as from memory it really isn't identifiable without setting some bounds

sbfnk:
> > which would either have to be dropped or functionality added
> 
> I think we would need to add this as from memory it really isn't identifiable without setting some bounds

To be explored in more detail in a future PR then as adding capability to specify a lower bound would require quite substantial changes.

seabbs:
I've got thursday pencilled in to go through this in detail

NEWS.md
```diff
@@ -15,8 +15,9 @@
     estimate_infections()
   ```
   
-  - A bug was fixed where the initial growth was never estimated (i.e. the prior mean was always zero). By @sbfnk in #853 and reviewed by @seabbs.
-  - A bug was fixed where an internal function for applying a default cdf cutoff failed due to a difference a vector length issue. By @jamesmbaazam in #858 and reviewed by @sbfnk.
+- A bug was fixed where the initial growth was never estimated (i.e. the prior mean was always zero). By @sbfnk in #853 and reviewed by @seabbs.
+- A bug was fixed where an internal function for applying a default cdf cutoff failed due to a difference a vector length issue. By @jamesmbaazam in #858 and reviewed by @sbfnk.
+- All parameters have been changed to the new parameter interface. By @sbfnk in # and reviewed by @.
```
seabbs, line 20:
```suggestion
- All parameters have been changed to the new parameter interface. By @sbfnk in #871 and reviewed by @seabbs.
```

R/create.R
```diff
@@ -684,20 +672,17 @@ create_initial_conditions <- function(data) {
...
     if (data$week_effect > 0) {
       out$day_of_week_simplex <- array(
         rep(1 / data$week_effect, data$week_effect)
       )
     }
+    out$params <- array(truncnorm::rtruncnorm(
+      data$n_params_variable,
+      a = data$params_lower,
+      b = data$params_upper,
+      mean = 0, sd = 1
```
seabbs, line 684: why is this initialising at 0, 1 and not the mean and sd as before?

R/opts.R
```diff
@@ -339,19 +340,18 @@ trunc_opts <- function(dist = Fixed(0), default_cdf_cutoff = 0.001,
 #' rt_opts()
 #'
 #' # add a custom length scale
-#' rt_opts(prior = list(mean = 2, sd = 1))
+#' rt_opts(prior = Normal(mean = 2, sd = 1))
 #'
 #' # add a weekly random walk
 #' rt_opts(rw = 7)
-rt_opts <- function(prior = list(mean = 1, sd = 1),
+rt_opts <- function(prior = LogNormal(mean = 1, sd = 1),
```
seabbs, line 347: Flagging for me to circle back to - should this really be lognormal now

seabbs: it looks like this is the case in all vignettes. I think probably this should still be Normal

seabbs (approved):
This is great I think. I was a bit sceptical as I thought it would make the models hard to follow but tbh if anything its the versus due to the reduction in control code. 

I noted that the Rt prior is treated a little differently and had a few comments on that. At a high level I am happy with it as is and even think its good moving to a normal prior where that has happened (its unclear how intentional that was but I am okay with it). 

I don't see any major issues. It looks like it should be relatively easy to support user set lower bounds (and hence also transition over the GP stuff. I noted it would be nice to give users an R helper so they can find the posterior for parameters they are setting but I think this can be a future issue. 

I see the stan benchmarks are failing but otherwise LGTM

Note I updated the vignettes in order to look at the git diff. All looks good to me. Happy to revert or keep this as the update.

seabbs:
Wow I have caused a lot of conflicts

R/create.R
sbfnk: Because we don't have a way of determining the mean/sd once we've done the conversion to stan code. My suggestion was 

> will have to monitor if this leads to any issues.

but if we're not happy with that we could add a function for going back from data to `<dist_spec>` and then use its functionality to get the mean and sd.

sbfnk: Because we don't have a way of determining the mean/sd once we've done the conversion to stan code. My suggestion was 

> will have to monitor if this leads to any issues.

but if we're not happy with that we could add a function for going back from data to `<dist_spec>` and then use its functionality to get the mean and sd.

vignettes/estimate_infections_options.Rmd
```diff
@@ -182,22 +177,23 @@ def <- estimate_infections(reported_cases,
...
+#> 4:                   Rate of growth -0.027 (-0.099 -- 0.047)
+#> 5:     Doubling/halving time (days)           -25 (15 -- -7)
 # elapsed time (in seconds)
 get_elapsed_time(def$fit)
 #>         warmup sample
-#> chain:1 39.295 14.837
-#> chain:2 30.967 17.867
-#> chain:3 19.880 20.187
-#> chain:4 30.845 17.029
+#> chain:1 23.430 14.301
```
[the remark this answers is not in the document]
sbfnk: I think I would prefer to go lognormal throughout. I did not intend to make a change.

vignettes/estimate_infections_workflow.Rmd
```diff
@@ -301,7 +296,8 @@ def <- estimate_infections(
   delays = delay_opts(delay),
   rt = rt_opts(prior = rt_prior)
 )
-#> Warning: There were 6 divergent transitions after warmup. See
+#> DEBUG [2024-12-05 15:12:16] estimate_infections: Running in exact mode for 2000 samples (across 4 chains each with a warm up of 250 iterations each) and 147 time steps of which 7 are a forecast
```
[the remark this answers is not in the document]
sbfnk: did you run this in an interactive session? Then you get DEBUGs by defeault.

sbfnk:
Sorry for the confusion on `rt` and normal vs. lognormal.

My reasoning was: so far we asked users for the mean and SD of a lognormal distribution which we had as logR ~ normal. In the new set up we could still ask for a prior on log R but that seems unnatural if people might want different distributions. Instead I changed the prior to be a prior on R and then defaulting to lognormal with the same mean and SD as before. If I'm not mistaken this should be no change to before. Where it is normal (in the vignettes?) this was by accident and should be LogNormal.

sbfnk:
> Wow I have caused a lot of conflicts

Might be easiest to just reset and then run the action post merging

seabbs (approved):
I've reverted the vignette updating (in a lazy way but will get squashed out) and am happy for this to be merged either with or without the updated Rt priors in the vignettes and examples.

seabbs:
I've reset the cmdstan actions cache to check if that is the issue (really need to get around to setting up a cache timer in the install cmdstan action).

sbfnk:
I've switched all priors to LogNormal now. Just needs someone to re-approve.

seabbs (approved):
LGTM

PR #888, opened 2024-12-11
Split prior infections and growth calculation into separate function

### LINKED ISSUE #887: Split up `create_stan_data()`?
I've been reviewing #871 to bring myself up to date with developments while I was away and I realise that `create_stan_data()` is getting quite long and difficult to parse. For one, it's combining two concerns: what I'd call static data creation and dynamic modelling of the initial growth. I would suggest pulling out the modeling of initial growth part (https://github.com/epiforecasts/EpiNow2/blob/main/R/create.R#L504-L524) as it will be easier to maintain and improve if need be and to point to. It only depends on `data`, `seeding_time`, and `horizon` and returns the `prior_infections` and `prior_growth`, making it quite focused. This is more of a maintenance issue than an enhancement and can be categorised as low priority.

Thoughts?

jamesmbaazam:
## Description

This PR closes #887.


R/create.R
```diff
@@ -501,28 +501,10 @@ create_stan_data <- function(data, seeding_time,
...
-      stan_data$prior_growth$coefficients[2]
-    )
-  } else {
-    stan_data$prior_growth <- 0
-  }
-
+  # Calculate prior infections and growth
+  prior_estimates <- calculate_prior_estimates(cases, seeding_time)
+  stan_data$prior_infections <- prior_estimates$prior_infections
+  stan_data$prior_growth <- prior_estimates$prior_growth
```
sbfnk, line 507:
```suggestion
  stan_data <- c(stan_data, 
    calculate_prior_estimates(cases, seeding_time)
  )
```

sbfnk (changes requested):
Cool!

Shall we call this something more specific than `calculate_prior_estimates()` (as could be confused with the other prior distributions)? Perhaps `estimate_early_dynamics()` or something like that?

Also, this is only used by `create_stan_data()` -- should it live in the same file for easier findability (not sure and open to other opinions).

jamesmbaazam:
> Cool!
> 
> Shall we call this something more specific than `calculate_prior_estimates()` (as could be confused with the other prior distributions)? Perhaps `estimate_early_dynamics()` or something like that?

I like `estimate_early_dynamics()`.
> 
> Also, this is only used by `create_stan_data()` -- should it live in the same file for easier findability (not sure and open to other opinions).

I didn't want to ruin the aesthetics of all the functions in that script being called "create_*()" but it makes sense to live there and probably right above the function that calls it.

R/create.R
jamesmbaazam: Good suggestion but I'm a little worried about readability here. Do we want the "reviewer" to know immediately what data is being created?

jamesmbaazam: I've made the change.

sbfnk (approved):

PR #890, opened 2024-12-12
move lengthscale prior to `dist_spec`

### LINKED ISSUE #878: Switch length scale prior to new interface
One thing to consider then would be if we want to scale the length scale parameter with the length of the time series so that a common prior can be used across data sets fitted to.

sbfnk:
## Description

This PR closes #878.

In order for there to be a prior choice that applies across applications it rescales the lengthscale with half the length of the time series as done in [Riutort-Mayol et al.](https://arxiv.org/abs/2004.11408) It also fixes the minimum to zero (for now) which was previously the default but changeable.

Todo before merging:
- [x] test with different data sets that this works and doesn't affect our ability to fit the model
- [x] update documentation

Todo after merging (future PRs):
- [ ] re-do the exploration in #855 
- [ ] maybe update the parameter names in places? We've got a bit of a mixture of rho/ls so need to decide in which places we want this to be the 


jamesmbaazam:
@seabbs Do you want to take this one?

sbfnk:
Testing this on a few data sets I encountered no issues but an eye would have to be kept on this in the near future.

seabbs:
@sbfnk can you plot the change in the prior here if you get a chance?

NEWS.md
```diff
@@ -17,7 +17,8 @@
   
 - A bug was fixed where the initial growth was never estimated (i.e. the prior mean was always zero). By @sbfnk in #853 and reviewed by @seabbs.
 - A bug was fixed where an internal function for applying a default cdf cutoff failed due to a difference a vector length issue. By @jamesmbaazam in #858 and reviewed by @sbfnk.
-- All parameters have been changed to the new parameter interface. By @sbfnk in #871 and reviewed by @seabbs.
+- All parameters have been changed to the new parameter interface. By @sbfnk in #871 and #890 and reviewed by @seabbs.
+- The Gaussian Process lengthscale is now scaled by half the length of the time series. By @sbfnk in #890 and reviewed by #
```
seabbs, line 21: This doesn't say what it was beefore and I think it should. I am also really not convinced this is a sensible choice as in most cases we would think about the lengthscale as being driven by the epi process and not the amount of data being fit (it will also really worse the difference we see between longer and shorter fits to data as the prior will change a lot)

R/opts.R
```diff
@@ -533,6 +534,7 @@ gp_opts <- function(basis_prop = 0.2,
                     ls_sd = 7,
                     ls_min = 0,
                     ls_max = 60,
+                    ls = Normal(mean = 0.5, sd = 0.1, max = 1),
```
seabbs, line 537: Noting that from a scientific view point I don't think this is a very good prior or very easy for people to reason on.

seabbs (commented):
From a technical stand point I think this is correct. 

From a scientific standpoint I have a lot of reservations about: 

1. Setting this prior on the relative scale
2. Setting this prior to be half the time series
3. Changing the default max from 60 to 1
4. Moving to normal from lognormal


1. How do people reason on this prior. It means as they fit to new data I would expect they need to constantly update the prior - that feels really quite odd to me
2. This is really quite a long lengthscale in many settings and I would expect it to negatively impact identifiability and performance during long periods of constant Rt (i.e locking in constant Rt beyond our prior expectations)
3. Setting the max to the length of the time series as the default seems likely to lead to non-identifiability to me (support for a range of very nearly the same lengthscales)
4. This has moved the prior to be normal vs lognormal unless I am missing a transformation. If there is a transformation what does this mean for the inverse gamma option

I know you have done some testing but I don't really believe our testing is deep enough to catch above. 

My preference would be we update interface but keep the current defaults. I realise this could be a problem as it means we need to rescale in the model (I think)

seabbs:
> Just noting (not taking away from your point) that it's what is done in the Riutort-Mayol paper.

Yup, I intentionally didn't implement it like this at the time. My view then was that they weren't really trying to implement something for reuse but just for their experiments and so they might not have thought about the implications of use across rolling periods and different datasets etc.

> These could be easily addressed by changing the exact prior (which I mean to choose to roughly mirror the current prior with the examples)

Yes agree

> - just would be a bit tricky as we'd have to implement transformation by multiplying with a constant for all the distributions used. 

Yes, agree. Its not ideal. Wouldn't it be divide by a constant? The reason I said in model is that this makes it much more flexible (as some prior distributions might be hard to rescale). If we wanted a quick fix this would also be much easier (though at the cost of some efficiency).

seabbs (approved):
Nice looks good to me.

PR #903, opened 2024-12-19
improve early calculation

### LINKED ISSUE #902: Recovery of frac_obs / priors for initial infections and growth
I have noticed that in the synthetic recovery code
https://github.com/epiforecasts/EpiNow2/blob/dbb6536262e68b32a60dab29e3af731b11b1875e/inst/dev/recover-synthetic/rt.R#L14
recovery of the `frac_obs` parameter is pretty poor. The prior is Normal(0.1, 0.025) and this should not really be informed by the data but the posterior is centred around 0.01, i.e. an order of magnitude smaller.

The initial infections and growth rates also end up pretty far from the prior values so there might be some interaction going on between the three -- especially as I don't think we're correcting for fraction observed when estimating initial infections from the data whereas this should probably be factored in.

sbfnk:
## Description

This PR closes #902 by
- changing the calculation of `prior_infections` to be based on the intercept of the linear model instead of mean of first week, if available
- fitting the initial model for `seeding_time` if greater than 7 (as that is the period for which exponential growth is assumed in the model)
- dividing `initial_infections` by `frac_obs` before modelling initial exponential growth
- changing the prior for `initial_infections` to be approximately Poisson

As a result we've go from

image

to

image

in the example mentioned in the Issue - see https://github.com/epiforecasts/EpiNow2/blob/dbb6536262e68b32a60dab29e3af731b11b1875e/inst/dev/recover-synthetic/rt.R#L14
More generally the values of `initial_growth` / `initial_infections` no longer seem to affect `frac_obs`, at the expense of (expected) correlation between the initial estimates.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
 - All parameters have been changed to the new parameter interface. By @sbfnk in #871 and #890 and reviewed by @seabbs.
 - The Gaussian Process lengthscale is now scaled internally by half the length of the time series. By @sbfnk in #890 and reviewed by #seabbs.
 - A bug was fixed where `plot.dist_spec()` wasn't throwing an informative error due to an incomplete check for the max of the specified delay. By @jamesmbaazam in #858 and reviewed by @.
-- Updated the early dynamics calculation to use the full linear model if available. Also changd the prior for initial infections to be approximately Poisson. By @sbfnk in #903 and reviewed by @seabbs and @SamuelBrand1 
+- Updated the early dynamics calculation to use the full linear model if available. Also changed the prior for initial infections to be approximately Poisson and the initial growth rate to the point estimate of the initial growth rate scaled linearly by the estimated initial infections term. By @sbfnk in #903 and reviewed by @seabbs and @SamuelBrand1 
 
 ## Package changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,6 +20,7 @@
 - All parameters have been changed to the new parameter interface. By @sbfnk in #871 and #890 and reviewed by @seabbs.
 - The Gaussian Process lengthscale is now scaled internally by half the length of the time series. By @sbfnk in #890 and reviewed by #seabbs.
 - A bug was fixed where `plot.dist_spec()` wasn't throwing an informative error due to an incomplete check for the max of the specified delay. By @jamesmbaazam in #858 and reviewed by @.
+- Updated the early dynamics calculation to use the full linear model if available. Also changd the prior for initial infections to be approximately Poisson. By @sbfnk in # and reviewed by
 
 ## Package changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -20,7 +20,7 @@
 - All parameters have been changed to the new parameter interface. By @sbfnk in #871 and #890 and reviewed by @seabbs.
 - The Gaussian Process lengthscale is now scaled internally by half the length of the time series. By @sbfnk in #890 and reviewed by #seabbs.
 - A bug was fixed where `plot.dist_spec()` wasn't throwing an informative error due to an incomplete check for the max of the specified delay. By @jamesmbaazam in #858 and reviewed by @.
-- Updated the early dynamics calculation to use the full linear model if available. Also changd the prior for initial infections to be approximately Poisson. By @sbfnk in # and reviewed by
+- Updated the early dynamics calculation to use the full linear model if available. Also changd the prior for initial infections to be approximately Poisson. By @sbfnk in #903 and reviewed by @seabbs and @SamuelBrand1 
 
 ## Package changes
 
--- a/inst/stan/estimate_infections.stan
+++ b/inst/stan/estimate_infections.stan
@@ -46,7 +46,6 @@ parameters {
   vector[fixed ? 0 : gp_type == 1 ? 2*M : M] eta;               // unconstrained noise
   // Rt
   array[estimate_r] real initial_infections;    // seed infections
-  array[estimate_r && seeding_time > 1 ? 1 : 0] real initial_growth; // seed growth rate
   array[bp_n > 0 ? 1 : 0] real<lower = 0> bp_sd; // standard deviation of breakpoint effect
   vector[bp_n] bp_effects;                   // Rt breakpoint effects
   // observation model
@@ -61,6 +60,12 @@ transformed parameters {
   vector[ot_h] reports;                                     // estimated reported cases
   vector[ot] obs_reports;                                   // observed estimated reported cases
   vector[estimate_r * (delay_type_max[gt_id] + 1)] gt_rev_pmf;
+  array[estimate_r && seeding_time > 1 ? 1 : 0] real initial_growth; // seed growth rate
+
+  if (num_elements(initial_growth) > 0) {
+    initial_growth[1] = prior_growth +
+      (prior_infections - initial_infections[1]) / seeding_time;
+  }
 
   // GP in noise - spectral densities
   profile("update gp") {
@@ -205,8 +210,7 @@ model {
     // priors on Rt
     profile("rt lp") {
       rt_lp(
-        initial_infections, initial_growth, bp_effects, bp_sd, bp_n,
-        seeding_time, prior_infections, prior_growth
+        initial_infections, bp_effects, bp_sd, bp_n, prior_infections
       );
     }
   }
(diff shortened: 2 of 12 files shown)


seabbs (approved):
This looks really great. Not quite understanding the CI fail for as CRAN (requiring future) and haven't checked out the benchmarks. Really good this has solved the correlation we were seeing

noting my preference would be we further explore using the approach from @samuelbrand1 used in `EpiAware` where we solve for the implied growth rate based on the generation time and initial Rt estimate. I can knock something up for this as its very similar to the stuff we needed to do in primarycensored (i.e its numerically solved integral)

seabbs:
Ah maybe touchstone is just having a fail party everywhere?

sbfnk:
> noting my preference would be we further explore using the approach from @SamuelBrand1 used in `EpiAware` where we solve for the implied growth rate based on the generation time and initial Rt estimate.

Oh yes that would be great!

SamuelBrand1:
> This looks really great. Not quite understanding the CI fail for as CRAN (requiring future) and haven't checked out the benchmarks. Really good this has solved the correlation we were seeing
> 
> noting my preference would be we further explore using the approach from @SamuelBrand1 used in `EpiAware` where we solve for the implied growth rate based on the generation time and initial Rt estimate. I can knock something up for this (it will look something like this: [CDCgov/Rt-without-renewal@`0d4095e`/EpiAware/src/EpiInfModels/utils.jl#L44](https://github.com/CDCgov/Rt-without-renewal/blob/0d4095ecf5d67b599b26a8822679adebb40b164a/EpiAware/src/EpiInfModels/utils.jl#L44) but I will probably use the [algebraic system solver](https://mc-stan.org/docs/stan-users-guide/algebraic-equations.html) for simplicity) in a future PR

Yeah, I did it this way because I didn't want to worry about AD through a zero finder so hand rolled a few steps of a Newton solver. Not optimal, I like the look of the algebraic system solver.

sbfnk:
Hmm, looks like we get a slowdown and fitting failure(s). This may need more exploration.

sbfnk:
I have removed estimation of initial growth and instead made this dependent on initial infections, given the strong correlation in these two parameters seen in the pairs plot above. This seems to yield a substantial speedup whilst retaining the prior exploration of `frac_obs` as per the linked issue.

The pairs plot looks very similar to the above (just that this time initial_growth is not estimated independently) making me think that this is preferable.

image

inst/stan/estimate_infections.stan
```diff
@@ -62,6 +61,12 @@ transformed parameters {
   vector[ot_h] reports;                                     // estimated reported cases
   vector[ot] obs_reports;                                   // observed estimated reported cases
   vector[estimate_r * (delay_type_max[gt_id] + 1)] gt_rev_pmf;
+  array[estimate_r && seeding_time > 1 ? 1 : 0] real initial_growth; // seed growth rate
+
+  if (num_elements(initial_growth) > 0) {
```
seabbs, line 66: not fitting anything for the growth here seems quite odd and I struggle with the idea it doesn't impact performance. If keeping suggest you chaagne the name as it isn't a prior.

sbfnk: > not fitting anything for the growth here seems quite odd and I struggle with the idea it doesn't impact performance

It's still being fit via the initial number of infections which it depends on linearly - just that the two aren't independently estimated any more which they don't seem to be identifiable as. I'll try it on a few more data sets to check for any issues.

> If keeping suggest you chaagne the name as it isn't a prior.

Will do.

SamuelBrand1: Just sliding in here... I think it makes sense to not treat log-infections before the dynamic modelling and their exp-growth independently (because as you say direct transforms of each other). In the `EpiAware` Renewal modelling component they are also not independent but both conditional on the first sampled `Rt` value. The idea being that this represents the initial exp-growth (via the usual MGF/PGF based transform) and therefore also the initial log-infections.

Are you finding any strong posterior correlation in the first `Rt` value and the prior_infections/initial_infections? Or is that the estimate you use based on the early cases (I think) already covers this?

sbfnk: > Are you finding any strong posterior correlation in the first Rt value and the prior_infections/initial_infections? Or is that the estimate you use based on the early cases (I think) already covers this?

No, they're uncorrelated (both in the versions where they're estimated independently and jointly).

SamuelBrand1: Nice! Thats good to know. That suggests that maybe the R -> r root finder we have could be overkill?

sbfnk: It's more elegant though! What are you doing for the prior on initial number of infections?

SamuelBrand1: Just a single number for log I0, the other values going backwards are constructed by scaling with `exp(rt)` for `t=-1,-2,...` where `r` is determined by the first Rt sample (or in log domain just slope). Its default prior is just a standard normal (in log domain) https://github.com/CDCgov/Rt-without-renewal/blob/104bdd8292f8b017a307913090d8cd9c5e10b4da/EpiAware/src/EpiInfModels/Renewal.jl#L82C1-L107C4

seabbs: The initial infections prior in `EpiAware` is probably not really robust enough for real world use. We don't yet have a good abstraction for applying data at run time hence not worrying that much about it at the moment

NEWS.md
```diff
@@ -19,6 +19,7 @@
 - A bug was fixed where an internal function for applying a default cdf cutoff failed due to a difference a vector length issue. By @jamesmbaazam in #858 and reviewed by @sbfnk.
 - All parameters have been changed to the new parameter interface. By @sbfnk in #871 and reviewed by @seabbs.
 - A bug was fixed where `plot.dist_spec()` wasn't throwing an informative error due to an incomplete check for the max of the specified delay. By @jamesmbaazam in #858 and reviewed by @.
+- Updated the early dynamics calculation to use the full linear model if available. Also changd the prior for initial infections to be approximately Poisson. By @sbfnk in # and reviewed by
```
seabbs, line 22:
```suggestion
- Updated the early dynamics calculation to use the full linear model if available. Also changd the prior for initial infections to be approximately Poisson. By @sbfnk in #903 and reviewed by @seabbs and @SamuelBrand1 
```

seabbs:
> changing the calculation of prior_infections to be based on the intercept of the linear model instead of mean of first week, if available
fitting the initial model for seeding_time if greater than 7 (as that is the period for which exponential growth is assumed in the model)
dividing initial_infections by frac_obs before modelling initial exponential growth
changing the prior for initial_infections to be approximately Poisson

All of this I like a lot and looks correct in the code. 

> I have removed estimation of initial growth and instead made this dependent on initial infections, given the strong correlation in these two parameters seen in the pairs plot above. This seems to yield a substantial speedup whilst retaining the prior exploration of frac_obs as per the linked issue.

> The pairs plot looks very similar to the above (just that this time initial_growth is not estimated independently) making me think that this is preferable.

For all the reasons you say this makes sense and I can't find any real issues in terms of performance in the models I have tested. I also agree that from an identifiability perspective it makes sense this helps stablise things. However, it makes me feel vaguely uncomfortable... :). 

I think I am happy to go ahead and merge. I would much rather move to the EpiAware version ASAP even if it proves to be slower so we can avoid the regression model outside of stan etc which I think is the real source of me being uncomfortable and is just heightened by moving to this new approach.

NEWS.md
```diff
@@ -19,6 +19,7 @@
 - A bug was fixed where an internal function for applying a default cdf cutoff failed due to a difference a vector length issue. By @jamesmbaazam in #858 and reviewed by @sbfnk.
 - All parameters have been changed to the new parameter interface. By @sbfnk in #871 and reviewed by @seabbs.
 - A bug was fixed where `plot.dist_spec()` wasn't throwing an informative error due to an incomplete check for the max of the specified delay. By @jamesmbaazam in #858 and reviewed by @.
+- Updated the early dynamics calculation to use the full linear model if available. Also changd the prior for initial infections to be approximately Poisson. By @sbfnk in #903 and reviewed by @seabbs and @SamuelBrand1
```
seabbs, line 22:
```suggestion
- Updated the early dynamics calculation to use the full linear model if available. Also changed the prior for initial infections to be approximately Poisson and the initial growth rate to the point estimate of the initial growth rate scaled linearly by the estimated initial infections term. By @sbfnk in #903 and reviewed by @seabbs and @SamuelBrand1 
```

seabbs (approved):

PR #905, opened 2024-12-20
skip future check on CRAN

sbfnk:
We're seeing some test failures with the check as CRAN because `{future}` (a suggested package) is missing. Perhaps this was previously pulled in as a lower level dependency and is no more due to a change in one of the dependent packages?

With this PR the check of `setup_future()` is skipped on CRAN - it remains to be seen if any other test fails without `{future}`.

### DIFF
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -16,4 +16,6 @@ if (identical(Sys.getenv("NOT_CRAN"), "true")) {
   }
 }
 
-withr::defer(future::plan("sequential"), teardown_env())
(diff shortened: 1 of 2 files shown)


seabbs (approved):

seabbs:
Hmm used somewhere outside of the tests within the testing framework?

seabbs:
Yes its here: https://github.com/epiforecasts/EpiNow2/blob/92c857a70559687eda87e607452de8c48b15216d/tests/testthat/setup.R#L19

PR #923, opened 2025-01-10
estimate r from initial R

### LINKED ISSUE #920: Use growth approximation to initialise growth rate
Proof of principle for suggestion in https://github.com/epiforecasts/EpiNow2/pull/903#issuecomment-2580960372 blatantly stolen from @SamuelBrand1's code at https://github.com/CDCgov/Rt-without-renewal/blob/main/EpiAware/src/EpiInfModels/utils.jl

``` r
library("rstan")

functions <- "
functions {

  real neg_MGF(real r, vector pmf) {
    int len = num_elements(pmf);
    vector[len] exp_r = exp(-r * linspaced_vector(len, 1, len));
    return(dot_product(pmf, exp_r));
  }

  vector eq(vector r, vector pmf, real R) {
    return([ neg_MGF(r[1], pmf) - 1 / R ]');
  }

  real R_to_r(real R, vector gt_rev_pmf) {
    int gt_len = num_elements(gt_rev_pmf);
    vector[gt_len] gt_pmf = reverse(gt_rev_pmf);
    real mean_gt = dot_product(gt_pmf, linspaced_vector(gt_len, 1, gt_len));
    vector[1] r_approx = [ (R - 1) / (R * mean_gt) ]';
    vector[1] r = solve_newton(eq, r_approx, gt_pmf, R);
  
    return(r[1]);
  }
}
"

model <- stan_model(model_code = functions)
expose_stan_functions(model)

## example
R <- 1.5
gt_rev_pmf <- c(0.2, 0.3, 0.4, 0.1, 0)

r <- R_to_r(R, gt_rev_pmf)
r
#> [1] 0.114143

## should be == R
1 / neg_MGF(r, rev(gt_rev_pmf))
#> [1] 1.5
```

<sup>Created on 2025-01-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

sbfnk:
## Description

This PR closes #920. For now it's using the existing estimate for initial number of infections. As an initial step I'm interested in how doing this affects performance and/or recovery of estimates.


seabbs:
it seems like there are some issues when accumulating to weekly

sbfnk:
This is now ready for initial review but it might require a bit more testing (and doc updates / news item etc) before releasing it in the wild. A few observations so far
- I couldn't get the package to compile when using the algebraic newton solver either locally or on GHA. I could get it to compile with the Powell solver but this led to occasional failures. It all seems to work well (and possibly faster) hand-rolled Newton solver. For now these two methods are both available as stan functions.
- Initialisation of the number of infections remains a critical issue. If doing this independently of estimated growth we get a strong correlation with the initial R as predicted by @SamuelBrand1 - for now what's estimated is roughly the initial number of infections related to the first week of data, which is then scaled based on the growth rate and length of the seeding period; doing it this way reduces the correlation and improves fits, but means we can no longer interpret the corresponding parameter as being the initial number of infections. If going ahead with this we'll have to adjust the simulation functions.
- touchstone is failing now for unknown reasons, which is annoying because it would be good to measure the performance impact of the latest version.

seabbs:
> touchstone is failing now for unknown reasons, which is annoying because it would be good to measure the performance impact of the latest version. 

I see this in epinocast as well. Precisely what is happening is not clear to me.

inst/stan/functions/rt.stan
```diff
@@ -52,16 +52,98 @@ vector update_Rt(int t, real R0, vector noise, array[] int bps,
...
   //breakpoint effects on Rt
   if (bp_n > 0) {
     bp_sd[1] ~ normal(0, 0.1) T[0,];
     bp_effects ~ normal(0, bp_sd[1]);
   }
-  // initial infections
-  initial_infections ~ normal(prior_infections, sqrt(prior_infections));
-  
+  // initial infections scaling (on the log scale)
+  int head_count = num_elements(cases) > 7 ? 7 : num_elements(cases);
```
seabbs, line 64: I would think we would want this step as transformed data?

inst/stan/functions/rt.stan
```diff
@@ -52,16 +52,98 @@ vector update_Rt(int t, real R0, vector noise, array[] int bps,
...
+  real infections_guess = fmax(1, mean(head(cases, head_count)));
+  initial_infections ~ normal(log(infections_guess), 2);
+}
+
+/**
+ * Calculate the negative moment generating function (nMGF) of a probability
```
seabbs, line 70: we should really give sam some attribution in here

seabbs: if he could contribute some maths so support in the vignette that would be cool @samuelbrand1

tests/testthat/_snaps/simulate-infections.md
```diff
@@ -2,67 +2,67 @@
...
-    20: reported_cases 2023-01-06 328.00000
-    21: reported_cases 2023-01-07 364.00000
-    22: reported_cases 2023-01-08 278.00000
-    23: reported_cases 2023-01-09 206.00000
-    24: reported_cases 2023-01-10 169.00000
-    25: reported_cases 2023-01-11 144.00000
-    26: reported_cases 2023-01-12 109.00000
-    27: reported_cases 2023-01-13  80.00000
-    28: reported_cases 2023-01-14  95.00000
+     1:     infections 2023-01-01 100.00000
```
seabbs, line 5: has something changed in the date indexing here?

tests/testthat/_snaps/simulate-infections.md
```diff
@@ -2,67 +2,67 @@
...
-    20: reported_cases 2023-01-06 121.00000
-    21: reported_cases 2023-01-07 215.00000
-    22: reported_cases 2023-01-08  84.00000
-    23: reported_cases 2023-01-09 116.00000
-    24: reported_cases 2023-01-10 132.00000
-    25: reported_cases 2023-01-11  91.00000
-    26: reported_cases 2023-01-12  77.00000
-    27: reported_cases 2023-01-13 120.00000
-    28: reported_cases 2023-01-14  41.00000
+     1:     infections 2023-01-01  59.65590
```
seabbs, line 39: really quite a big change here?

seabbs (commented):
t a high level  I think this looks good and it seems reasonable to rescale initial infections I think. A few comments and questions about output changes. I have just site read so far so need to have a play with the models.

sbfnk:
To be picked up again once proposed model changes in PRs have been finalised.

tests/testthat/_snaps/simulate-infections.md
sbfnk: to be revisited once simulation code has been updated

tests/testthat/_snaps/simulate-infections.md
sbfnk: to be revisited once simulation code has been updated

inst/stan/functions/rt.stan
sbfnk: I've linked out to the original code now - contributions to the methodology docs welcome!

tests/testthat/_snaps/simulate-infections.md
sbfnk: this is changed now but still quite different; it's because of the updated initial growth calculation I think

sbfnk:
> > > touchstone is failing now for unknown reasons, which is annoying because it would be good to measure the performance impact of the latest version.
> > 
> > 
> > I see this in epinowcast as well. Precisely what is happening is not clear to me.
> 
> I now see the same error locally on a linux box. Pretty sure it's unrelated to touchstone or package code but why exactly it happens is a mystery to me, too (as the action goes from success to failure with nearly identical logs, i.e. same package versions, same runner version etc.).

It's this bug in dev stan reported upstream https://github.com/stan-dev/stanc3/issues/1491 - we'll have to either wait for this to be fixed or try to convince touchstone not to update rstan to dev. I've had no success with the latter so it might be easier to just go with the former.

andrjohns:
> > > > touchstone is failing now for unknown reasons, which is annoying because it would be good to measure the performance impact of the latest version.
> > > 
> > > 
> > > I see this in epinowcast as well. Precisely what is happening is not clear to me.
> > 
> > 
> > I now see the same error locally on a linux box. Pretty sure it's unrelated to touchstone or package code but why exactly it happens is a mystery to me, too (as the action goes from success to failure with nearly identical logs, i.e. same package versions, same runner version etc.).
> 
> It's this bug in dev stan reported upstream [stan-dev/stanc3#1491](https://github.com/stan-dev/stanc3/issues/1491) - we'll have to either wait for this to be fixed or try to convince touchstone not to update rstan to dev. I've had no success with the latter so it might be easier to just go with the former.

Apologies for the breakage! The `rstan` develop branch and R-Universe binaries have been updated with the fix

sbfnk:
I think this is good to go - potentially of interest to @SamuelBrand1 @seabbs I had to introduce a truncation step to avoid numeric overflows when the first value of R came very close to zero.

https://github.com/epiforecasts/EpiNow2/blob/dba0c2ef34412f88d5ea527818dbe41bb7e9ec61/inst/stan/functions/rt.stan#L103

seabbs:
> I think this is good to go - potentially of interest to @SamuelBrand1 @seabbs I had to introduce a truncation step to avoid numeric overflows when the first value of R came very close to zero.

Just looking now. Did you compare this approach to padding the initial guess as in theory fmax should be a bit of a no no?

sbfnk:
> > I think this is good to go - potentially of interest to @SamuelBrand1 @seabbs I had to introduce a truncation step to avoid numeric overflows when the first value of R came very close to zero.
> 
> Just looking now. Did you compare this approach to padding the initial guess as in theory fmax should be a bit of a no no?

Padding how? You mean add a small number? I looked at this one but it would be expected to be slower because you now always in the "wrong" place -- what would you expect the issue to be with fmax? It's not applied to anything being sampled.
```r
real r = (R - 1) / (R * mean_gt + 1);
```

sbfnk:
The last two comments tell you the speed difference between the two versions.

I also remember seeing more fitting failures (initial conditions leading to NaN) with that version but could check again. Do you have any other suggestion?

sbfnk:
> The last two comments tell you the speed difference between the two versions.

Overall seems indistinguishable from stochasticity at the used sample size (n = 5)

> I also remember seeing more fitting failures (initial conditions leading to NaN) with that version but could check again. 

Don't see these upon checking again - perhaps got muddled up with other rstan issues. 

I'm just updating the snapshots to see if checks pass then, in which case I'd be happy to go with this version.

seabbs (commented):
>  Do you have any other suggestion?

Aside from the padding you are trying no not really. I think the fmax approach is probably the way to go and think on it later

seabbs (dismissed):
I really like this. I've checked back against the Julia code and agree it aligns. Running locally all seems basically the same as the version on main which is good. 

In terms of the padding vs max I don't feel very strongly and agree there is a bit of a tradeoff

seabbs (approved):

PR #925, opened 2025-01-13
Issue 917: Create a markdown with the results of the synthetic recovery runs

### LINKED ISSUE #197: Potential issue with real time prior for back calculation method
See below example (from examples of `regional_epinow` with `rt = NULL`). Real time Rt appears to be slightly out of line with fully data supported Rt and this is likely an issue with the prior being slightly misaligned. Needs further exploration. As this is currently not a recommended real time measure it should not be in active use anywhere but it needs resolution as this method has many advantages. 

Screenshot 2020-12-03 at 18 30 22

### LINKED ISSUE #917: better expose synthetic recovery
Our synthetic recovery action could be used for checking model correctness. It would be good to make it a bit more visible / user friendly.

kaitejohnson:
## Description

This PR closes #917 (part of it)

It creates a `synthetic_recovery.md` which uses the figures generated in `inst/dev/recover_synthetic/rt.R` and `inst/dev/recover_synthetic/eval_rt.R` to create visualisations of the fits to simulated data under different model specifications. 

This markdown is then rendered and uploaded as an artifact. The last component, posting a comment to the PR with the rendered markdown as an object, I am still stuck on. I could try using touchstone? I am not sure why the live-codes version isn't recognising the id of the artifact. 


sbfnk:
Great idea to use live-codes! I didn't know of its existence.

> I am not sure why the live-codes version isn't recognising the id of the artifact.

I think it might be because it expects the artifact to be uploaded by a different action (for whatever reason) which then triggers the live-codes action - as in the example at https://github.com/live-codes/pr-comment-from-artifact/blob/15cb31a249e2336571f5d8cc244f6132b7d90ead/README.md?plain=1#L29

jamesmbaazam:
Linked issue number should be 917.

kaitejohnson:
> Great idea to use live-codes! I didn't know of its existence.
Clearly I haven't either. I made a separate action since the comment is dependent on the success of the "synthetic validation check" (though they probably can be combined into one, I just wasn't sure). 

I am also testing out an alternative since it seems a bit clunky to have to make another directory `.livecodes` directory and another action just to comment with the artifact!

kaitejohnson:
> you should be able to do this within one file like this https://github.com/epiforecasts/eval-germany-sp-nowcasting/blob/main/.github/workflows/build-and-publish-documentation.yaml I think

My (very limited) understanding is that it requires a separate action and that's why the current one is failing (per @sbfnk's comment) though I could be misunderstanding. 

My attempt at that though, with the separate `comment-markdown.yaml` isn't working either (again, probably just my limited understanding). 

An alternative approach uses a GH action that George wrote for a previous project https://github.com/kaitejohnson/EpiNow2/pull/5#issuecomment-2589387497 which seems to be working, but not sure if we want to use that.

sbfnk:
> My (very limited) understanding is that it requires a separate action and that's why the current one is failing (per @sbfnk's comment) though I could be misunderstanding.

It's how I read the error message but that may have been wrong of course.

> My attempt at that though, with the separate `comment-markdown.yaml` isn't working either (again, probably just my limited understanding).

What went wrong? Looking at the action log for [`885ee49` (#925)](https://github.com/epiforecasts/EpiNow2/pull/925/commits/885ee4963f9daa135bc614fac8556583e0d500db) was the comment action never triggered or is this just not showing because the workflow no longer exists?

kaitejohnson:
> What went wrong? Looking at the action log for [`885ee49` (#925)](https://github.com/epiforecasts/EpiNow2/pull/925/commits/885ee4963f9daa135bc614fac8556583e0d500db) was the comment action never triggered or is this just not showing because the workflow no longer exists?

The action still [exists](https://github.com/kaitejohnson/EpiNow2/blob/197-synth-rec-md/.github/workflows/comment-markdown.yaml) but was never triggered. Do you manually add a new action to the required actions for a pull request? I think this is how previous repos worked for me but only an admin could add them.. not sure if this is the same but that might be why the separate action isn't working.

seabbs:
. Do you manually add a new action to the required actions for a pull request?

No. this suggests there is an issue with the trigger

seabbs:
wait is it just because its not on main? so the trigger isn't quite right

kaitejohnson:
> wait is it just because its not on main? so the trigger isn't quite right

Possibly, I'm not sure. I opened a PR for the alternative method which uses what George had set up (though now appears to be failing) https://github.com/epiforecasts/EpiNow2/pull/928

seabbs:
I don't think we want to depend on CFA internal tooling if we can avoid it .

kaitejohnson:
I had thought the action was just a public action (similar to live codes) but now I am getting an error in the `epiforecasts` org that I didn't get when running from my repo... https://github.com/kaitejohnson/EpiNow2/pull/5

I will try as a temporary test, setting the branches to be this branch to see if I can get the action with live codes to trigger

PR #949, opened 2025-01-31
fix final period estimation when shifting cases

### LINKED ISSUE #948: non-mechanistic model doesn't connect final fit correcly
Looking at #695 I noticed that there's a bit of a discontinuity in the non-mechanistic model (or whatever we may want to call it), see also e.g. the vignette plot (not super clear but visible in the infections plot):

https://epiforecasts.io/EpiNow2/dev/articles/estimate_infections_options-nonparametric-1.png

I think this is because there's an issue with how the data is extrapolated to fill future dates in
https://github.com/epiforecasts/EpiNow2/blob/6c6d90c9527e677c4a12797077b9b36d460cfa6d/R/create.R#L122

sbfnk:
## Description

This PR closes #948 and adds a regression test.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -78,12 +78,13 @@ create_clean_reported_cases <- function(data,
 #' horizon <- 7
 #' smoothing_window <- 14
 #' ## add NAs for horizon
-#' cases <- create_clean_reported_cases(example_confirmed, horizon = horizon)
+#' cases <- create_clean_reported_cases(example_confirmed[1:30])
+#' cases <- add_horizon(cases, horizon)
 #' ## add zeroes initially
 #' cases <- data.table::rbindlist(list(
 #'    data.table::data.table(
 #'      date = seq(
-#'        min(cases$date) - smoothing_window,
+#'        min(cases$date) - 10,
 #'        min(cases$date) - 1,
 #'        by = "days"
 #'      ),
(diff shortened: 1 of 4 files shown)


seabbs:
(no biggy but some more PR info like what is done at a high level would be useful at least for me)

R/create.R
```diff
@@ -109,34 +110,26 @@ create_shifted_cases <- function(data, shift,
   ]
 
   ## Forecast trend on reported cases using the last week of data
-  final_period <- data.table::data.table(
-    confirm =
-      shifted_reported_cases[!is.na(confirm)][
-        max(1, .N - smoothing_window):.N
-      ]$confirm
-  )[,
+  final_period <- shifted_reported_cases[!is.na(confirm)][
```
seabbs, line 113: In the spirit of recent changes there is quite a good arg this should move inside the model

seabbs: as it has always been a big hack

R/create.R
```diff
@@ -109,34 +110,26 @@ create_shifted_cases <- function(data, shift,
...
-  ][
-    ,
-    t := t - (.N - horizon - shift - 6)
-  ][
-    ,
+    date >= min(final_period$date), t := seq_len(.N)
+  ][,
     confirm := data.table::fifelse(
-      t >= 7,
+      !is.na(t) & t >= 0,
```
seabbs, line 124: I think this is the key change. If so makes sense to me.

seabbs (approved):

R/create.R
```diff
@@ -109,34 +110,26 @@ create_shifted_cases <- function(data, shift,
...
   ## Estimate unreported future infections using a log linear model
   shifted_reported_cases <- shifted_reported_cases[
-    ,
-    t := seq_len(.N)
-  ][
-    ,
-    t := t - (.N - horizon - shift - 6)
-  ][
-    ,
+    date >= min(final_period$date), t := seq_len(.N)
```
seabbs, line 121: I think this is the key change and if so all makes sense to me.

R/create.R
sbfnk: yes good point

PR #973, opened 2025-02-04
stan-to-r

sbfnk:
This is a function that I use sometimes for debugging the model by stepping through individual steps. It replicates the `estimate_infections` model (not always up-to-date but updated whenever needed). Putting it here in case others find it useful.


seabbs:
Nice. Could maybe use this in some tests as a motivator for keeping it up to date?

seabbs (approved):

PR #1006, opened 2025-02-27
vignette on forecasting multiple data streams

sbfnk:
I thought it might be nice to have a bit of a demonstration of current functionality to forecast multiple data streams - partly because this is a bit hidden at the moment (even if used e.g. in [Meakin et al., 2022](https://doi.org/10.1186/s12916-022-02271-x)) and partly because it'll be interesting eventually to contrast this with the ideas from https://github.com/epiforecasts/EpiNow2/discussions/371 if we ever go down that route.

It's pretty sparse right now - if anyone would like to add some additional detail/text this would be very welcome in follow up PR(s).


sbfnk:
The eagle-eyed will notice that I snuck in a minor bug fix and a small doc update that I came across while doing this.

seabbs (commented):
This is really nice. Good idea. I like the details drop down for the simulated data a lot. I see a few linting issues some of which are real and some are annoying false positives from the new defaults. It might be nice to reference where people can go to find out what these models are assuming (most importantly the contestant CFR assumption in the secondary model)

seabbs (approved):

PR #1007, opened 2025-02-27
Fix partial matching warning

Bisaloo:
## Description

This PR fixes a partial matching warning

## Initial submission checklist

- [ ] My PR is based on a package issue and I have explicitly linked it.
- [ ] I have tested my changes locally (using `devtools::test()` and `devtools::check()`).
- [ ] I have added or updated unit tests where necessary.
- [ ] I have updated the documentation if required and rebuilt docs if yes (using `devtools::document()`).
- [ ] I have followed the established coding standards (and checked using `lintr::lint_package()`).
- [ ] I have added a news item linked to this PR.

**I could not test the PR locally at this time. I don't expect this small change to cause any issues but I'll keep an eye on the CI checks.**

## After the initial Pull Request 

- [ ] I have reviewed Checks for this PR and addressed any issues as far as I am able.

PR #1010, opened 2025-02-28
Fix partial match of `h` to `horizon`

jamesmbaazam:
## Description

This is not based on an issue but is a follow up to #1007 by replacing partial matches of `$h` with `$horizon`. 

### DIFF
--- a/R/estimate_secondary.R
+++ b/R/estimate_secondary.R
@@ -668,7 +668,7 @@ forecast_secondary <- function(estimate,
   )
   stan_data$n <- nrow(stan_data$primary)
   stan_data$t <- ncol(stan_data$primary)
-  stan_data$h <- nrow(primary[sample == min(sample)])
+  stan_data$horizon <- nrow(primary[sample == min(sample)])
 
   # extract samples for posterior of estimates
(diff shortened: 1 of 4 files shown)


R/estimate_secondary.R
```diff
@@ -668,7 +668,7 @@ forecast_secondary <- function(estimate,
   )
   stan_data$n <- nrow(stan_data$primary)
   stan_data$t <- ncol(stan_data$primary)
-  stan_data$h <- nrow(primary[sample == min(sample)])
+  stan_data$horizon <- nrow(primary[sample == min(sample)])
```
sbfnk, line 671: I think this one is actually called `h` in https://github.com/epiforecasts/EpiNow2/blob/fdeac6a54202eee6219f3917e80f63157bc7721e/inst/stan/simulate_secondary.stan#L14 so would have to change it in the stan model, too

jamesmbaazam: Thanks. Fixed in 59dec8649db78f00ef3bf8b8605f3f2d5a40ff53.

sbfnk:
I think there's another instance in `simulate_secondary()` which is failing now

sbfnk (approved):

PR #1018, opened 2025-03-05
Initial draft of design doc for fitting to multiple data streams

sbfnk:
# Combining multiple observation models

The overall premise is that there is a single latent process (modelled as GP or with the renewal equation) that informs multiple observation processes. Each of these can have:

- a delay or set of delays
- truncation
- scaling
- accumulation
- week(end) effects
- different weights
- overdispersion

The delays need to be identified in case they are common across different observations, e.g.

- incubation period + reporting delay for symptomatic cases
- incubation period + hospital admission + another reporting delay for hospitalisations

We're assuming that delays can be equal but the observations are independent (not e.g. hospitalisation conditional on case report)

This means that the current options specified in `delay_opts()`, `trunc_opts()`, `obs_opts()`, the `accumulation` argument in `forecast_opts()` are all specific to the particular observation type.

### Stan

Observations are converted from an array to a ragged array.
The main required changes here will be lookups so that observations are correctly associated with different observation models.
Some variables e.g. for day of the week effects will have to become 2D arrays.

### R

The observation options mentioned above will have to be combined in a single interface.
The proposed solution is that the current `data`, `delays`, `truncation` and `future_accumulation` become folded into a new `obs_model()` (based on `obs_opts()` and these can be passed as `obs`.

#### Independent observations

```r
estimate_infections(
  obs = c(
    "Confirmed cases" = obs_model(
      data = reported_cases,
      delays = incubation_period + reporting_delay,
      truncation = trunc_opts(Lognormal(1, 2)),
      family = "poisson",
      accumulation = 7
    ),
    "Hospital admissions" = obs_model(
      data = admissions,
      delays = incubation_period + admission_delay,
      truncation = trunc_opts(Gamma(1, 1)),
      family = "negbin",
      weight = 0.5
    )
  ),
  generation_time = generation_time_opts(generation_time)
)
```

#### Dependent observations

This covers the case where only confirmed cases are ever admitted to hospital.

```r
estimate_infections(
  ## "Confirmed cases" is the default name
  obs = obs_model(
      data = reported_cases,
      delays = incubation_period + reporting_delay,
      truncation = trunc_opts(Lognormal(1, 2)),
      family = "poisson",
      accumulation = 7,
      obs = c(
        "Hospital admissions" = obs_model(
          data = admissions,
          delays = admission_delay,
          truncation = trunc_opts(Gamma(1, 1)),
          family = "negbin",
          weight = 0.5
        )
      )
    ),
  ),
  generation_time = generation_time_opts(generation_time)
)
```

### DIFF
--- a/inst/dev/design_obs.md
+++ b/inst/dev/design_obs.md
@@ -37,17 +37,15 @@ The proposed solution is that the current `data`, `delays`, `truncation` and `fu
 ```r
 estimate_infections(
   obs = c(
-    obs_model(
+    "Confirmed cases" = obs_model(
       data = reported_cases,
-      type = "Confirmed cases",
       delays = incubation_period + reporting_delay,
       truncation = trunc_opts(Lognormal(1, 2)),
       family = "poisson",
       accumulation = 7
     ),
-    obs_model(
+    "Hospital admissions" = obs_model(
       data = admissions,
-      type = "Hospital admissions",
       delays = incubation_period + admission_delay,
       truncation = trunc_opts(Gamma(1, 1)),
       family = "negbin",
@@ -64,24 +62,24 @@ This covers the case where only confirmed cases are ever admitted to hospital.
 
 ```r
 estimate_infections(
-  obs = c(
-    obs_model(
+  ## "Confirmed cases" is the default name
+  obs = obs_model(
       data = reported_cases,
-      type = "Confirmed cases",
       delays = incubation_period + reporting_delay,
       truncation = trunc_opts(Lognormal(1, 2)),
       family = "poisson",
       accumulation = 7,
-      obs = obs_model(
-        data = admissions,
-        type = "Hospital admissions",
-        delays = admission_delay,
-        truncation = trunc_opts(Gamma(1, 1)),
-        family = "negbin",
-        weight = 0.5
+      obs = c(
+        "Hospital admissions" = obs_model(
+          data = admissions,
+          delays = admission_delay,
+          truncation = trunc_opts(Gamma(1, 1)),
+          family = "negbin",
+          weight = 0.5
+        )
       )
     ),
   ),
   generation_time = generation_time_opts(generation_time)
 )
-`
+```
--- a/inst/dev/design_obs.md
+++ b/inst/dev/design_obs.md
@@ -1,0 +1,55 @@
+# Combining multiple observation models
+
+The overall premise is that there is a single latent process (modelled as GP or with the renewal equation) that informs multiple observation processes. Each of these can have:
+
+- a delay or set of delays
+- truncation
+- scaling
+- accumulation
+- week(end) effects
+- different weights
+- overdispersion
+
+The delays need to be identified in case they are common across different observations, e.g.
+
+- incubation period + reporting delay for symptomatic cases
+- incubation period + hospital admission + another reporting delay for hospitalisations
+
+We're assuming that delays can be equal but the observations are independent (not e.g. hospitalisation conditional on case report)
+
+This means that the current options specified in `delay_opts()`, `trunc_opts()`, `obs_opts()`, the `accumulation` argument in `forecast_opts()` are all specific to the particular observation type.
+
+## Implementation
+
+### Stan
+
+Observations are converted from an array to a ragged array.
+The main required changes here will be lookups so that observations are correctly associated with different observation models.
+Some variables e.g. for day of the week effects will have to become 2D arrays.
+
+### R
+
+The observation options mentioned above will have to be combined in a single interface.
+The proposed solution is that the current `data`, `delays`, `truncation` and `future_accumulation` become folded into a new `obs_model()` (based on `obs_opts()` and these can be passed as `...`, e.g.
+
+```r
+estimate_infections(
+  cases = obs_model(
+    data = reported_cases,
+    delays = incubation_period + reporting_delay,
+    truncation = trunc_opts(Lognormal(1, 2)),
+    family = "poisson",
+    accumulation = 7
+  ),
+  hospitalisations = obs_model(
+    data = admissions,
+    delays = incubation_period + admission_delay,
+    truncation = trunc_opts(Gamma(1, 1)),
+    family = "negbin",
+    weight = 0.5
+  )
+  generation_time = generation_time_opts(generation_time)
+)
+```
+
+The main work will have to be on the interface, e.g. redefining seeding times., accessing predictions for the different observation types, merging dates, etc.
--- a/inst/dev/design_obs.md
+++ b/inst/dev/design_obs.md
@@ -17,7 +17,7 @@ The delays need to be identified in case they are common across different observ
 
 We're assuming that delays can be equal but the observations are independent (not e.g. hospitalisation conditional on case report)
 
-This means that the current options specified in `delay_opts()`, `trunc_opts()`, `obs_opts()`, the `accumulation` argument in `forecast_opts()` are all specific to the particular observation type.
+This means that the current options specified in `delay_opts()`, `trunc_opts()`, `obs_opts()`, the `accumulation` argument in `forecast_opts()` are all specific to the particular observation type, except delays that can be shared.
 
 ## Implementation
 
--- a/inst/dev/design_obs.md
+++ b/inst/dev/design_obs.md
@@ -30,26 +30,58 @@ Some variables e.g. for day of the week effects will have to become 2D arrays.
 ### R
 
 The observation options mentioned above will have to be combined in a single interface.
-The proposed solution is that the current `data`, `delays`, `truncation` and `future_accumulation` become folded into a new `obs_model()` (based on `obs_opts()` and these can be passed as `...`, e.g.
+The proposed solution is that the current `data`, `delays`, `truncation` and `future_accumulation` become folded into a new `obs_model()` (based on `obs_opts()` and these can be passed as `obs`.
+
+#### Independent observations
 
 ```r
 estimate_infections(
-  cases = obs_model(
-    data = reported_cases,
-    delays = incubation_period + reporting_delay,
-    truncation = trunc_opts(Lognormal(1, 2)),
-    family = "poisson",
-    accumulation = 7
+  obs = c(
+    obs_model(
+      data = reported_cases,
+      type = "Confirmed cases",
+      delays = incubation_period + reporting_delay,
+      truncation = trunc_opts(Lognormal(1, 2)),
+      family = "poisson",
+      accumulation = 7
+    ),
+    obs_model(
+      data = admissions,
+      type = "Hospital admissions",
+      delays = incubation_period + admission_delay,
+      truncation = trunc_opts(Gamma(1, 1)),
+      family = "negbin",
+      weight = 0.5
+    )
   ),
-  hospitalisations = obs_model(
-    data = admissions,
-    delays = incubation_period + admission_delay,
-    truncation = trunc_opts(Gamma(1, 1)),
-    family = "negbin",
-    weight = 0.5
-  )
   generation_time = generation_time_opts(generation_time)
 )
 ```
 
-The main work will have to be on the interface, e.g. redefining seeding times., accessing predictions for the different observation types, merging dates, etc.
+#### Dependent observations
+
+This covers the case where only confirmed cases are ever admitted to hospital.
+
+```r
+estimate_infections(
+  obs = c(
+    obs_model(
+      data = reported_cases,
+      type = "Confirmed cases",
+      delays = incubation_period + reporting_delay,
+      truncation = trunc_opts(Lognormal(1, 2)),
+      family = "poisson",
+      accumulation = 7,
+      obs = obs_model(
+        data = admissions,
+        type = "Hospital admissions",
+        delays = admission_delay,
+        truncation = trunc_opts(Gamma(1, 1)),
+        family = "negbin",
+        weight = 0.5
+      )
+    ),
+  ),
+  generation_time = generation_time_opts(generation_time)
+)
+`


sbfnk:
Building on previous discussion in https://github.com/epiforecasts/EpiNow2/discussions/371

inst/dev/design_obs.md
```diff
@@ -0,0 +1,55 @@
...
+### R
+
+The observation options mentioned above will have to be combined in a single interface.
+The proposed solution is that the current `data`, `delays`, `truncation` and `future_accumulation` become folded into a new `obs_model()` (based on `obs_opts()` and these can be passed as `...`, e.g.
+
+```r
+estimate_infections(
+  cases = obs_model(
+    data = reported_cases,
+    delays = incubation_period + reporting_delay,
```
seabbs, line 39: we spoke in the linked issue about linking delays together across the observations models. I don't think we need this for a first version but within this design could we do so? I think the answer is yes if we make dist_spec have an ID arg and then check that internally? 

If we went that route though what this would preclude is I think your serial case where you have some delay and ascertainment to cases and then some delay and ascertainment to deaths. Another option (and what we went with in Julia) is that you could have a obs_model arg within obs_model that nests one observation model in another (so this would be the serial case). I think I prefer this approach though it requires quite a bit more internal changes to delay handling etc. If that is the design pattern we want to go for I think its all an extension of this interface.

seabbs (dismissed):
I think this looks really good. I have added a note about potential designs for supporting the serial as well as parallel patterns. I thin the judgement call is whether its more work to do this initial feature work and then implement that vs trying to do it all at once

inst/dev/design_obs.md
sbfnk: > If we went that route though what this would preclude is I think your serial case where you have some delay and ascertainment to cases and then some delay and ascertainment to deaths. 

And then have multiple latent processes? I agree that it would be a natural extension but at the same time it would require quite a lot more infrastructure which makes me hesitant to prioritise it. That said we can include it in the design for future work - can you give an example of how this would look like (e.g. in the cases/hospitalisations case above)?

seabbs: No have a single underlying latent process. The use case is the same as the dist id approach but this would be a more explicit way of saying in the UI that one observation is nested within/depends on another. So it would be instead of the dist id approach essentially and give you that + the same thing for shared ascertainment etc.

seabbs: I agree multiple latent processes is a separate and resource intensive problem (that really would only be used in the parallel use case the baseline implementation here would support and not this nested approach).

sbfnk: Oh, so you're thinking that the (deterministic) predicted cases would e.g. be mapped on to hospitalisations? And the syntax would be

```r
estimate_infections(
  cases = obs_model(
    data = reported_cases,
    delays = incubation_period + reporting_delay,
    truncation = trunc_opts(Lognormal(1, 2)),
    family = "poisson",
    accumulation = 7,
    hospitalisations = obs_model(
      data = admissions,
      delays = incubation_period + admission_delay,
      truncation = trunc_opts(Gamma(1, 1)),
      family = "negbin",
      weight = 0.5
    )
  ),
  generation_time = generation_time_opts(generation_time)
)
```

That makes sense to me but personally I don't find this a particularly intuitive syntax. Will try to think of an alternative to suggest (and invite anyone else reading this to put forward ideas).

seabbs: yes but you wouldn't have it be a hospitalisations arg with would look like this. 

```r
estimate_infections(
  obs_model = obs_model(
    data = reported_cases,
    id = "cases",
    delays = incubation_period + reporting_delay,
    truncation = trunc_opts(Lognormal(1, 2)),
    family = "poisson",
    accumulation = 7,
    obs_model = obs_model(
      data = admissions,
      id = "cases"
      delays = incubation_period + admission_delay,
      truncation = trunc_opts(Gamma(1, 1)),
      family = "negbin",
      weight = 0.5
    )
  ),
  generation_time = generation_time_opts(generation_time)
)
```

or you could past named lists as the ID approach. 

Thats interesting. I do because it looks like the model does in reality vs using an ID based approach or similar but I am very open to other suggestions.

sbfnk: Well I was imagining to pass these at `...` args to reduce repeated typing but perhaps that's too confusing.

inst/dev/design_obs.md
```diff
@@ -0,0 +1,55 @@
...
+    delays = incubation_period + admission_delay,
+    truncation = trunc_opts(Gamma(1, 1)),
+    family = "negbin",
+    weight = 0.5
+  )
+  generation_time = generation_time_opts(generation_time)
+)
+```
+
+The main work will have to be on the interface, e.g. redefining seeding times., accessing predictions for the different observation types, merging dates, etc.
```
jamesmbaazam, line 55: I am interested in knowing how we plan to combine the posteriors into a single estimate. Example, given estimate x_{i, t} and y_{j, t} where i and j and data sources, how will we obtain the x_t, the final estimate?

sbfnk: There should be predictions for both i and j - with potentially different missingness patterns.

inst/dev/design_obs.md
sbfnk: Yes exactly.

inst/dev/design_obs.md
```diff
@@ -0,0 +1,55 @@
+# Combining multiple observation models
+
+The overall premise is that there is a single latent process (modelled as GP or with the renewal equation) that informs multiple observation processes. Each of these can have:
```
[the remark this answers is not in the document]
sbfnk: I'd prefer to avoid the repetition.

inst/dev/design_obs.md
```diff
@@ -0,0 +1,55 @@
...
+- overdispersion
+
+The delays need to be identified in case they are common across different observations, e.g.
+
+- incubation period + reporting delay for symptomatic cases
+- incubation period + hospital admission + another reporting delay for hospitalisations
+
+We're assuming that delays can be equal but the observations are independent (not e.g. hospitalisation conditional on case report)
+
+This means that the current options specified in `delay_opts()`, `trunc_opts()`, `obs_opts()`, the `accumulation` argument in `forecast_opts()` are all specific to the particular observation type.
```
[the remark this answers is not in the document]
sbfnk: Sorry, not sure what you mean with this - can you clarify / rephrase?

inst/dev/design_obs.md
sbfnk: I've gone with the suggested approach now (see updated main message) - can always tweak later if anyone has any particularly bright ideas.

inst/dev/design_obs.md
jamesmbaazam: So we will not return a single/combined time series of estimates but separate ones for each data input?

inst/dev/design_obs.md
```diff
@@ -0,0 +1,87 @@
...
+      data = admissions,
+      type = "Hospital admissions",
+      delays = incubation_period + admission_delay,
+      truncation = trunc_opts(Gamma(1, 1)),
+      family = "negbin",
+      weight = 0.5
+    )
+  ),
+  generation_time = generation_time_opts(generation_time)
+)
```
jamesmbaazam, line 58: This looks intuitive.

Side question. Are there packages/interfaces out there that are similar to this?

inst/dev/design_obs.md
```diff
@@ -0,0 +1,87 @@
...
+        type = "Hospital admissions",
+        delays = admission_delay,
+        truncation = trunc_opts(Gamma(1, 1)),
+        family = "negbin",
+        weight = 0.5
+      )
+    ),
+  ),
+  generation_time = generation_time_opts(generation_time)
+)
```
jamesmbaazam, line 86: This is a little hard on the eye and difficult to see.

How about we have the same flat structure and introduce a `obs_nesting` or `obs_dependence` argument that specifies which one depends on the other?

```r
estimate_infections(
  obs = c(
    obs_model(
      data = reported_cases,
      type = "Confirmed cases",
      delays = incubation_period + reporting_delay,
      truncation = trunc_opts(Lognormal(1, 2)),
      family = "poisson",
      accumulation = 7
    ),
    obs_model(
      data = admissions,
      type = "Hospital admissions",
      delays = incubation_period + admission_delay,
      truncation = trunc_opts(Gamma(1, 1)),
      family = "negbin",
      weight = 0.5
    )
  ),
   obs_dependence = c("Confirmed cases", "Hospital Admissions"), # admissions is nested in cases
  generation_time = generation_time_opts(generation_time)
)

File: inst/dev/design_obs.md (hunk 5 above)
Comment (seabbs) [reply inst/dev/design_obs.md:58]: epidemia

File: inst/dev/design_obs.md (hunk 6 above)
Comment (seabbs) [reply inst/dev/design_obs.md:86]: I don't really like this as it two interface layers and I don't think it will be as naturally flexible

File: inst/dev/design_obs.md (hunk 5 above)
Comment (jamesmbaazam) [reply inst/dev/design_obs.md:58]: I didn't mean modelling multiple datasets but the way the models are specified here. `epidemia` uses the formula approach.

File: inst/dev/design_obs.md (hunk 4 above)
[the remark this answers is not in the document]
Comment (sbfnk) [reply inst/dev/design_obs.md:20]:
```suggestion
This means that the current options specified in `delay_opts()`, `trunc_opts()`, `obs_opts()`, the `accumulation` argument in `forecast_opts()` are all specific to the particular observation type, except delays that can be shared.
```

inst/dev/design_obs.md
sbfnk: Given that we specify an ID/type we could think about having a `based_on` (or `reference` or something else) argument which has a default of "Latent infections", e.g.

```{r}
estimate_infections(
  obs = c(
    obs_model(
      data = reported_cases,
      type = "Confirmed cases",
      delays = incubation_period + reporting_delay,
      truncation = trunc_opts(Lognormal(1, 2)),
      family = "poisson",
      accumulation = 7
    ),
    obs_model(
      data = admissions,
      type = "Hospital admissions",
      based_on = "Confirmed cases"
      delays = admission_delay,
      truncation = trunc_opts(Gamma(1, 1)),
      family = "negbin",
      weight = 0.5
    )
  ),
  generation_time = generation_time_opts(generation_time)
)
```

I don't know if this is better but I think @jamesmbaazam's point that mixing observation options with specifying another observation model (which will be based not on the observations but the generated latent counts) could be confusing is valid.

jamesmbaazam: Yes, I like this. It's flat to the eye but flexible in that you can specify dependence. We will need to think about how to handle the "type" specification. I'm not a fan of the free text as it can be difficult to manipulate. We should enforce an acceptable format. I would even suggest just one word. If we are considering using it downstream in plots, etc, then maybe we can have a `label` argument somewhere.

sbfnk: So you think it should be 

```r
estimate_infections(
  obs = c(
    obs_model(
      data = reported_cases,
      type = "cases",
      label = "Confrmed cases"
      delays = incubation_period + reporting_delay,
      truncation = trunc_opts(Lognormal(1, 2)),
      family = "poisson",
      accumulation = 7
    ),
    obs_model(
      data = admissions,
      type = "admissions",
      label = "Hospital admissions",
      based_on = "cases"
      delays = admission_delay,
      truncation = trunc_opts(Gamma(1, 1)),
      family = "negbin",
      weight = 0.5
    )
  ),
  generation_time = generation_time_opts(generation_time)
)
```

Perhaps I'm misunderstanding but I would fear the separation between type and label to be just be a source of confusion.

jamesmbaazam: Yes, this is what I was imagining. Will the confusion come from the naming? Would properly documenting it not fix that? 

`type`: Unique id for identifying and tracking the dataset internally.
`label`: Label to be used in plots and summaries.

seabbs: > Perhaps I'm misunderstanding but I would fear the separation between type and label to be just be a source of confusion.

I agree.

I find this very hard to parse and the major downside is you can specify low level modules and look at them and slowly build them up over time (as the building up is really a concatenation followed by a second layer of structural declaration). 

In my mind this would be like saying a nested tibble should be defined as a flat list with a structure argument or as in the way matrices are typically are defined which I think can have a lot of painful edge cases

jamesmbaazam: I don't feel strongly about my suggestion, really. It's just another way of looking at it. Two options, we choose whichever is not too disruptive and potentially more familiar to the users. So, maybe, some power user feedback here would be helpful.

sbfnk: I would tend to agree that the nested design is clearer overall - perhaps there are still improvements we could make to it.

seabbs: I think my ideal would be this: 


```r
estimate_infections(
  obs = list(
    "Confirmed cases" = obs_model(
      data = reported_cases,
      delays = incubation_period + reporting_delay,
      truncation = trunc_opts(Lognormal(1, 2)),
      family = "poisson",
      accumulation = 7,
      obs = list(
		"Hospital admissions" = obs_model(
        	data = admissions,
        	delays = admission_delay,
       		 truncation = trunc_opts(Gamma(1, 1)),
        	family = "negbin",
        	weight = 0.5
      	))
    ),
  ),
  generation_time = generation_time_opts(generation_time)
)

```

with the option of not specifying the names vs using a type arg

sbfnk: I like the named arguments, maybe with a custom `c.obs_model()` instead of a generic list? Both would be fairly easy to implement, not sure which one is easier to use.

I'm not a huge fan of supporting both named arguments and a `type` argument and feel we should perhaps be more prescriptive so as to keep the interface as unambiguous as possible.

seabbs: Sorry, I meant do away with type argument but not enforce have a name in the custom c method i.e it just defaults to something if you don't provide one.

sbfnk: OK, let's go with this for now and then we can always tweak later when pain points emerge.

PR #1032, opened 2025-03-12
only use shifted cases if needed

### LINKED ISSUE #1030: `estimate_infections()` errors when there are too many consecutive NAs
Summary:
`estimate_infections()` errors when there are too many consecutive NAs

Description:
This is because the moving average in `create_shifted_cases()` will leave NAs in `shifted_cases`. None of this is needed in the default model so we should only use this if using the deconvolution model and in that case leave an informative error message.

Reproducible Steps:
``` r
library(EpiNow2)
#> 
#> 
#>     Gamma

options(mc.cores = 4)

# get example case counts
reported_cases <- example_confirmed[1:60]
# remove some
reported_cases[30:50, confirm := NA]

# for more examples, see the "estimate_infections examples" vignette
def <- estimate_infections(reported_cases)
#> Warning: ! No generation time distribution given. Using a fixed generation time of 1
#>   day, i.e. the reproduction number is the same as the daily growth rate.
#> ℹ If this was intended then this warning can be silenced by setting `dist =
#>   Fixed(1)`'.
#> Error in FUN(X[[i]], ...) : 
#>   Stan does not support NA (in shifted_cases) in data
#> failed to preprocess the data; sampling not done
#> Error in `fit_model_with_nuts()`:
#> ! model fitting was timed out or failed
```

<sup>Created on 2025-03-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>
EpiNow2 Version:
`main`

sbfnk:
## Description

This PR closes #1030, also removing a bit of code overhead along the way (including returning of shifted cases). Locally I get an update of the simulation snapshots which I don't quite understand so monitoring here to see if it's just an `rstan` version mismatch.

I have not added a regression test as it seems a bit of an edge case but could do so if people think it would be useful.


R/create.R
```diff
@@ -409,6 +424,18 @@ create_stan_data <- function(data, seeding_time, rt, gp, obs, backcalc,
   imputed_times <- cases[!(accumulate), lookup]
   accumulate <- cases$accumulate
   confirmed_cases <- cases[1:(.N - forecast$horizon)]$confirm
+  if (is.null(rt)) {
```
seabbs, line 427: This should already be tested if we are checking both the Rt and deconvolution methods?

sbfnk: Yes just not explicitly that it works with lots of NAs (i.e. the current failure state).

seabbs (dismissed):
This looks good to me. I think we do want a all NA test as this is functionality people might want (it will look like sampling from the prior). My assumption though is that there will be other parts of the code where all NA falls over? If this is the case that seems like a new feature issue to enable this functionality.

If it does break for all NA we could and maybe should add a test here for as much missingness as it does support and then use that to build on for this hypothetical new feature.

sbfnk:
That is a really nice idea.

sbfnk:
>  assumption though is that there will be other parts of the code where all NA falls over?

Turns out this worked pretty much out of the box (with one minimal change).

seabbs (approved):
Nice. LGTM. Do we want a new issue to surface the idea of being able to do prior predictives now?

sbfnk:
Maybe. They look a bit mad, perhaps predictably so

image

seabbs:
> They look a bit mad, perhaps predictably so

Yes I think this is a good insight though that says something about the default parameterization we use.

PR #1076, opened 2025-04-27
Growth estimators

### LINKED ISSUE #768: Growth rate estimators
The growth rate is currently estimated via the log derivative of infections, see https://github.com/epiforecasts/EpiNow2/blob/35483d03327cdcce4be0a3ecf5aa7c9466aa42be/inst/stan/functions/generated_quantities.stan#L33-L39

I know that this is the classical interpretation of the growth rate, but wanted to point out the alternative version suggested in [Parag et al.](https://doi.org/10.1111/rssa.12867) which instead computes the log derivative on the _infectiousness_ (and shifts it by the mean generation time). When testing this in [EpiSewer](https://github.com/adrian-lison/EpiSewer/blob/1eecafeeafa9570ea2dcc0b761296948bd1927dc/R/sewer_summary.R#L80-L91), it seemed to be slightly more stable and `r=0` had a better temporal matching with `Rt=1`.

adrian-lison:
## Description

This PR closes #768.

The PR adds a new option to calculate the growth rate in `estimate_infections()` and `simulate_infections()` that is based on an estimator by Parag et al. 2022.

## Summary by CodeRabbit

- New Features
  - Added a selectable growth rate method (“infections” or “infectiousness”) based on Parag et al., available in options and simulations.
  - Introduced a numerically stable convolution utility (stable_convolve).
- Documentation
  - Expanded guidance for growth_method, updated defaults, and added references; clarified argument defaults elsewhere.
- Tests
  - Added comprehensive tests covering the new growth methods.
- Chores
  - Updated package authorship metadata.
  - Minor cleanup to reduce unnecessary dependency imports.

### DIFF
--- a/R/extract.R
+++ b/R/extract.R
@@ -221,7 +221,7 @@ extract_parameter_samples <- function(stan_fit, data, reported_dates,
     samples,
     reported_dates[-1]
   )
-  incomplete_dates <- unique(out$growth_rate[is.na(value),][["date"]])
+  incomplete_dates <- unique(out$growth_rate[is.na(value), ][["date"]])
   out$growth_rate[date %in% incomplete_dates, value := NA]
   if (data$week_effect > 1) {
     out$day_of_week <- extract_parameter(


adrian-lison:
@seabbs @sbfnk 

Just opened a draft PR while working on this. I have drafted the basic stan code, but couldn't test it locally yet because of issues with my rstan installation ^^

I have for now added an argument `growth_method` to the rt_opts(), because I didn't find it justified to create a completely new opts type. There are a couple of open questions to discuss, in particular how much of a problem it is that with the alternaitve approach the most recent growth rates are undefined due to the shift applied.

Will get back with more details soonish.

R/simulate_infections.R
```diff
@@ -104,7 +107,10 @@ simulate_infections <- function(R, initial_infections,
     initial_infections = array(log(initial_infections), dim = c(1, 1)),
     initial_as_scale = 0,
     R = array(R$R, dim = c(1, nrow(R))),
-    pop = pop
+    pop = pop,
+    growth_method = list(
+      "infections" = 0, "infectiousness" = 1
+    )[[rt$growth_method]]
```
sbfnk:
```suggestion
    )[[growth_method]]
```

This is causing the check failures I think

inst/stan/functions/generated_quantities.stan
```diff
@@ -60,11 +60,50 @@ vector calculate_Rt(vector infections, int seeding_time,
...
+  vector[ot] growth = rep_vector(1e-5, ot);
+  int gt_length = num_elements(gt_rev_pmf);
+  // mean generation time, will always be >= 1
+  int mean_gen = to_next_int_index( // round weighted mean to next int
+    dot_product(reverse(linspaced_vector(gt_length, 1, gt_length)), gt_rev_pmf)
+    );
+  growth[1:(ot-mean_gen)] = (
+    infness_log[(1+mean_gen):ot] - infness_log[mean_gen:(ot - 1)]
+    );
+  //growth[(ot-mean_gen+1):ot] = rep_vector(-negative_infinity(), mean_gen); // remains undefined due to shift
```
sbfnk: Should we set these to [`not_a_number()`](https://mc-stan.org/docs/stan-users-guide/floating-point.html#not-a-number-values)? For plotting / data extraction I'm not sure we'll want these to be set to zero.

sbfnk:
Also please add a news item and add yourself as author to the `DESCRIPTION`.

R/opts.R
```diff
@@ -290,14 +290,16 @@ rt_opts <- function(prior = LogNormal(mean = 1, sd = 1),
                     use_breakpoints = TRUE,
                     future = "latest",
                     gp_on = c("R_t-1", "R0"),
-                    pop = 0) {
+                    pop = 0,
+                    growth_method = c("infections", "infectiousness")) {
```
seabbs, line 310: Missing docs

inst/stan/estimate_infections.stan
```diff
@@ -258,26 +259,39 @@ generated quantities {
...
+        gt_rev_pmf_for_growth = sampled_gt_rev_pmf;
+        // calculate Rt using infections and generation time
+        gen_R = calculate_Rt(
+          infections, seeding_time, sampled_gt_rev_pmf, rt_half_window
+        );
+      } else {
+        gt_rev_pmf_for_growth = gt_rev_pmf;
+      }
+  
+      // estimate growth from infections
```
seabbs, line 285: to reduce noise here do you think we should wrap it in a calculate_growth wrapper?

seabbs (commented):
this is really nice sorry it got no response for ages. Some minor cleaning up suggestions. 

On an related but not for now point I wonder if we should be doing the prostprocessing in stan or R for these measures? I see upsides and downsides to both options.

R/opts.R
```diff
@@ -290,14 +304,16 @@ rt_opts <- function(prior = LogNormal(mean = 1, sd = 1),
...
   opts <- list(
     use_rt = use_rt,
     rw = rw,
     use_breakpoints = use_breakpoints,
     future = future,
     pop = pop,
-    gp_on = arg_match(gp_on)
+    gp_on = arg_match(gp_on),
+    growth_method = arg_match(growth_method)
```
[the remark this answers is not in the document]
adrian-lison: The documentation says: 'Options are "infections" (default) and "infectiousness".'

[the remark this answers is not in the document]
adrian-lison: I don't see a problem

adrian-lison:
@seabbs @sbfnk Thanks for your feedback and very sorry that this fell through the cracks! I have incorporated the suggested changes and made a couple of further edits:

- the shifting of the growth rate time series that results from the differencing is now internally handled by the `calculate_growth_` methods. So far, this was handled more ad-hoc by passing `seeding_time + 1` to `seeding_time`, which has potential for confusion.
- Since the `infectiousness` growth rate method has an additional shift by the mean generation time, causing the most recent growth rates to be `NaN`, there was another issue: if the generation time is sampled in the model, its mean may of course vary by more than 1 day, and therefore the number of undefined dates can differ across draws. This led to a bias towards shorter generation times in the growth rate summary, which only averaged over the draws with a defined rate for a given date. I addressed this for now by setting the growth rate to `NaN` consistently over all draws for these dates post-hoc (on the extracted samples).

Below is a comparison of the growth rate estimates from both methods on the `example_confirmed` cases. They are really similar except for the start of the time series (but this seems to match with the simulation from Parag et al. 2022).

growth_rate_methods

sbfnk (approved):
Looks great, thank you very much!

PR #1140, opened 2025-10-29
Improve discretise with primarycensored

### LINKED ISSUE #776: Improve discretise
`primarycensoreddist` now supports numerically solved double censored pmf  calculation (https://primarycensoreddist.epinowcast.org/dev/articles/primarycensoreddist.html#compute-the-primary-event-censored-probability-mass-function-pmf-with-dprimarycensoreddist) which we should replace our current approximate `dicretise` approach with either wrapped or directly.

seabbs:
**This is entirely from an agent so do not review until I have pinged for review as I will do a first pass**

## Summary

Replaces the approximate CDF difference discretisation approach with double censored PMF calculations from the `primarycensored` package.

## Changes

- Uses `primarycensored::dprimarycensored()` for more accurate PMF computation
- Updates all test expected values to reflect the more accurate discretisation
- Adds `primarycensored` to package dependencies
- Updates documentation with new methodology and references
- Adds breaking change notice to NEWS.md

## Breaking Change

⚠️ This changes the exact numerical values returned by `discretise()`. The new approach is more numerically accurate and correctly implements double censoring, but code depending on specific numerical outputs will produce different results. The differences are small and represent improvements in accuracy. The function interface remains unchanged.

## Testing

- All existing tests updated with new correct values
- Tests verify PMF properties (sums to 1, non-negative)
- Verified with LogNormal, Gamma, Normal, and Fixed distributions

Fixes #776

## Release Notes

* **Breaking Changes**
  * discretise() now uses an updated methodology for probability mass function calculations; function interface remains unchanged but numerical outputs may vary slightly.

* **Documentation**
  * Updated methodology descriptions for probability mass function computations in related documentation.

* **Chores**
  * Reorganised test suite into core and integration tests, improving local development performance by 77%.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.github/workflows/document.yaml
+++ b/.github/workflows/document.yaml
@@ -39,7 +39,7 @@ jobs:
           use-public-rspm: true
 
       - name: Cache R package
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: ${{ env.R_LIBS_USER }}
           key: ${{ runner.os }}-r-${{ hashFiles('DESCRIPTION') }}
--- a/.github/workflows/lint-package.yaml
+++ b/.github/workflows/lint-package.yaml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Create or update issue
         if: steps.lintr.outputs.has_lints == 'true'
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             const fs = require('fs');
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Deploy to GitHub pages 🚀
         if: github.event_name != 'pull_request'
-        uses: JamesIves/github-pages-deploy-action@v4.7.4
+        uses: JamesIves/github-pages-deploy-action@v4.7.6
         with:
           clean: false
           branch: gh-pages
--- a/.github/workflows/pr-commands.yaml
+++ b/.github/workflows/pr-commands.yaml
@@ -80,12 +80,12 @@ jobs:
       NOT_CRAN: true
       EPINOW2_SKIP_INTEGRATION: false
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: r-lib/actions/pr-fetch@v2
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Post start comment
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             github.rest.issues.createComment({
@@ -114,7 +114,7 @@ jobs:
         shell: Rscript {0}
       - name: Post success comment
         if: success()
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             github.rest.issues.createComment({
@@ -125,7 +125,7 @@ jobs:
             })
       - name: Post failure comment
         if: failure()
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             github.rest.issues.createComment({
--- a/.github/workflows/render-EpiNow2.yaml
+++ b/.github/workflows/render-EpiNow2.yaml
@@ -44,14 +44,14 @@ jobs:
         shell: Rscript {0}
 
       - name: Upload vignette as an artifact
-        uses: actions/upload-artifact@v5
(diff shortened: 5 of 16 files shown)


sbfnk:
Related: https://github.com/epiforecasts/EpiNow2/issues/410

I'm pretty much on board with what you suggest as at the moment we're essentially changing the upper bound when discretising. That said I think it might be worth thinking about this a bit more because:
1. it's a fairly breaking change as it'll change pretty much all the results.
1. It's strictly speaking the supremum, not the maximum?
1. It might still be confusing that the maximum discretised value is max-1

They're all ultimately acceptable issues but I wonder if there's perhaps an argument name change that would make things clearer?

sbfnk:
# Claude says:

I think the core issue is that there are fundamentally **two different maximums** at play:

1. **Truncation point for the continuous distribution** (D = 5 in `primarycensored`)
2. **Maximum discrete delay in the output** (4, when width = 1)

Any name like `max_delay` is ambiguous about *which* maximum you mean, which is where the confusion comes from.

## Options to clarify:

### 1. `max_discrete_value` or `max_discrete_delay`
Make the argument represent what users actually **see** in the output:
```r
discretise(..., max_discrete_value = 4)  # Output: delays 0, 1, 2, 3, 4
# Internally calculates D = max_discrete_value + width = 5
```

**Pro:** No confusion - you specify exactly what maximum you want in your output  
**Con:** Requires calculating D internally

### 2. `upper_bound` or `truncate_at`
Make it very clear this is a **boundary**, not a value:
```r
discretise(..., truncate_at = 5)  # Clear: truncated at 5
# Documented: maximum discrete delay will be truncate_at - width = 4
```

**Pro:** Statistically precise  
**Con:** Still requires users to do mental math to know their max discrete value

### 3. `n_delays`
Specify the number of discrete delays:
```r
discretise(..., n_delays = 5)  # Number of discrete delays (0, 1, 2, 3, 4)
```

**Pro:** Completely sidesteps the "which maximum?" question  
**Con:** Changes the semantics more substantially

## My recommendation: `n_delays`

This might be the cleanest because it completely avoids the ambiguity. Users think: "I want to model 5 discrete delays" rather than "I want to truncate at 5" or "I want max delay of 4". The implementation detail of what D needs to be becomes internal.

Alternatively, if keeping the current semantics (specifying truncation point), use `truncate_at` with **very** clear documentation that max discrete delay = `truncate_at - width`.

tests/testthat/test-seeding-time.R
```diff
@@ -4,7 +4,7 @@ test_that("Seeding times are correctly calculated", {
   delay1 <- LogNormal(mean = 5, sd = 1, max = 9)
   delay2 <- LogNormal(mean = 7, sd = 3, max = 14)
   expect_equal(
-    EpiNow2:::get_seeding_time(delay1, gt1 + gt2), 23L ## 9 + 14
+    EpiNow2:::get_seeding_time(delay1, gt1 + gt2), 21L ## 8 + 13 (max reduced by 1 each due to discretisation)
```
seabbs:
```suggestion
    EpiNow2:::get_seeding_time(delay1, gt1 + gt2), 21L ## 8 + 13
```

seabbs:
I'm not seeing these snapshot issues locally but I am seeing a test fail for `estimate_secondary` when it tries to fit the model. Very unclear to me what is the snapshot issue so a bit blocked here.

sbfnk:
The snaps are still distinct from main -- recommend a `git checkout main tests/testthat/snaps` and then commit/push.

sbfnk:
> Yes I agree with `truncate_at` or `truncate_beyond` as the way forward with no strong preference between those.

Let's go with `truncate_at` - perhaps put in a separate Issue as if doing this it'll be a fairly substantial change requiring lots of changes in the examples etc.

R/dist_spec.R
```diff
@@ -7,23 +7,30 @@
 #'
 #' # Methodological details
 #'
+#' The probability mass function is computed using the primarycensored package,
+#' which provides double censored PMF calculations. This correctly represents
+#' the probability mass function of a double censored distribution arising from
+#' the difference of two censored events.
```
sbfnk:
```suggestion
#' The probability mass function is computed using the `{primarycensored}`
#' package, which provides double censored PMF calculations. This correctly
#' represents the probability mass function of a double censored distribution
#' arising from the difference of two censored events.
```

jamesmbaazam:
Just to flag that #1178 aspirationally mentions that `estimate_delay()` is a wrapper around `{primarycensored}`.

PR #1144, opened 2025-10-31
Standardise S3 interface for estimate_infections and add get_samples()

### LINKED ISSUE #1141: S3 class for estimate_infections
from the design doc:

- `samples`: `get_samples()` (or `extract_samples()`) built on the existing `extract_parameter_samples()`; this could also take a parameter name and look up the ID, or take a delay distribution name and look up the ID; 
- `summarised`: `summary()`
- `fit`: `get_stan_fit()`
- `args`: not needed (only internally to identify parameter ids)?
- `observations`: not needed (only internally by `summary()` and `get_samples()`

sbfnk:
## Summary

This PR starts the S3 refactoring work by:

1. **Standardised `estimate_infections()` return structure** - now consistently returns `list(fit, args, observations)` with class `"estimate_infections"`
2. **Added `get_samples()` S3 generic** - provides a clean interface to extract posterior samples from fitted models
3. **Fixed `forecast_infections()` compatibility** - handles objects with pre-formatted `$samples` instead of `$fit`
4. **Deprecated `summary(type = "samples")`** - in favor of the new `get_samples()` function
5. **Added comprehensive tests** - for new functionality and deprecation warnings
6. **Updated documentation and vignettes** - to use the new `get_samples()` function

Closes #1141

### Standardised `estimate_infections()` S3 interface
- Returns consistent structure: `list(fit, args, observations)`
- Assigned `"estimate_infections"` class
- Enables proper S3 method dispatch

### New `get_samples()` S3 generic
- Provides consistent interface for extracting posterior samples
- Works with both `estimate_infections()` and `forecast_infections()` results
- Auto-detects and extracts all parameters from Stan fits
- Returns samples in standardized long-format data.table

### Issue Found: `forecast_infections()` Object Structure
During this work, we discovered that `forecast_infections()` returns objects with class `"estimate_infections"` but a completely different structure:
- `estimate_infections()`: `list(fit, args, observations)` with Stan fit
- `forecast_infections()`: `list(samples, summarised, observations)` with pre-formatted results

This is problematic because `forecast_infections()` re-simulates all MCMC samples with a different R trajectory (so it has samples), while `estimate_infections()` stores the Stan fit. They represent fundamentally different computations but share the same class.

### Workaround
For now, `get_samples.estimate_infections()` checks if `$samples` already exists before extracting from `$fit`. This works but is a temporary solution.

### Future Work
In a follow-up PR, we should:
1. Create a separate `"forecast_infections"` S3 class for `forecast_infections()` results
2. Refactor `plot.estimate_infections()` to be a thin wrapper that extracts samples and calls a shared low-level plotting function
3. Create `plot.forecast_infections()` that uses the same low-level plotting function

This would make the class hierarchy clearer and avoid special cases in methods. Note that `simulate_infections()` is different again - it creates a single trajectory from given parameters (no samples), so doesn't need this treatment.

## Test Results

- ✅ Manual test of `forecast_infections()` plot passes
- ✅ All existing tests pass
- ✅ New deprecation tests added
- ✅ Comprehensive test coverage for `get_samples()`

## Breaking Changes

None - all changes are backward compatible with deprecation warnings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  - Added `get_samples()` function to access formatted posterior samples.
  - Added print method for epinowfit objects.
  - Direct object property access: `object$fit`, `object$args`, and `object$observations`.

* **Deprecated**
  - `summary(type = "samples")` is now deprecated; use `get_samples()` instead.
  - Removed `CrIs` parameter from `estimate_infections()`, `estimates_by_report_date()`, `simulate_infections()`, and related functions.

### DIFF
--- a/R/epinow.R
+++ b/R/epinow.R
@@ -97,13 +97,6 @@ epinow <- function(data,
                    logs = tempdir(), id = "epinow", verbose = interactive(),
                    filter_leading_zeros = TRUE, zero_threshold = Inf, horizon
                    ) {
-  if (!missing(CrIs)) {
-    lifecycle::deprecate_stop(
-      "1.8.0",
-      "estimate_infections(CrIs)",
-      detail = "Specify credible intervals when using `summary()` or `plot()`."
-    )
-  }
   if (!missing(filter_leading_zeros)) {
     lifecycle::deprecate_stop(
       "1.7.0",


sbfnk:
This is ready for review now. Some minor changes:
- moved `format_fit()` to the new `format.R` - it's unchanged
- some issue with memory use blowing up in GHA on Ubuntu necessitated specifying the globals that are passed to workers in `forecast_infections()`. Probably a good idea anyway.

R/estimate_infections.R
```diff
@@ -143,6 +145,13 @@ estimate_infections <- function(data,
                                 filter_leading_zeros = TRUE,
                                 zero_threshold = Inf,
                                 horizon) {
+  if (!missing(CrIs)) {
+    lifecycle::deprecate_warn(
```
seabbs: Should this error this doesn't do anything now does it?

R/get.R
```diff
@@ -217,3 +217,53 @@ get_seeding_time <- function(delays, generation_time, rt = rt_opts()) {
...
+#' # Filter to specific parameters
+#' R_samples <- samples[variable == "R"]
+#' }
+get_samples <- function(object, ...) {
+  UseMethod("get_samples")
+}
+
+#' @rdname get_samples
+#' @export
+get_samples.estimate_infections <- function(object, ...) {
```
seabbs, line 247: I assume we want a new issue to make a method for this for epinow and for other functions where it makes sense or we don't need it to be s3.

R/plot.R
```diff
@@ -418,5 +399,39 @@ plot.estimate_infections <- function(x,
 #' @return List of plots as produced by [report_plots()]
 #' @export
 plot.epinow <- function(x, type = "summary", ...) {
-  plot.estimate_infections(x$estimates, type = type, ...)
+  out <- report_plots(
```
seabbs, line 402: What is the benefit of getting rid of this being an alias for plot.estimate_infections?

seabbs: Also do we have an issue for nuking report_plots from space? I think the postprocessing code is some of the most offensive work I have ever done.

R/get.R
sbfnk: yes I was thinking that `estimate_secondary()` and `estimate_truncation()` might need their own methods but this remains to be seen.

sbfnk: https://github.com/epiforecasts/EpiNow2/issues/1151

R/plot.R
sbfnk: The problem is that `<epinow>` objects have quite a different structure to `<estimate_infections>` ones - in particular they don't store samples/fits by default. I thought it made sense to preserve this (otherwise what's the point of `epinow` but it does lead to some clunky multiplication of code.

I think we might want a discussion on whether we want to get of the `epinow()` function/class altogether.

sbfnk: https://github.com/epiforecasts/EpiNow2/issues/1152

R/estimate_infections.R
sbfnk: Debatable. What's the greater risk:
1. people's workflows breaking because they specify this argument or
1. people's workflows breaking because they get the wrong CrIs when doing summary/plot

Probably (2) as you suggest so I'll make this change.

PR #1156, opened 2025-11-06
Extend S3 class structure to forecast_infections()

### LINKED ISSUE #1153: Extend S3 class structure to `forecast_infections()`
From https://github.com/epiforecasts/EpiNow2/pull/1144#issue-3573851933

We should:
1. Create a separate `"forecast_infections"` S3 class for `forecast_infections()` results
2. ~Refactor `plot.estimate_infections()` to be a thin wrapper that extracts samples and calls a shared low-level plotting function~ (already done)
3. Create `plot.forecast_infections()` that uses the same low-level plotting function

This would make the class hierarchy clearer and avoid special cases in methods. Note that `simulate_infections()` is different again - it creates a single trajectory from given parameters (no samples), so doesn't need this treatment.

sbfnk:
## Description

This PR closes #1153.

This PR extends the S3 class structure to `forecast_infections()` by creating an independent `"forecast_infections"` class instead of having it inherit from `"estimate_infections"`. This clarifies the architectural distinction between:
- **`estimate_infections`** objects: contain a Stan `fit` for diagnostics (e.g., `loo()`)
- **`forecast_infections`** objects: contain pre-computed `samples` without the underlying fit

### Key changes:

1. **Independent S3 class**: `forecast_infections()` now returns `c("forecast_infections", "list")` instead of `c("forecast_infections", "estimate_infections", "list")`

2. **Dedicated S3 methods** for `forecast_infections`:
   - `get_samples.forecast_infections()`: Returns pre-formatted samples
   - `summary.forecast_infections()`: Provides summary statistics
   - `plot.forecast_infections()`: Creates plots with shared infrastructure

3. **Reduced code duplication**: Extracted common summary logic into `create_infection_summary()` helper function in `create.R`

4. **Enhanced user control**: Both `plot.estimate_infections()` and `plot.forecast_infections()` now accept a `CrIs` argument to customize displayed credible intervals

5. **Comprehensive tests**: Added tests to verify class structure and CrIs functionality

## Release Notes

* **New Features**
  * Forecast outputs now have dedicated summary(), plot(), and get_samples() methods for improved handling of simulated data.
  * plot() functions now accept a CrIs parameter to control which credible intervals are displayed.
  * New infection summary function supporting snapshot and parameters modes for flexible output summarisation.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -409,11 +409,7 @@ plot.forecast_infections <- function(x,
                                      type = "summary",
                                      CrIs = c(0.2, 0.5, 0.9),
                                      ...) {
-  out <- report_plots(
-    summarised_estimates = summary(x, type = "parameters", CrIs = CrIs),
-    reported = x$observations, ...
-  )
-  select_plots(out, type)
+  plot.estimate_infections(x, type = type, CrIs = CrIs, ...)
 }
 
 #' Plot method for epinow


man/create_infection_summary.Rd
```diff
@@ -0,0 +1,53 @@
...
+
+\item{date}{A date in the form "yyyy-mm-dd" to inspect estimates for.
+Defaults to the latest date with data.}
+
+\item{params}{A character vector of parameters to filter for.}
+
+\item{CrIs}{Numeric vector of credible intervals to calculate. Defaults
+to c(0.2, 0.5, 0.9).}
+
+\item{...}{Additional arguments passed to \code{\link[=report_summary]{report_summary()}}.}
```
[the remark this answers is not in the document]
sbfnk: There's no ... argument in calc_summary_measures

R/create.R
```diff
@@ -843,3 +843,82 @@ create_stan_params <- function(params) {
   }
   c(ret, as.list(ids), as.list(null_ids))
 }
+
+#' Create Summary Output from Infection Estimation Objects
```
seabbs: Are we consistent not using caps outside the first word throughout the package? It looks like we are in the functions around this one.

R/create.R
```diff
@@ -843,3 +843,82 @@ create_stan_params <- function(params) {
...
+#' @param object An infection estimation object (either from
+#'   [estimate_infections()] or [forecast_infections()]).
+#'
+#' @param type A character vector of data types to return. Defaults to
+#'   "snapshot" but also supports "parameters". "snapshot" returns
+#'   a summary at a given date (by default the latest date informed by data).
+#'   "parameters" returns summarised parameter estimates that can be further
+#'   filtered using `params` to show just the parameters of interest and date.
+#'
+#' @param date A date in the form "yyyy-mm-dd" to inspect estimates for.
```
seabbs: If it actually is a date object does the format matter? or is it a character string to be coerced?

R/get.R
```diff
@@ -267,3 +262,10 @@ get_samples.estimate_infections <- function(object, ...) {
     observations = object$observations
   )
 }
+
+#' @rdname get_samples
+#' @export
+get_samples.forecast_infections <- function(object, ...) {
+  # forecast_infections objects already have pre-formatted samples
```
seabbs: dont' need this comment or the one about extracting the fit

R/plot.R
```diff
@@ -372,17 +372,45 @@ plot_summary <- function(summary_results,
...
+#'
+#' @seealso [report_plots()] [forecast_infections()]
+#' @method plot forecast_infections
+#' @return List of plots as produced by [report_plots()]
+#' @export
+plot.forecast_infections <- function(x,
+                                     type = "summary",
+                                     CrIs = c(0.2, 0.5, 0.9),
+                                     ...) {
+  out <- report_plots(
```
seabbs: As internally this is the same as the function above (I think) do we want to call it in here vs duplicating code

R/summarise.R
```diff
@@ -843,43 +844,40 @@ summary.estimate_infections <- function(object,
...
-  samples <- get_samples(object)
-
-  # Calculate summary measures
-  summarised <- calc_summary_measures(
-    samples,
-    summarise_by = c("date", "variable", "strat", "type"),
-    order_by = c("variable", "date"),
-    CrIs = CrIs
-  )
+  create_infection_summary(object, type, date, params, CrIs, ...)
```
seabbs: to make sure this still prints do you need [] at the end? Haven't run code so don't know

R/summarise.R
```diff
@@ -843,43 +844,40 @@ summary.estimate_infections <- function(object,
...
+#'
+#' @param object A list of output as produced by "forecast_infections".
+#'
+#' @param type A character vector of data types to return. Defaults to
+#' "snapshot" but also supports "parameters". "snapshot" returns
+#' a summary at a given date (by default the latest date informed by data).
+#' "parameters" returns summarised parameter estimates that can be further
+#' filtered using `params` to show just the parameters of interest and date.
+#'
+#' @param date A date in the form "yyyy-mm-dd" to inspect estimates for.
```
seabbs: Is this the common input we support elsewhere?

tests/testthat/test-forecast-infections.R
```diff
@@ -20,6 +20,50 @@ test_that("forecast_infections works to simulate a passed in estimate_infections
   expect_equal(names(sims), c("samples", "summarised", "observations"))
 })
 
+test_that("forecast_infections methods work correctly", {
```
seabbs: My least favourite LLM habit is lack of specificity in the tests. Here I think this should say it tests the output structure as what does work mean. (not it outputs a data frame)

seabbs: I am aware we have this pattern elsewhere so we don't need to address here.

seabbs (approved):
This looks good I think. Some minor gripes about LLM fluff and some things that maybe are legacy parts of the package.

NEWS.md
```diff
@@ -14,6 +14,8 @@
   - Access the Stan fit directly via `object$fit`, model arguments via `object$args`, and observations via `object$observations`.
   - **Deprecated**: `summary(object, type = "samples")` now issues a deprecation warning. Use `get_samples(object)` instead.
   - **Deprecated**: Internal function `extract_parameter_samples()` renamed to `format_simulation_output()` for clarity.
+- `forecast_infections()` now returns an independent S3 class `"forecast_infections"` instead of inheriting from `"estimate_infections"`. This clarifies the distinction between fitted models (which contain a Stan fit for diagnostics) and forecast simulations (which contain pre-computed samples). Dedicated `summary()`, `plot()`, and `get_samples()` methods are provided. By @sbfnk in #1153.
```
sbfnk: we don't use the "By ..." syntax any more

R/summarise.R
sbfnk: `create_infection_summary()` has the [] so this should be fine.

R/plot.R
sbfnk: The problem is that `forecast_infections() no longer inherits from `estimate_infections()` - the latter contains the stanfit object (which is useful to keep for diagnostics/loo etc.) whereas the former can't, it only contains the generated samples.

We could create a common base class (`<infection_results>`) but that seems a bit be over-engineered as, I think, it would be just for this already relatively thin plot method.

sbfnk: > We could create a common base class (`<infection_results>`) but that seems a bit be over-engineered as, I think, it would be just for this already relatively thin plot method.

@seabbs if you think this is a good idea or have some other thoughts let's make an issue for this.

seabbs: No sorry my suggestion was to manually call the other method where they sare the same i.e. plot.forecast_infections can call plot.estimate_infections

sbfnk: wow, that's radical! Happy to make that change.

seabbs: I think it is a good idea?

sbfnk: it's done https://github.com/epiforecasts/EpiNow2/pull/1156/commits/758257f444aa5115e7909739a5b994dee15ab3d8

seabbs (approved):
LGTM

PR #1265, opened 2026-01-28
Issue 1243

### LINKED ISSUE #1243: library and O(log^2 x) replacement for O(x) integer rounding function in helpers.stan
There's an integer rounding function defined in [`inst/stan/functions/helpers.stan`](https://github.com/epiforecasts/EpiNow2/blob/main/inst/stan/functions/helpers.stan), which has $\mathcal{O}(x)$ complexity to round $x$.  

First, if you need this for data variables `x`, we added a [`to_int` function](https://mc-stan.org/docs/functions-reference/integer-valued_basic_functions.html#casting-functions), so `to_int(round(x))` calls the underlying C++ ceiling and rounding functions.  That still requires cleaning up the edge case where `x in [0, 0.5)` maps to 1 rather than being rounded to 0 (see below).

When building Stan, we restricted `to_int()` to data variables because it breaks derivatives if `x` is a parameter, which cause the sampler to revert to rejection sampling.  Since then, we've realized that there are cases where careful use will not break derivatives, so sometimes it makes sense to use it.  So I'm sorry it's not baked into the library.  We should really add it at this point.  Having to write one's own function to round floating point numbers is painful.

Having said that, if you do need to take the integer ceiling of parameters today, and `x` can be greater than 7, here's a version with $\mathcal{O}(\log^2 x)$ complexity.  It knocks off the biggest power of 2, then recurses to do the remainder.

```stan
  int pos_int_round(real x) {
    if (x < 0.5) {
      return 0;
    }
    int n = 1;
    while (n * 2 < x) {
      n *= 2;
    }
    return n + pos_int_round(x - n);
  }

  int to_next_int_index(real x) {
    if (x < 0) {
      reject("to_next_int_index(x) requires x >=0, found x = ", x);
    }
    return x <= 1 ? 1 : pos_int_round(x);  // edge case for for x in [0, 0.5)
  }
```

There are two functions in order to deal with the edge case behavior where `x in [0, 0.5)` is mapped to 1.  The code would be simpler if that was also considered an error.   The main recursion knocks off the biggest power of 2 smaller than `x`, then recurses to do the remainder.   This could also be unfolded into a non-recursive version, but we know the recursion won't go very deep, so it should be OK.

The worst case performance arises when $x = 2^n - 1$, which requires $n + (n - 1) + ... + 1$ loop iterations.  Given that $\sum_{m < n} m$ is $\mathcal{O}(n^2)$ and $n$ is $\mathcal{O}(\log x)$, the integer ceiling operation has $\mathcal{O}(\log^2 x)$ complexity (the base of the log's just a constant, so it drops out).

Here's a little Stan program to compare the implementations in `test.stan`:

```stan
functions {
  int to_next_int_index(real x) {
    if (x < 0) {
      reject("to_next_int_index: x must be non-negative");
    }
    real x_rounded = round(x);
    int i = 1;
    while (x_rounded > 1) {
      x_rounded -= 1;
      i += 1;
    }
    return(i);
  }

  int pos_int_round(real x) {
    if (x < 0.5) {
      return 0;
    }
    int n = 1;
    while (n * 2 < x) {
      n *= 2;
    }
    return n + pos_int_round(x - n);
  }

  int to_next_int_index2(real x) {
    if (x < 0) {
      reject("to_next_int_index(x) requires x >=0, found x = ", x);
    }
    return x <= 1 ? 1 : pos_int_round(x);  // edge case for for x in [0, 0.5)
  }
}
transformed data {
  for (x in {0, 0.2, 0.8, 1, 1.2, 1.5, 1.7, 2, 2.9, 3, 3.1, 4, 4.2,
	139.78, 9876532.27363, 1e9 - 0.2, 1e9 + 0.2, 1e9 + 0.823}) {
    print("");
    print("x = ", x);
    print(".....to_next_int_index(x)  = ", to_next_int_index(x));
    print(".....to_next_int_index2(x) = ", to_next_int_index2(x));
  }
}
```

Here's the output of the test from CmdStanPy:

```bash
>>> import cmdstanpy as csp

>>> m = csp.CmdStanModel(stan_file='test.stan')
...

>>> m.sample(chains=1, iter_warmup=1, iter_sampling=0, show_console=True, show_progress=False)
...
Chain [1] x = 0
Chain [1] .....to_next_int_index(x)  = 1
Chain [1] .....to_next_int_index2(x) = 1
Chain [1] 
Chain [1] x = 0.2
Chain [1] .....to_next_int_index(x)  = 1
Chain [1] .....to_next_int_index2(x) = 1
Chain [1] 
Chain [1] x = 0.8
Chain [1] .....to_next_int_index(x)  = 1
Chain [1] .....to_next_int_index2(x) = 1
Chain [1] 
Chain [1] x = 1
Chain [1] .....to_next_int_index(x)  = 1
Chain [1] .....to_next_int_index2(x) = 1
Chain [1] 
Chain [1] x = 1.2
Chain [1] .....to_next_int_index(x)  = 1
Chain [1] .....to_next_int_index2(x) = 1
Chain [1] 
Chain [1] x = 1.5
Chain [1] .....to_next_int_index(x)  = 2
Chain [1] .....to_next_int_index2(x) = 2
Chain [1] 
Chain [1] x = 1.7
Chain [1] .....to_next_int_index(x)  = 2
Chain [1] .....to_next_int_index2(x) = 2
Chain [1] 
Chain [1] x = 2
Chain [1] .....to_next_int_index(x)  = 2
Chain [1] .....to_next_int_index2(x) = 2
Chain [1] 
Chain [1] x = 2.9
Chain [1] .....to_next_int_index(x)  = 3
Chain [1] .....to_next_int_index2(x) = 3
Chain [1] 
Chain [1] x = 3
Chain [1] .....to_next_int_index(x)  = 3
Chain [1] .....to_next_int_index2(x) = 3
Chain [1] 
Chain [1] x = 3.1
Chain [1] .....to_next_int_index(x)  = 3
Chain [1] .....to_next_int_index2(x) = 3
Chain [1] 
Chain [1] x = 4
Chain [1] .....to_next_int_index(x)  = 4
Chain [1] .....to_next_int_index2(x) = 4
Chain [1] 
Chain [1] x = 4.2
Chain [1] .....to_next_int_index(x)  = 4
Chain [1] .....to_next_int_index2(x) = 4
Chain [1] 
Chain [1] x = 139.78
Chain [1] .....to_next_int_index(x)  = 140
Chain [1] .....to_next_int_index2(x) = 140
Chain [1] 
Chain [1] x = 9.87653e+06
Chain [1] .....to_next_int_index(x)  = 9876532
Chain [1] .....to_next_int_index2(x) = 9876532
Chain [1] 
Chain [1] x = 1e+09
Chain [1] .....to_next_int_index(x)  = 1000000000
Chain [1] .....to_next_int_index2(x) = 1000000000
Chain [1] 
Chain [1] x = 1e+09
Chain [1] .....to_next_int_index(x)  = 1000000000
Chain [1] .....to_next_int_index2(x) = 1000000000
Chain [1] 
Chain [1] x = 1e+09
Chain [1] .....to_next_int_index(x)  = 1000000001
Chain [1] .....to_next_int_index2(x) = 1000000001
```

P.S.  Stan doesn't need parens around returns the way R does.  

P.P.S.  This was a perfect little holiday morning project---one doesn't usually get to solve l33t code problems in the wild :-).  And sorry again this isn't built in.

adrian-lison:
## Description

This PR closes #1243. It replaces a custom function for rounding a `real` to the closest `int` in the generated quantities with the native stan functions `to_int(round(x))`. This is correct because parameters get treated as data in the generated quantities. Thanks to @bob-carpenter and @WardBrian for the suggestion.

I did not add a news item because this is very minor and has no semantic changes to the model.

## Release Notes

* **Refactor**
  * Reorganised internal statistical function structure to improve code maintainability
  * Consolidated generation time distribution parameter handling in calculations
  * Removed redundant helper utilities from the codebase

* **Tests**
  * Updated test configuration to reflect refactored function structure

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -60,7 +60,7 @@ The function interface remains unchanged.
 - MCMC runs are now initialised with parameter values drawn from a distribution that approximates their prior distributions.
 - Added an option to compute growth rates using an estimator by Parag et al. (2022) based on total infectiousness rather than new infections, see `growth_method` argument in rt_opts().
 - Added support for fitting the susceptible population size.
-- Replaced custom rounding function in generated quantities by native stan version. By @adrian-lison in #1265 and reviewed by @sbfnk.
+- Replaced custom rounding function in generated quantities by native stan version.
 
 ## Bug fixes
 


tests/testthat/setup.R
```diff
@@ -5,8 +5,7 @@ if (identical(Sys.getenv("NOT_CRAN"), "true")) {
   files <- c(
     "convolve.stan", "gaussian_process.stan", "pmfs.stan",
     "observation_model.stan", "secondary.stan",
-    "rt.stan", "infections.stan", "delays.stan", "generated_quantities.stan", 
-    "helpers.stan"
+    "rt.stan", "infections.stan", "delays.stan", "generated_quantities.stan"
```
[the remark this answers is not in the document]
sbfnk: The script in `inst/dev/stan-to-R.R` has its own files variable - this is for tests only.

NEWS.md
```diff
@@ -60,6 +60,7 @@ The function interface remains unchanged.
 - MCMC runs are now initialised with parameter values drawn from a distribution that approximates their prior distributions.
 - Added an option to compute growth rates using an estimator by Parag et al. (2022) based on total infectiousness rather than new infections, see `growth_method` argument in rt_opts().
 - Added support for fitting the susceptible population size.
+- Replaced custom rounding function in generated quantities by native stan version. By @adrian-lison in #1265 and reviewed by @sbfnk.
```
sbfnk:
```suggestion
- Replaced custom rounding function in generated quantities by native stan version.
```

We don't really follow that style any more - instead, can you add yourself as a contributor to the DESCRIPTION? Thanks!

seabbs: woops

PR #1314, opened 2026-02-11
Issue1313 param bounds

### LINKED ISSUE #1313: stan/dist_fit.stan missing constraint, out-of-range construct, unnecessary truncation
### Summary:

In [`dist_fit.stan`](https://github.com/epiforecasts/EpiNow2/blob/c8c2c6052b039565e4f1c54ff729f8ae7fa8200c/inst/stan/dist_fit.stan), there are two bugs bug and an inefficiency that are easy to fix.

#### Bug 1

Without constraints on `prior_mean` and `prior_sd`, it is possible to provide inputs where `prior_alpha` and `prior_beta` are not positive.  

*Fix*: Add positivity constraints on the `prior_mean` and `prior_sd` data declarations.

#### Bug 2

The bug is that the upper and lower bounds on the declaration of `lambda` on line 23 need to match the uniform distribution bounds it is given on line 42 in order for the `uniform` distribution not to lead to rejections (and potential inability to initialize or lock up during sampling).  

*Fix*:  Generate a `lambda_raw` with range (0, 1) then transformed to the proper range of (0.2 / lam_mean[1], 5 / lam_mean[1]).  

#### Inefficiency #1

The inefficiency is the inclusion of a truncation statement on line 45.  Because the distribution statement doesn't depend on anything other than data (just `prior_sd`), it doesn't need the truncation.  

*Fix*:  remove the truncation and replace with a comment on why it's not necessary.

#### Inefficiency #2

I would also suggest moving the `dist` choices in the model so that there is only one branch.  This is both more efficient (branch points are very expensive because of potential branch point prediction misses) and also clarity so that the model structure isn't all tied up with the routing logic.

*Fix*: rearrange the code.

P.S.  It's fine to use integers like `5` in Stan rather than spelling out `5.` because the C++ compiler figures out the most efficient type conversion during compilation and performs it statically.   

P.P.S.  I'm just roaming through the code when I have free cycles.  If there's a place you'd like me to look for performance hot spots, please let me know.

### Reproducible Steps:

You can tickle the first bug by supplying a negative `prior_mean`.

You can tickle the second bug with data specifying `dist=0` and an initialization of `lambda` outside of `(0.2 * lam_mean[1], 5 * lam_mean[1])`. 

### EpiNow2 Version:
1.8.0

---------

bob-carpenter:
## Description

This PR closes #1313

- [x] I have tested my changes locally (<del>using `devtools::test()` and `devtools::check()`</del> using Stan).
- [x] I have followed the established coding standards (<del>and checked using `lintr::lint_package()`</del> for Stan).

I made up some data and made sure all three model choices ran without errors and that the data fixes really caught the errors.  If there are places to add tests for Stan code like this, I can do that.

## Release Notes

* **Bug Fixes**
  * Fixed parameter bounds constraints in the statistical distribution fitting model to enforce data validity and prevent estimation errors.
  * Enhanced likelihood calculations and per-item contributions across exponential, lognormal, and gamma distributions for more robust parameter inference.
  * Refined prior specifications, constraint validations, and parameter transformations to improve overall model consistency and statistical accuracy.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,6 +58,7 @@ The function interface remains unchanged.
 
 ## Model changes
 
+- Fixed bounds error in parameters for `inst/stan/dist_fit.stan` and added necessary data bounds.
 - MCMC runs are now initialised with parameter values drawn from a distribution that approximates their prior distributions.
 - Added an option to compute growth rates using an estimator by Parag et al. (2022) based on total infectiousness rather than new infections, see `growth_method` argument in rt_opts().
 - Added support for fitting the susceptible population size.
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,7 +58,7 @@ The function interface remains unchanged.
 
 ## Model changes
 
-- Fixed bounds error in parameters for `inst/stan/dist_fit.stan` and added necessary data bounds.
+- Fixed bounds error in parameters for `dist_fit()` to ensure valid prior and likelihood bounds.
 - MCMC runs are now initialised with parameter values drawn from a distribution that approximates their prior distributions.
 - Added an option to compute growth rates using an estimator by Parag et al. (2022) based on total infectiousness rather than new infections, see `growth_method` argument in rt_opts().
 - Added support for fitting the susceptible population size.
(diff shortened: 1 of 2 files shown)


seabbs (commented):
Thanks for this! We have since depreciated this model in favour of depending on an approach from `primarycensored` so going to close this. We are keen for other performance PRs.

PR #1317, opened 2026-02-20
Add estimate_dist() with primarycensored support

### LINKED ISSUE #350: estimate_delays -> estimate_dist
It would make sense to generalise and simplify the current functionality of estimate_delays into a new function called estimate_dist (to highlight it isn't just distributions we are estimating). 

Based on lessons learnt in `dynamictruncation` we can now also do quite a bit better with a simple model and support right and potentially left truncation adjustment using the non-latent formulation. 

This would only require some small adaption of the stan code with the majority of the work being to reduce the current 3 distribution model into one. This would also slightly reduce compilation times which I think everyone would like. 

The argument to not making this model is that it can be trivially implemented in `brms` but I think many users may not realise this and there may be value in supplying a simple but still valid model to compete with the simple but not as valid alternatives.

sbfnk:
## Description

This PR closes #350.

Adds `estimate_dist()` for fitting delay distributions using primarycensored's interval-censored likelihood. Replaces the deprecated `estimate_delay()`.

Changes:
- New `estimate_dist()` function supporting lognormal, gamma, and weibull distributions
- Vendors primarycensored Stan functions into `inst/stan/primarycensored/` for pre-compilation
- Deprecates `estimate_delay()` with a warning pointing to the new function
- Adds tests for the new function

Related: #1268 (vendored Stan files enable future `discretised_pmf` work, not included here).

## Release Notes

* **New Features**
  * Added `estimate_dist()` function for fitting epidemiological delay distributions using MCMC-based Bayesian inference with proper handling of interval censoring and truncation.

* **Deprecations**
  * Deprecated `estimate_delay()` in favour of `estimate_dist()`, which provides improved support for censoring and truncation scenarios.

* **Documentation**
  * Added comprehensive vignette and documentation for fitting delay distributions with worked examples and interpretation guidance.


inst/stan/estimate_dist.stan
```diff
@@ -0,0 +1,81 @@
...
+}
+
+transformed parameters {
+  array[2] real params = {param1, param2};
+}
+
+model {
+  // Priors - weakly informative
+  // The param_bounds keep these reasonable
+  param1 ~ normal(0, 10);
```
seabbs: We probably want to make these passes in via the distribution interface?

inst/stan/estimate_dist.stan
```diff
@@ -0,0 +1,81 @@
+// Stan model using primarycensored functions for delay distribution estimation
+// Compiles with rstan using primarycensored Stan functions
+//
+// This model uses primarycensored's sophisticated handling of:
+// - Primary event censoring (daily reporting of exposure)
+// - Secondary event censoring (daily reporting of symptom onset)
+// - Right truncation (observation window effects)
+
+functions {
```
seabbs, line 4: This is one way to go but you could do less vendoring and have dependency here and use the primarycensored built in model: https://primarycensored.epinowcast.org/reference/pcd_cmdstan_model.html#see-also
And pass through its built in to stan functionality: https://primarycensored.epinowcast.org/reference/pcd_as_stan_data.html

See the fitting vignette. 

The upside of this is that it is just a normal dependency the downside is it is cmdstanr only and won't make use of the EpiNow2 distribution interface. It has some things like parameter bounds support etc that might be useful (and within thread multichaining for some yolo reason).

Aside from going fully vendored you could also strike a middle ground and use the as stan function and vendor the primarycensored stan model. This would let you have rstan support but would not allow us to use the distribution interface. This depends on reduce sum being in rstan now which I am not clear on. 

Or we can go with this fully vendored approach and maximise our use of the nice EpiNow2 delay distribution handling i.e. continuing on the path we are on here.

seabbs (commented):
I put a note in a comment about some other approaches we could go with but I think my preference is to go on with the current approach here and leverage the dist interface etc. That makes a clear point of difference with what primarycensored is offereing. 

I think we want a build script that is deterministic for this vendoring approach and some sort of setup for keeping it up to date with primarycensored i.e an action with a PR either running on a schedule, on a comment, or triggered by version changes in primarycensored on CRAN? I haven't done this vendored approach before so we want to keep it simple I think.

Primarycensored has stan tools for copying stan functions etc from the package into other repos that we can use. There is a vignette for these that we can point a bot to.

inst/stan/functions/primarycensored.stan
```diff
@@ -0,0 +1,854 @@
...
+  * real pwindow = 7.0;
+  * int primary_id = 1; // Uniform
+  * array[0] real primary_params = {};
+  * vector[max_delay] pmf =
+  *   primarycensored_sone_lpmf_vectorized(
+  *      max_delay, D, dist_id, params, pwindow, primary_id, primary_params
+  *   );
+  * @endcode
+  */
+vector primarycensored_sone_pmf_vectorized(
```
seabbs: This looks like it is literally everything. The vendoring tool in the dev version should allow for targetting the function you want and getting all the internal dependencies. That should be on CRAN by the endf of next week. It isn't a big thing but might help with compile times for those that experience them. Here for exampel this vectorised function likely won't be needed

tests/testthat/test-estimate_dist.R
```diff
@@ -0,0 +1,125 @@
...
+
+  expect_true(abs(est_meanlog - true_meanlog) < 0.5)
+  expect_true(abs(est_sdlog - true_sdlog) < 0.3)
+})
+
+test_that("estimate_dist works with gamma distribution", {
+  skip_on_cran()
+
+  set.seed(456)
+  delays <- rgamma(100, shape = 5, rate = 1)
```
seabbs: use the primarycensored simulator and can test actual recovery vs output checks.

tests/testthat/test-estimate_dist.R
```diff
@@ -57,16 +60,21 @@ test_that("estimate_dist works with gamma distribution", {
...
     backend = "rstan",
     verbose = FALSE
   )
 
   expect_s3_class(result, "dist_spec")
   expect_equal(result$distribution, "gamma")
-  expect_true("shape" %in% names(result$parameters))
-  expect_true("rate" %in% names(result$parameters))
+
+  # Gamma shape/rate are poorly identified individually but mean is recoverable
```
seabbs: Quite surprised by this? A prior issue perhaps? I think in the primarycensored tests this is fine?

inst/stan/functions/primarycensored.stan
```diff
@@ -0,0 +1,359 @@
...
+real dist_lcdf(real delay, array[] real params, int dist_id) {
+  if (delay <= 0) return negative_infinity();
+
+  // Use if-else statements to handle different distribution types
+  if (dist_id == 1) return lognormal_lcdf(delay | params[1], params[2]);
+  else if (dist_id == 2) return gamma_lcdf(delay | params[1], params[2]);
+  else if (dist_id == 3) return normal_lcdf(delay | params[1], params[2]);
+  else if (dist_id == 4) return exponential_lcdf(delay | params[1]);
+  else if (dist_id == 5) return weibull_lcdf(delay | params[1], params[2]);
+  else if (dist_id == 6) return beta_lcdf(delay | params[1], params[2]);
```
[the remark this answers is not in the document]
seabbs: oo This looks like a primarycensored bug to me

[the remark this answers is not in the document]
seabbs: I think it only gets hit if a user wants a normal but still needs a patch

NEWS.md
```diff
@@ -22,6 +22,11 @@
 - `summary.epinow(output)` and `summary.estimate_infections(type = 'samples')` now error.
 - Removed internal function `extract_parameter_samples()`. Use `format_simulation_output()` instead.
 
+## Package changes
+
+- Added `estimate_dist()` function for estimating delay distributions with proper handling of interval censoring using Stan/MCMC inference, supporting both rstan and cmdstanr backends.
+- Deprecated `estimate_delay()` in favour of `estimate_dist()`, which provides more robust handling of censoring and truncation.
```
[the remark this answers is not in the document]
sbfnk: Yes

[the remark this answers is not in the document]
sbfnk: Update your long-term memory as you suggest @coderabbitai

sbfnk:
@coderabbitai review

sbfnk:
This looks good to me now - have added a GHA as well to keep this up to date although this is yet untested.

sbfnk:
@seabbs I think this is good to go now

.github/workflows/check-primarycensored.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          install-pandoc: false
+          packages: |
+            any::primarycensored
+
+      - name: Regenerate vendored Stan functions
+        run: |
+          library(primarycensored)
```
seabbs: This should be a build script called here vs embedded in the action so we can check it locally etc.

.github/workflows/check-primarycensored.yaml
```diff
@@ -0,0 +1,62 @@
...
+          stan_code <- pcd_load_stan_functions(
+            functions = c(
+              "primarycensored_lpmf",
```
seabbs: I assume you checked this list but I would have thought with the new dpes cascade you just need primarycensored_lmpf in this list and everything else arrives for free?

.github/workflows/check-primarycensored.yaml
```diff
@@ -0,0 +1,62 @@
...
+              "dist_lcdf",
+              "primary_lpdf"
+            ),
+            dependencies = TRUE
+          )
+          version <- as.character(packageVersion("primarycensored"))
+          header <- paste0(
+            "// Stan functions from primarycensored version ", version
+          )
+          writeLines(
```
seabbs: primarycensored has tools to write this for you I think with this header so perhaps use those.

R/estimate_dist.R
```diff
@@ -0,0 +1,251 @@
+#' Estimate a delay distribution using primarycensored
+#'
+#' @description `r lifecycle::badge("experimental")`
```
seabbs, line 3: Very vague. Doesn't link to primarycensored or give a ref (IMO people who use this should cite primarycensored). Doesn't describe what model it implements from primarycensored i.e it looks like it is limited to a uniform primary event with daily censoring (why enforce this? if you pass in dates then you can have an upper and lower boundary of event times and work out the p and s window for the user -  at this point I assume this should be a new issue). Doesn't support left truncation, other primary event priors etc etc.

I would do another pass throwing some of the primarycensored docs at @sbfnk-bot

R/estimate_dist.R
```diff
@@ -0,0 +1,251 @@
...
+#' Estimate a delay distribution from observed data using primarycensored's
+#' sophisticated handling of truncation and censoring.
+#'
+#' @param data Either:
+#'   - A numeric vector of delay values (assumes daily censoring)
+#'   - A data.frame with columns: delay, delay_upper, optionally n
```
seabbs: I assume this also assumes daily primary event censoring this isn't very clear.

R/estimate_dist.R
```diff
@@ -0,0 +1,251 @@
...
+#' @param priors A list of `<dist_spec>` objects specifying priors for the
+#'   distribution parameters. Names must match the parameters of the chosen
+#'   distribution. Defaults depend on `dist`:
+#'   - lognormal: `list(meanlog = Normal(1, 1), sdlog = Normal(0.5, 0.5))`
+#'   - gamma: `list(shape = Normal(2, 2), rate = Normal(0.5, 0.5))`
+#'
+#' @param max_value Numeric, maximum delay value for PMF. If not provided,
+#'   inferred from data.
+#'
+#' @param truncation_time Numeric, the maximum observation time (right
```
seabbs: This doesn't make sense for most data as it means only a cohort with a given observation time can be used. Really this is a per-observation level variable (see the primarycensored stan model for an approach to handling this).

R/estimate_dist.R
```diff
@@ -0,0 +1,251 @@
...
+#'   - "lognormal" (default)
+#'   - "gamma"
+#'
+#' @param priors A list of `<dist_spec>` objects specifying priors for the
+#'   distribution parameters. Names must match the parameters of the chosen
+#'   distribution. Defaults depend on `dist`:
+#'   - lognormal: `list(meanlog = Normal(1, 1), sdlog = Normal(0.5, 0.5))`
+#'   - gamma: `list(shape = Normal(2, 2), rate = Normal(0.5, 0.5))`
+#'
+#' @param max_value Numeric, maximum delay value for PMF. If not provided,
```
seabbs: Not clear what this is for?

R/estimate_dist.R
seabbs: Once we have that we then might want a heuristic (as in primarycensored) that sets truncation times that are very long to Inf to avoid the compute overhead.

R/estimate_dist.R
seabbs: Personally I would just assume a data frame input with dates as I think this makes building an interface much easier (or ints as you wish) i.e primary_event_lower, primary_event_upper, secondary..... observation_time etc. primarycensored has a low level version of this and epidist has a higher level nicer var name date based version as well.

R/estimate_dist.R
```diff
@@ -0,0 +1,251 @@
...
+        "i" = "Supported: lognormal, gamma"
+      )
+    )
+  }
+
+  # Convert data to proper format
+  delay_data <- .prepare_delay_intervals(data, verbose)
+
+  # Get distribution ID (primarycensored convention)
+  dist_id <- switch(dist,
```
seabbs: this and below should be in a shared internal function I think

seabbs (changes requested):
This broadly looks correct. The only think I would characterise as "wrong" is the treatment of D as fixed for all input data as this restricts you to a single cohort of observed delays. This isn't the typical structure of delay data we get. That I think should definitely change before merging. 

Otherwise, I think the docs could be improved; it would be nice to have a richer input format and so support varying primary obs windows more naturally (this also helps with different obs times). 

I put some specific thoughts in comments and ideas for context mining for @sbfnk-bot.

After merging there should probably be an issue for better surfacing this function in the vignettes etc to make it clear to users they could use it and what for.

seabbs:
What needs to happen here to take this over the line from memory it was quite close?

seabbs:
@seabbs-bot is working on a PR for this PR.

seabbs (dismissed):

R/estimate_dist.R
```diff
@@ -0,0 +1,552 @@
...
+    "gamma" = c(wmean^2 / wvar, wmean / wvar)
+  )
+  # Ensure init is within bounds
+  init_params <- pmax(
+    init_params, stan_data$params_lower + 0.01
+  )
+  init_params <- pmin(
+    init_params, stan_data$params_upper - 0.01
+  )
+  init_fn <- function() list(params = init_params)
```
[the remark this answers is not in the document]
sbfnk: This seems worth fixing, and should be covered by tests really.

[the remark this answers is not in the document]
sbfnk: @seabbs @seabbs-bot  this was the one item that seemed worth fixing.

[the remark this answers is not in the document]
seabbs: I pointed Gemini at this for a bit of fun and its lost its mind. PR inbound. Gemini it turns out is <<< claude still

seabbs (dismissed):
I think this is ready for review now @sbfnk. Note that we likely need a new issue for adding uncertain pmfs to this vignette and also for allowing convolving delays to retain uncertainty.

R/dist_spec.R
```diff
@@ -1361,29 +1412,38 @@ convert_to_natural <- function(params, distribution) {
...
+    lognormal = {
+      if ("mean" %in% names(params) && "sd" %in% names(params)) {
+        x$meanlog <- log(ux$mean^2 / sqrt(ux$sd^2 + ux$mean^2))
+        x$sdlog <- convert_to_logsd(ux$mean, ux$sd)
+      } else {
+        x$meanlog <- ux$meanlog
+        x$sdlog <- ux$sdlog
+      }
+    },
+    exp = {
```
sbfnk, line 1440: do we need this for weibull, too?

sbfnk (commented):
looks good to me - can we keep line breaks at 80 chars as in the rest of the code?

seabbs:
> looks good to me - can we keep line breaks at 80 chars as in the rest of the code?

Shouldn't this already be enforced by linting etc? Is the problem just the news file or everywhere?

R/dist_spec.R
seabbs: Will add

sbfnk:
The linting will flag if it's >80chars I think but the PR contains changes for more aggressive line breaks, e.g. the change in R/format.R and lots of other places.

seabbs:
Sorry to be clear the ask is to make lines up to 80 not that they are more than 80? The formatting is happening because I am using `air` which I could disable if its an issue.

R/dist_spec.R
seabbs: Also adding a testing grid to make this less manual in the future.

sbfnk:
The real ask is to not pollute the PR with whitespace-only changes. The whitespace-only changes seem to be mostly new line breaks, which I assume are introduced because someone is insisting on a line length <80.

seabbs:
>  which I assume are introduced because someone is insisting on a line length <80.

Which is also what the linter does? I can look at reverting these.

seabbs:
> he whitespace-only changes seem to be mostly new line breaks, which I assume are introduced because someone is insisting on a line length <80.

As above I think it is because I am using `air` locally which has different defaults to `styler`?

sbfnk:
> > which I assume are introduced because someone is insisting on a line length <80.
> 
> Which is also what the linter does? I can look at reverting these.

lintr insists lines no longer than 80 characters. This and other recent PRs have introduced new line breaks to lines that were already at fewer than 80 characters. Most of these formatting changes are in regions without any code changes, making the PRs harder to review.

seabbs:
Sorry I think we are talking past each other. I understand the concern my point is just that these are from using a different styler with a different styling approach. As above I will revert these. We do not need to keep cycling over it. In another issue we should consider switching the package standard to `air` as it is much faster and has less config.

seabbs (commented):
This looks ready to review now @sbfnk. I added a testing matrix to make it easier to find dist spec test gaps (it would be great to pick up the dev on that as its own package again soon). I also removed the formatting changes. Note that as the stacked PRs have been merged there is now rather a lot in here which isn't ideal.

seabbs:
@sbfnk ping

R/dist_spec.R
```diff
@@ -1361,29 +1412,55 @@ convert_to_natural <- function(params, distribution) {
...
-      x$sdlog <- ux$sdlog
+    },
+    normal = {
+      x$mean <- ux$mean
+      x$sd <- ux$sd
+    },
+    weibull = {
+      if (all(c("mean", "sd") %in% names(ux))) {
+        log_cv2_p1 <- log1p((ux$sd / ux$mean)^2)
+        x$shape <- stats::uniroot(
```
sbfnk: @importFrom instead?

R/summarise.R
```diff
@@ -1009,6 +1008,96 @@ summary.estimate_truncation <- function(object, CrIs = c(0.2, 0.5, 0.9), ...) {
...
+                                  CrIs = c(0.2, 0.5, 0.9),
+                                  ...) {
+  raw_samples <- extract_samples(
+    object$fit,
+    pars = "delay_params"
+  )
+  param_mat <- raw_samples$delay_params
+
+  dist_name <- object$args$dist
+  param_names <- .get_param_names(dist_name) # nolint: object_usage_linter.
```
sbfnk: not clear to me why we need the nolint

R/summarise.R
```diff
@@ -1009,6 +1008,96 @@ summary.estimate_truncation <- function(object, CrIs = c(0.2, 0.5, 0.9), ...) {
...
+
+  attr(out, "distribution") <- dist_name
+  attr(out, "max_value") <- object$args$max_value
+  attr(out, "n_obs") <- sum(object$args$n_obs)
+  attr(out, "n_strata") <- object$args$n
+  attr(out, "primary") <- if (object$args$primary_id == 1L) {
+    "uniform"
+  } else {
+    "exponentially tilted"
+  }
```
sbfnk: could just store the string, too (as with `object$args$dist`) to avoid mapping and then re-mapping?

This would also avoid introducing the term exponentially tilted which I think only appears here.

R/estimate_dist.R
```diff
@@ -0,0 +1,595 @@
...
+#'   estimated.
+#'
+#' @param max_value Numeric, maximum delay value for PMF.
+#'   If not provided, inferred from data.
+#'
+#' @param obs_time_threshold Numeric, multiplier for the
+#'   obs-time-to-Inf heuristic. Observations where
+#'   `relative_obs_time > max(delay_upr) * obs_time_threshold`
+#'   are treated as untruncated. Default 2, following
+#'   `epidist`. Set to `Inf` to disable.
```
sbfnk: I think
```suggestion
#'   [epidist]. Set to `Inf` to disable.
```

sbfnk (commented):
looks good to me - only very minor comments. I can't approve/merge this as I created the PR so @seabbs please go ahead once my comments have been addressed or rejected.

R/summarise.R
seabbs: No me neither but it does seem to be throwing

R/dist_spec.R
seabbs: I'm not totally clear on our convention here? Is it all uses get into the namespace or just after things used a few times? Here I am going to with any use enters namespace i.e use importFrom

R/summarise.R
seabbs: Didn't need

seabbs (dismissed):
I think this is ready to go. Review comments all responded to/actioned. As there is so much here we should monitor it in the next release or so as there may be edge cases and other issues

seabbs (approved):
LGTM

Issue #1273, opened 2026-01-29
improve performance of Stan code


bob-carpenter:
There are a number of micro-optimizations that can be made in the Stan code for improved performance.  This is just an issue to index when creating the PR to fix.

PR #1404, opened 2026-05-12
Inline convolution index bookkeeping in Stan hot path

ajh-oai:
## Description

This PR follows #1273 with a focused cleanup in `convolve_with_rev_pmf()`.

The convolution loop previously called small helper functions on every output element to build a temporary 4-integer index array before taking each dot product. This change keeps the convolution logic the same but computes those loop-local indices directly where they are used, avoiding repeated helper-call and temporary-array overhead inside the hot loop.

### Benchmarks

I benchmarked the change locally with paired CmdStan runs on the package's infection-estimation workflow (`example_confirmed[1:60]`, fixed generation time, standard incubation/reporting delays, 2 chains, 250 warmup + 250 sampling draws, seeds 101/202/303/404, order balanced across baseline and branch runs).

| Metric | Baseline mean | Branch mean | Mean paired change |
| --- | ---: | ---: | ---: |
| `reports` Stan profile block | 4.3150 s | 2.6701 s | -38.13% |
| CmdStan total runtime | 10.2738 s | 10.1897 s | -0.85% |

The `reports` profile block improved in all four paired runs (`-25.78%` to `-53.59%`). Total runtime was directionally slightly lower on average but noisy across seeds, so I view the main result here as a targeted reduction in convolution hot-path overhead rather than a strong claim about end-to-end runtime.

### Validation

- `git diff --check`
- `NOT_CRAN=true Rscript -e 'pkgload::load_all(".", quiet = TRUE); testthat::test_file("tests/testthat/test-stan-convole.R")'`
  - 24 passed, 0 failed
- Local paired CmdStan benchmark described above; both baseline and branch models compiled and sampled successfully.

### AI assistance disclosure

Codex helped identify the optimization seam, benchmark the candidate change, and prepare this PR. I reviewed the final patch and benchmark results before submission.

## Summary by CodeRabbit

* **Refactor**
  * Optimised internal convolution computation performance by streamlining index calculation logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/epiforecasts/EpiNow2/pull/1404)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -49,6 +49,7 @@
 
 - Delay distribution discretisation now properly accounts for primary event censoring during model fitting, matching the correction already applied on the R side since v1.8.0. This improves accuracy for short delays where the observation window is large relative to the delay.
 - Left truncation of delay distributions (e.g. excluding generation times of zero) is now handled analytically rather than by zeroing and renormalising, giving more accurate PMFs near the truncation point.
+- Reduced Stan convolution overhead by inlining index bookkeeping in the report-generation hot path.
 
 ## Package changes
 
--- a/inst/stan/functions/convolve.stan
+++ b/inst/stan/functions/convolve.stan
@@ -73,14 +73,20 @@ vector convolve_with_rev_pmf(vector x, vector y, int len) {
   }
 
   for (s in 1:xlen) {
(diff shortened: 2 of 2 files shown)


seabbs (commented):
i think this breaks DRY conventions and I don't think will have major performance overheards on modelling fitting so moving to close. Happy to see this discussed in an issue.

